PR #22, opened 2023-07-18
Modify production generation job for real-time parallel

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR removes the 3-km North America grib2 files on the polar stereographic grid and adds the testbed files on the 3-km CONUS grid.  PPN_RUN_PRDGEN is also changed from 128 to 12 because only 12 tasks are needed for generating the various output files with CFP.  I'm not sure why PPN_RUN_PRDGEN was set to 128 because only 22 tasks were previously needed.

## TESTS CONDUCTED: 

- On machines/platforms:

- [x ] WCOSS2
- [ ] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [ ] Non-DA engineering test
- [x ] Sample script: I tested the relevant changes on WCOSS2 outside of the workflow and had no issues.  I can provide my test script if needed.


MatthewPyle-NOAA (commented):
Quick question on the four lines removed (bgdawp, etc.).  Assuming that is just cleanup unrelated to the removal of the NA output grid?

MatthewPyle-NOAA (approved):
Approving now...

Issue #11, opened 2023-07-06
Move templates to PARM directory


chan-hoo:
- Currently the template files are located in the 'ush' directory.
- These templates files should be moved into the 'PARM' directory under the 'HOME' directory (HOMErrfs/PARM).

PR #23, opened 2023-07-18
Move templates to parm directory

chan-hoo:
## DESCRIPTION OF CHANGES: 

- Move the template files from `ush/templates` to `parm`.
- Change `EXECDIR` to `EXECdir`.
- Remove the `ush/wrappers` directory.

## TESTS CONDUCTED: 

- On machines/platforms:

- [x] WCOSS2
- [ ] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [x] Non-DA engineering test
- [ ] Sample script:

## ISSUE: 

- Fixes the issue(s) mentioned in #11


BenjaminBlake-NOAA (commented):
@chan-hoo Thanks for putting these changes together!  I tested them with the 3-km CONUS grid on WCOSS2, and I had an issue with the run_post tasks where the wgrib2 module is not loaded.  Do we need to add that to the run_post.local.lua modulefile?

I also have two questions for you regarding these changes:

1. What were the issues that led to the place=excl and place=vscatter change?  I thought for the post job there was an error in a previous PR, and we need to use the place=excl setting there.
2. What is the reason for the OMP_NUM_THREADS change in exregional_make_lbcs.sh?  Are there instances where we won't want to use 2 threads for that task?

chan-hoo:
@BenjaminBlake-NOAA, 1) Yes. NCO requires that each task should load its own module files, so WCOSS2 does not load the build modules when running the tasks. I've added `wgrib2` to the `run_post` module file. 2) When I tested this PR on WCOSS2, the tasks of `make_ics`, `make_lbcs`, and `run_post` failed. Once I changed `place=excl` to `place=vscatter`, the error was gone. I am not sure what happened there. 3) I didn't intend the change of OMP_NUM_THREADS for make_lbcs. Previously, it was fixed to "2" in the script. I just makes it vary with the parameter value `TPP_MAKE_LBCS` in the configuration file. I didn't test it with TPP_MAKE_LBCS=2, but I think it should work with 2 too. I'll change `TPP_MAKE_LBCS=2` by default.

BenjaminBlake-NOAA:
@chan-hoo  That all sounds good to me, thanks!  I approve of these changes but would like to make sure the run_post task works for me now on WCOSS2.  I'll check that after the production switch is complete.

.github/CODEOWNERS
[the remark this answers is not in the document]
chan-hoo: @MatthewPyle-NOAA, No, it doesn't have any power because others do not have write access. I remove them just for safety :)

BenjaminBlake-NOAA (approved):

parm/FV3LAM_wflow_nonDA.xml
MatthewPyle-NOAA, line 1: Was the problem with place=excl reproducible?  That is the only change that gives me concern.

chan-hoo: @MatthewPyle-NOAA, On Cactus, both `place=vscatter` and `place=excl` work well. Do you prefer `place=excl`?

MatthewPyle-NOAA: My preference is not to change it if the original place=excl is working okay.

BenjaminBlake-NOAA: @chan-hoo  In that case we should go with whichever option is used in the real-time parallel for consistency.  I'm guessing that's place=excl?

chan-hoo: Done!

MatthewPyle-NOAA: Thanks - will approve and merge now.

MatthewPyle-NOAA (approved):
Looks good to me now.

Issue #76, opened 2023-09-27
update the hash of the rrfs_bridgeloan repo


guoqing-noaa:
update the hash of the rrfs_bridgeloan repo where changes are made to accommodate rrfs-workflow

guoqing-noaa:
Close this issue as the PR was merged.

PR #54, opened 2023-09-04
Cherry-pick commits from RRFS_dev1 workflow

chan-hoo:
## DESCRIPTION OF CHANGES: 

- Cherry-pick the commits from RRFS_dev1 between 8/21/23 - 9/1/23.
- Add temporary fixes to `app_build.sh`.
- Re-organize the DA sample scripts.

## TESTS CONDUCTED: 

- On machines/platforms:

- [ ] WCOSS2
- [x] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [x] Non-DA engineering test
- [x] DA engineering test
- [ ] Other sample scripts:

## ISSUE: 

- Fixes the issue(s) mentioned in #49

PR #77, opened 2023-09-27
update the hash of the rrfs_bridgeloan repo

guoqing-noaa:
## DESCRIPTION OF CHANGES: 

update the hash of the rrfs_bridgeloan repo

## TESTS CONDUCTED: 

- On machines/platforms:

- [ ] WCOSS2
- [ ] Hera
- [x] Orion
- [x] Jet

- Test cases: 

- [x] Non-DA engineering test
- [ ] DA engineering test
- [ ] Other sample scripts:

## ISSUE: 

- Fixes the issue(s) mentioned in #76

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 #*       @defunkt
-*       @MatthewPyle-NOAA @ShunLiu-NOAA @JacobCarley-NOAA @chan-hoo @BenjaminBlake-NOAA
+*       @MatthewPyle-NOAA @ShunLiu-NOAA @JacobCarley-NOAA @chan-hoo @BenjaminBlake-NOAA @guoqing-noaa @hu5970
 
 # Order is important. The last matching pattern has the most precedence.
 # So if a pull request only touches javascript files, only these owners
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 #*       @defunkt
-*       @MatthewPyle-NOAA @ShunLiu-NOAA @JacobCarley-NOAA @chan-hoo @BenjaminBlake-NOAA @guoqing-noaa @hu5970
+*       @MatthewPyle-NOAA @ShunLiu-NOAA @JacobCarley-NOAA @chan-hoo @BenjaminBlake-NOAA
 
 # Order is important. The last matching pattern has the most precedence.
 # So if a pull request only touches javascript files, only these owners
--- a/sorc/Externals.cfg
+++ b/sorc/Externals.cfg
@@ -57,7 +57,7 @@ protocol = git
 repo_url = https://github.com/NOAA-GSL/rrfs_bridgeloan
 # Specify either a branch name or a hash but not both.
 # branch = main
-hash = 7d95e89
+hash = 64ce19d
 local_path = rrfs_bridgeloan
 required = True
 
--- a/sorc/manage_externals/checkout_externals
+++ b/sorc/manage_externals/checkout_externals
@@ -29,10 +29,6 @@ if __name__ == '__main__':
     ARGS = manic.checkout.commandline_arguments()
     try:
         RET_STATUS, _ = manic.checkout.main(ARGS)
-        if os.path.isfile('rrfs_bridgeloan/bridge_loan.sh'):
-            os.system('rrfs_bridgeloan/bridge_loan.sh')
-        if os.path.isfile('../Init.sh'):
-            os.system('../Init.sh')
         sys.exit(RET_STATUS)
     except Exception as error:  # pylint: disable=broad-except
         manic.printlog(str(error))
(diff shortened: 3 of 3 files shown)


chan-hoo:
@guoqing-noaa, the change in this pr has been already include in the rrfs-workflow (but not in the same way). The copy commands to replace several source codes in the weather model with a fixed version can be found in 'sorc/app_build.sh' (https://github.com/NOAA-EMC/rrfs-workflow/blob/b325b2668c543a6272582c5b22f44f13f7d21a95/sorc/app_build.sh#L469). The reason why we put them in app_build.sh rather than `checkout_external` itself is that this is a temporary fix and this change would not be valid for other hashes. You can check the following PR for the details (https://github.com/NOAA-EMC/rrfs-workflow/pull/54/files). @hu5970 also agreed with this approach. Please let us know what you think about this.

chan-hoo:
@guoqing-noaa, regarding the `Init.sh`, there is an if-statement for this in `ush/generate_FV3LAM_wflow.sh` (https://github.com/NOAA-EMC/rrfs-workflow/blob/b325b2668c543a6272582c5b22f44f13f7d21a95/ush/generate_FV3LAM_wflow.sh#L118). So, I think we don't have to run it in the build script `app_build.sh`. Please let us know what you think about this.

MatthewPyle-NOAA:
@guoqing-noaa I don't believe the CODEOWNERS file has the control for this repository.  But I did add you to the "RRFS-devs" team on the NOAA-EMC github (Ming Hu already is there) that we do use to control access.

guoqing-noaa:
> @guoqing-noaa, the change in this pr has been already included in the rrfs-workflow (but not in the same way). The copy commands to replace several source codes in the weather model with a fixed version can be found in `sorc/app_build.sh` (
> 
> https://github.com/NOAA-EMC/rrfs-workflow/blob/b325b2668c543a6272582c5b22f44f13f7d21a95/sorc/app_build.sh#L469
> 
> ). The reason why we put them in app_build.sh rather than `manage_externals/checkout_external` itself is that this is a temporary fix and this change would not be valid for other hashes. You can check the following PR for the details (https://github.com/NOAA-EMC/rrfs-workflow/pull/54/files). @hu5970 also agreed with this approach. Please let us know what you think about this.

Hi @chan-hoo Thanks!
I did not notice that the copy is now done in app_build.sh. I would think it is fine to do copying anywhere.

The only thing is that as you mentioned, this is a temporary fix and this fix may change (slowly) with time. In our previous opinion, it may be better to handle it separately in the bridge_loan.sh without the need to update the app_build.sh. However, updating app_build.sh is not a big deal now. Also, running "bridge_loan.sh" may give users explicit information what files are changed in a temporary fix.  
There are just some miniature considerations, and I don't think we need to make any changes now.

guoqing-noaa:
> @guoqing-noaa, regarding the `Init.sh`, there is an if-statement for this in `ush/generate_FV3LAM_wflow.sh` (
> 
> https://github.com/NOAA-EMC/rrfs-workflow/blob/b325b2668c543a6272582c5b22f44f13f7d21a95/ush/generate_FV3LAM_wflow.sh#L118
> 
> ). So, I think we don't have to run it in the build script `app_build.sh`. Please let us know what you think about this.

Hi @chan-hoo Yes, if "Init.sh" does not run at the checkout_externals time, it will be done at workflow generation time. This is a backup plan at the ufs-srweather-app era. Also at that time, someone may checkout regional_workflow alone and this acts as a safeguard measure to make sure fix files are linked correctly.
It is still better to run "init.sh" right after checkout_externals as one may work on changing fix files directly without a need to generate a workflow.

guoqing-noaa:
> @guoqing-noaa I don't believe the CODEOWNERS file has the control for this repository. But I did add you to the "RRFS-devs" team on the NOAA-EMC github (Ming Hu already is there) that we do use to control access.

Good to know. Thanks, @MatthewPyle-NOAA !

chan-hoo:
@guoqing-noaa, it makes sense to me in case that `regional_workflow` is one of the external components. However, in rrfs-workflow, we don't have to check out the workflow separately as one component (with a hash) any more. I am doubting that it is still a valid approach. I understand there is a case that someone may need to run the `Init.sh` file without generating a workflow. However, even in this case, they don't have to run `manage_externals/checkout_externals` either. Did I miss something?

guoqing-noaa:
> @guoqing-noaa, it makes sense to me in case that `regional_workflow` is one of the external components. However, in rrfs-workflow, we don't have to check out the workflow separately as one component (with a hash) any more. I am doubting that it is still a valid approach. I understand there is a case that someone may need to run the `Init.sh` file without generating a workflow. However, even in this case, they don't have to run `manage_externals/checkout_externals` either. Did I miss something?

@chan-hoo Right, I am still implicitly in the thought of regional_workflow as a separate repo. One can just run Init.sh directly instead of  `manage_externals/checkout_externals`  in the situation I mentioned. Thanks!

guoqing-noaa:
> @MatthewPyle-NOAA, since @guoqing-noaa agreed, we can close this PR without merging.

@chan-hoo I would want to update the hash for the rrfs_bridgeloan repo. rrfs-workflow uses `ufs-weather-model` instead of `ufs_weather_model`.

chan-hoo:
@guoqing-noaa, oops! I am sorry. I didn't catch it. Approving now.

chan-hoo (approved):

MatthewPyle-NOAA (approved):
Thanks to @chan-hoo and @guoqing-noaa for sorting it all out.  Definitely good with a simple hash update!

guoqing-noaa:
@chan-hoo @MatthewPyle-NOAA Thank you very much for the discussions and quick actions!

Issue #92, opened 2023-11-07
v0.7.3 (using this workflow) didn't exactly match v0.7.1 using RRFS_dev1


MatthewPyle-NOAA:
Sounds like some small changes are on hand to correct this issue (reported as a slower running model in v0.7.3, and a lack of fit-to-obs matching between the two systems which are intended to be identical).  Creating an issue to link to PR that should be submitted later today.

MatthewPyle-NOAA:
close

PR #93, opened 2023-11-07
Updates from RRFS_A v0.7.3 retro tests 

hu5970:
The current workflow is tested on Dogwood with retrospective experiments and compared with the GSL RRFS_dev1 tag 0.7.1.
Several small updates needed to match the current RRFS_A v0.7.1:
1) turn on OPM thread by updating model_configure and nems configure.
2) set tg3_from_soil=True for make_ics
3) update configure files for RRFS_A retro runs.
4) changes in prep_cyc for initial Jet/Hera runs.

This PR addresses issue #92 

The retro experiments are conducted on Dogwood and got identical results between v0.7.1 and v0.7.3 for control cycle and ensemble DA cycle in initial/boundary, GSI/EnKF analysis, nonvar cloud analysis, data process, and model forecast restart files.

Also, the wall time for each components are similar to the v0.7.1.


hu5970:
@chan-hoo @MatthewPyle-NOAA Please review this PR. Hopefully, we can build v0.7.3 based on this PR.

parm/nems.configure
```diff
@@ -2,14 +2,6 @@
 ####  NEMS Run-time Configuration File  #####
 #############################################
 
-# ESMF #
```
chan-hoo, line 5: @hu5970, do you have any reason for removing this part. This if-statement can keep the ufs-weather-model printing out a large number of PET* log files.

MatthewPyle-NOAA (approved):
No red flags for me based on a quick look.  But would like to see approval from someone else as well before merging.

parm/nems.configure
hu5970: Yes, this part somehow blocked OMP to use multiple threads. I do not know why but we have to live with a large number of PET* log files for now.

MatthewPyle-NOAA: @chan-hoo Are you okay with it despite this little annoyance?

Issue #15, opened 2023-07-11
Add fire weather functionality into rrfs-workflow


MatthewPyle-NOAA:
The fire weather domain capability was developed using its own production-like workflow.  We need to fold this functionality into the authoritative workflow for RRFS.

BenjaminBlake-NOAA:
I think the approach here would be to incorporate the RRFSFW into the workflow as a 4th cycle type.  It will need to run after the RRFS since it uses the RRFS for ICs/BCs.  The make_grid/orog/sfc_climo tasks have to be run for every cycle since the grid's location changes, but the other jobs (e.g. forecast, post) are similar and won't need to be modified much if at all.

MatthewPyle-NOAA:
This item is waiting on the shift of the RRFS parallel to rrfs-workflow (#??)

PR #95, opened 2023-11-09
Add fire weather grid and make_grid/make_orog/make_sfc_climo tasks for RRFSFW

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This is the first of several small PRs for adding the RRFS fire weather nest into the rrfs-workflow.  In this PR, the firewx grid has been added to the set_predef_gridparams.sh file, and a sample config file has been created.  The config file is based on the non-DA engineering test for WCOSS2; the job sequence is identical except for the addition of the make_grid, make_orog, and make_sfc_climo tasks which are turned on and will need to run for each cycle to generate the fix files.  So I am now thinking a separate xml for RRFSFW is not going to be needed actually.  These 3 tasks have been tested on WCOSS2 and complete successfully.  Here are a few additional comments/questions because I expect I will need to make additional changes before this PR can be merged:

1. GWD needs to be turned off with the RRFS_firewx_1.5km grid.  To do this, I added some logic to setup.sh.  However, the oro_data_ls and oro_data_ss files are still generated as part of the make_orog task, and they are not needed.  Can we turn those off?
2. The ESGgrid DELX and DELY are set to 1501.18 instead of 1500.0 m because delx and dely in regional_grid.nml needs to be exactly 0.00675 in order to match the current RRFSFW configuration.
3. I set ESGgrid_WIDE_HALO_WIDTH = 5 in order to get the correct dimensions for the firewx grid, but most of the other grids have this variable equal to 6.  I don't know if that is an issue or not.
4. Fractional vegetation and soil information needs to be added to the fix files as part of the make_sfc_climo task.  I added the fract_vegsoil_type=.true. option to fort.41, but perhaps this setting should be defined as a config option/variable somewhere.  Let me know where the best place to do this would be.
5. Specific high resolution fix files are needed for the make_sfc_climo task, specifically substrate_temperature.gfs.0.5.nc, soil_type.bnu.v2.30s.nc, and vegetation_type.viirs.v2.igbp.30s.nc.  These files are not part of the "official" fix files directory, at least not yet.  So I am pointing to my own SFC_CLIMO_INPUT_DIR for now until we decide how to handle these fix files.

## TESTS CONDUCTED: 

- On machines/platforms:

- [x] WCOSS2
- [ ] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [ ] Non-DA engineering test
- [ ] DA engineering test
- [x] Other sample scripts: ush/sample_configs/RRFSFW/config.sh_RRFSFW

## ISSUE: 

- Begins to address issue #15


ush/valid_param_vals.sh
```diff
@@ -22,6 +22,7 @@ valid_vals_PREDEF_GRID_NAME=( \
 "GSD_HRRR_AK_50km" \
 "GSD_RAP13km" \
 "RRFS_NA_3km" \
+"RRFS_firewx_1.5km" \
```
chan-hoo, line 25: @BenjaminBlake-NOAA, as we can see above, all the domain (grid) names are in upper case. What do you think about changing the domain name to `RRFS_FIREWX_1.5km`?

scripts/exrrfs_make_sfc_climo.sh
```diff
@@ -105,12 +105,12 @@ cd $workdir
 cat << EOF > ./fort.41
 &config
 input_facsf_file="${SFC_CLIMO_INPUT_DIR}/facsf.1.0.nc"
-input_substrate_temperature_file="${SFC_CLIMO_INPUT_DIR}/substrate_temperature.2.6x1.5.nc"
+input_substrate_temperature_file="${SFC_CLIMO_INPUT_DIR}/substrate_temperature.gfs.0.5.nc"
 input_maximum_snow_albedo_file="${SFC_CLIMO_INPUT_DIR}/maximum_snow_albedo.0.05.nc"
 input_snowfree_albedo_file="${SFC_CLIMO_INPUT_DIR}/snowfree_albedo.4comp.0.05.nc"
 input_slope_type_file="${SFC_CLIMO_INPUT_DIR}/slope_type.1.0.nc"
-input_soil_type_file="${SFC_CLIMO_INPUT_DIR}/soil_type.statsgo.0.05.nc"
-input_vegetation_type_file="${SFC_CLIMO_INPUT_DIR}/vegetation_type.igbp.0.05.nc"
```
chan-hoo, line 113: @BenjaminBlake-NOAA, I have a question for you. Doesn't the above file change (3 files) cause any issues to others than FW?

ush/valid_param_vals.sh
BenjaminBlake-NOAA: Sure!  I'll get that changed.

scripts/exrrfs_make_sfc_climo.sh
BenjaminBlake-NOAA: Yes.  Would it make sense to create a variable name for these 3 files and set default values for them as the current files?

chan-hoo: Yes, it is. Or you can add an if-statement for the specific fire weather domain or a flag for fire-weather.

BenjaminBlake-NOAA: I added an if statement to use these specific files when running with the fire weather grid.  Let me know if that works.

BenjaminBlake-NOAA:
To address 1), I modified exrrfs_make_orog.sh to not generate the oro_data_ls and oro_data_ss GWD files when using the fire weather grid.

2) and 3) are more questions/comments and may not require changes.

To address 4), I modified exrrfs_make_sfc_climo.sh to point to the RRFSFW-specific settings when using that grid.

5) is a temporary solution until a more permanent one is in place for these fix files.

@chan-hoo @MatthewPyle-NOAA  Let me know if you have any other questions or concerns related to this PR.  Thanks

chan-hoo:
It looks good to me. Approving.

chan-hoo (approved):

ush/setup.sh
```diff
@@ -1123,51 +1127,52 @@ files are located.  Currently, this is not the case:
...
   GWD_HRRRsuite_BASEDIR = \"${GWD_HRRRsuite_BASEDIR}\""
+      fi
     fi
-  fi
 #
 # Check that GWD_HRRRsuite_BASEDIR exists and is a directory.
 #
-  if [ ! -d "${GWD_HRRRsuite_BASEDIR}" ]; then
-    print_err_msg_exit "\
+    if [ ! -d "${GWD_HRRRsuite_BASEDIR}" ]; then
```
MatthewPyle-NOAA, line 1135: There are some differences starting here that I think are just formatting differences (from about line 1135 to 1173 in the updated file in a few different blocks).  Could these be cleaned up to make the actual changes clearer?

MatthewPyle-NOAA (changes requested):
One nitpicky request, but otherwise looks good to me.

ush/setup.sh
BenjaminBlake-NOAA: Yeah, those are just formatting differences.  The actual change is simply to add the if [ "PREDEF_GRID_NAME" != "RRFS_FIREWX_1.5km" ] statement at the beginning.  I think what happened is I did not indent all of the commented lines, which is why the changes look so messy.  I can clean that up.

BenjaminBlake-NOAA: I indented the lines that I did not indent previously, but it's still a little tough to discern the actual changes from formatting differences.  Let me know if I should do anything more.  Thanks

MatthewPyle-NOAA: I think you modified more than was necessary - seems like there are more differences than before, but hard to tell now.    Could you try not indenting the comment blocks starting with "Ensure that" at about lines 1144 and 1157?

BenjaminBlake-NOAA: Okay I went ahead and remove the unnecessary indenting and now the changes should be much clearer.  I am used to working with Python code where indenting is important, but in this case it added complexity where it wasn't needed.

MatthewPyle-NOAA (approved):
Will call it good and approve - thanks for indulging me, @BenjaminBlake-NOAA

PR #99, opened 2023-11-16
Add RRFS fire weather grid: make_ics and make_lbcs tasks

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This is the second PR for adding the RRFS fire weather grid to rrfs-workflow.  In this PR, the make_ics and make_lbcs tasks were tested and complete successfully.  The RRFS is added as an input source option, since the fire weather grid is initialized off the RRFS.  The RRFS NA grib2 files cannot be used at this time due to a g2 library issue where files > 2 GB cannot be processed by chgres_cube, so the RRFS CONUS grib2 files are used for now.  Some additional comments about this PR are listed below:

1. A bug in the link_fix.sh script was fixed.  The fvcom_mask.nc file is searched for with all grids, but it is only available for the RRFS_CONUS_3km and RRFS_NA_3km grids.  For other grids, the make_grid task fails when it cannot find the fvcom_mask.nc file.
2. In the make_ics and make_lbcs ex-scripts, external_model is set to 'NAM' for RRFS.  Chgres_cube produces an error if the external_model is not GFS, NAM, RAP, or HRRR, and the error message suggests using external_model='NAM' for other grid options.  This is also how Eric R set up the fire weather nest.  If this solution is not desirable long-term, changes will need to be made to the UFS_UTILS code to support RRFS as an input source.
3. The RRFSFW make_ics task uses 8 nodes with 6 tasks per node.  I am not sure why Eric R settled on these numbers, but if fewer nodes would be more desirable for this task I can run some additional tests.
4. The FV3LAM_wflow_nonDA.xml file is modified so that the make_lbcs task is a metatask for individual forecast hours (one task per hour) instead of one single task for all forecast hours.  I ran a non-DA engineering test on WCOSS2 with these changes and it completed successfully.

## TESTS CONDUCTED: 

- On machines/platforms:

- [x] WCOSS2
- [ ] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [x] Non-DA engineering test
- [ ] DA engineering test
- [x] Other sample scripts: ush/sample_configs/RRFSFW/config.sh_rrfsfw

## ISSUE: 

- Continues to address issue #15 
- Addresses issue #100


ush/get_extrn_mdl_file_dir_info.sh
MatthewPyle-NOAA, line 1: Is this pointing at the CONUS RRFS output a temporary item?  I know the long term hope is to use the full NA domain, but believe that bumps into the 2 GB GRIB2 item.

ush/link_fix.sh
MatthewPyle-NOAA, line 1: Awesome on already having fixed the issue with the fvcom_mask linking for non-3 km grids.

ush/sample_configs/RRFSFW/config.sh_rrfsfw
MatthewPyle-NOAA, line 1: Should the default case inputs be moved from your space to under /lfs/h2/emc/lam/noscrub/UFS_SRW_App/develop/input_model_data/?

MatthewPyle-NOAA (commented):
Had a mix of compliments and concerns, but think we can wrap this one up pretty quickly.  Eager to get it processed to fix the make_grid problem.

ush/get_extrn_mdl_file_dir_info.sh
BenjaminBlake-NOAA: Yes, initializing off the RRFS CONUS grib2 output is temporary until the issue with using the full NA domain is fixed.

ush/sample_configs/RRFSFW/config.sh_rrfsfw
BenjaminBlake-NOAA: That's a good idea, sure.  I don't have permission to add files to that directory because @chan-hoo owns it so I will likely need his help to do this.

MatthewPyle-NOAA (approved):
Looks good to me, but would be nice to get the RRFS files used for the firewx sample case into the same location as other sample initializing model data.

MatthewPyle-NOAA:
Can leave the shift of the new RRFS data to /lfs/h2/emc/lam/noscrub/UFS_SRW_App/develop/input_model_data/ for a subsequent PR.  Want to get this one in to fix the make_grid problem.

Issue #114, opened 2023-11-20
simplify launch_FV3LAM_wflow.sh


guoqing-noaa:
Currently, launch_FV3LAM_wflow.sh is a little bit heavy to run by cron jobs.

When launched by cron each time, this script sources var_defns.sh once, run rocotorun once, and rocotostats twice.
It also generates lots of log outputs that are often NOT used by real-time runs.

For example, in recent RTMA real-time runs on Jet, this script generates 10G+ log files under` $EXPDIR` in less than 2 days. This yielded an out-of-space situation and stopped all RRFS/RTMA and other real-time runs on Jet.

At the same time, the rocoto running status is already logged by the` FV3LAM_wflow_c3v68.log` files under `com/logs/rrfs.YYYYMMDD/HH`

I will create a PR to make a very lightweight launch_FV3LAM_wflow.sh and hope this will also reduce system load.

PR #115, opened 2023-11-21
 simplify launch_FV3LAM_wflow.sh

guoqing-noaa:
## DESCRIPTION OF CHANGES: 

- simplify launch_FV3LAM_wflow.sh
- hotfix on Init.sh

## TESTS CONDUCTED: 

- On machines/platforms:

- [ ] WCOSS2
- [x] Hera
- [ ] Orion
- [x] Jet

- Test cases: 

- [x] Non-DA engineering test
- [ ] DA engineering test
- [x] Other sample scripts:  Realtime runs

## ISSUE: 

- Fixes the issue(s) mentioned in #114

### DIFF
--- a/ush/Init.sh
+++ b/ush/Init.sh
@@ -6,7 +6,7 @@ utildir="$basedir"/bash_utils
 # set the pre-commit hook
 echo -e "#!/bin/sh" > $basedir/../.git/hooks/pre-commit
 chmod +x $basedir/../.git/hooks/pre-commit
-echo -e "bash_utils/preCommit">>$basedir/../.git/hooks/pre-commit
+echo -e "ush/bash_utils/preCommit">>$basedir/../.git/hooks/pre-commit
 
 $utildir/make_agent_link_for_fix
 touch $basedir/../fix/INIT_DONE
--- a/ush/generate_FV3LAM_wflow.sh
+++ b/ush/generate_FV3LAM_wflow.sh
@@ -597,15 +597,25 @@ $settings"
 #
 #-----------------------------------------------------------------------
 #
-print_info_msg "Generating the workflow launch script
-${EXPTDIR}/${WFLOW_LAUNCH_SCRIPT_FN}"
+print_info_msg "
+Creating symlink in the experiment directory (EXPTDIR) that points to the
+workflow launch script (WFLOW_LAUNCH_SCRIPT_FP):
+  EXPTDIR = \"${EXPTDIR}\"
+  WFLOW_LAUNCH_SCRIPT_FP = \"${WFLOW_LAUNCH_SCRIPT_FP}\""
+ln -fs "${WFLOW_LAUNCH_SCRIPT_FP}" "$EXPTDIR"
 
-cat > ${EXPTDIR}/${WFLOW_LAUNCH_SCRIPT_FN} <<EOF
-#!/bin/bash
-module load rocoto
-rocotorun -w "${WFLOW_XML_FN}" -d "${WFLOW_XML_FN%.*}.db"
-EOF
+print_info_msg "Generating an alternate simple launch script
+${EXPTDIR}/run_rocoto.sh"
 
+echo "#!/bin/bash" > ${EXPTDIR}/run_rocoto.sh
+if [[ $MACHINE == "wcoss2" ]] ; then
+  echo "module use /apps/ops/test/nco/modulefiles" >> ${EXPTDIR}/run_rocoto.sh
+  echo "module load core/rocoto/${rocoto_ver}" >> ${EXPTDIR}/run_rocoto.sh
+else
+  echo "module load rocoto" >> ${EXPTDIR}/run_rocoto.sh
+fi
+echo "rocotorun -w ${WFLOW_XML_FN} -d ${WFLOW_XML_FN%.*}.db" >> ${EXPTDIR}/run_rocoto.sh
+chmod +x ${EXPTDIR}/run_rocoto.sh
 #
 #-----------------------------------------------------------------------
 #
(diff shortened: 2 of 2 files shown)


ush/generate_FV3LAM_wflow.sh
```diff
@@ -597,12 +597,15 @@ $settings"
...
-ln -fs "${WFLOW_LAUNCH_SCRIPT_FP}" "$EXPTDIR"
+print_info_msg "Generating the workflow launch script
+${EXPTDIR}/${WFLOW_LAUNCH_SCRIPT_FN}"
+
+cat > ${EXPTDIR}/${WFLOW_LAUNCH_SCRIPT_FN} <<EOF
+#!/bin/bash
+module load rocoto
+rocotorun -w "${WFLOW_XML_FN}" -d "${WFLOW_XML_FN%.*}.db"
+EOF
+
```
chan-hoo, line 608: @guoqing-noaa, I understand that a huge log file is generated for a DA run by the launch script. The rocoto xml file for a DA run is incomplete. There are so many tasks unlaunched each cycle. Some useful capabilities of the launch script are not used in a DA run. However, this script is very useful for a non-DA run. So I suggest you to put an if-statement here like
```
if [ "${DO_NON_DA_RUN}" = "TRUE" ]; then
  (current link)
else
  (your new script)
fi

File: ush/Init.sh [hunk 2]
```diff
@@ -6,7 +6,7 @@ utildir="$basedir"/bash_utils
 # set the pre-commit hook
 echo -e "#!/bin/sh" > $basedir/../.git/hooks/pre-commit
 chmod +x $basedir/../.git/hooks/pre-commit
-echo -e "bash_utils/preCommit">>$basedir/../.git/hooks/pre-commit
+echo -e "ush/bash_utils/preCommit">>$basedir/../.git/hooks/pre-commit
```
chan-hoo, line 9: @guoqing-noaa, thanks for this fix! I was looking for this.

ush/generate_FV3LAM_wflow.sh
chan-hoo: @guoqing-noaa, in addition, `module load rocoto` doesn't work on WCOSS2. Please check `modulefiles/wflow_wcoss2.lua`.

MatthewPyle-NOAA:
Is it worth splitting this into 2 PRs?  Seems like the Init.sh fix is something we want committed quickly, while the launch script work may need more work.

ush/generate_FV3LAM_wflow.sh
chan-hoo: @guoqing-noaa, I am looking through the launch script. As I mentioned above, it has some useful functions can be used when a rocoto xml file is complete. So my new suggestion is: how about generating a new script with a new name without removing the current link? I'll update the launch script later.

chan-hoo:
@MatthewPyle-NOAA, yes I agree with you. The fix is urgent, but we'll need more discussion for the launch script.

ush/generate_FV3LAM_wflow.sh
guoqing-noaa: @chan-hoo Thanks for the suggestion. It is a good idea to use an alternate name for a simplified version. I will make the change soon.

guoqing-noaa: Regarding the heavy launch script, I have alternate thoughts:
1. Users can always run rocotostat to check task status manually. So running rocotostat twice in each crob job is not a must. If the cycledef spans a few months (for example, 4 months in my RTMA run situation), the log file will grow very big in size quickly.
2. var_defns.sh is a huge file with 2991 lines. Sourcing it in each cron job but only using 3 or 4 variables is not that efficient for a cron job.
3. I understand that rocotorun may fail to submit some tasks due to various issues. However, the `FV3LAM_wflow_*.log` file may already provide some useful information for debugging.  If an issue is transient, we may not need to or may not be able to address it. If an issue is persistent, we can always debug based on the current cycle runs.

Just my two cents. Thanks!

guoqing-noaa:
@chan-hoo I made changes based on the feedback from you and Matt. 
The simplified version is now named "run_rocoto.sh" and the original launch script is still linked.

I don't have access to wcoss2 so I cannot test the changes related to wcoss2. Could you help me test it on wcoss2?
Thanks!

ush/generate_FV3LAM_wflow.sh
```diff
@@ -603,6 +603,19 @@ workflow launch script (WFLOW_LAUNCH_SCRIPT_FP):
   EXPTDIR = \"${EXPTDIR}\"
   WFLOW_LAUNCH_SCRIPT_FP = \"${WFLOW_LAUNCH_SCRIPT_FP}\""
 ln -fs "${WFLOW_LAUNCH_SCRIPT_FP}" "$EXPTDIR"
+
+print_info_msg "Generating an alternate simple launch script
+${EXPTDIR}/run_rocoto.sh"
+
+echo "#!/bin/bash" > ${EXPTDIR}/run_rocoto.sh
+if [[ $MACHINE == "wcoss2" ]] ; then
```
chan-hoo, line 611: @guoqing-noaa, it should be uppercases: wcoss2 => WCOSS2

guoqing-noaa: @chan-hoo Thanks a lot for catching this.  
I changed it to:
`
if [[ "${MACHINE,,}" == "wcoss2" ]] ; then
`
Hope it is a safer way and you are okay with it.

chan-hoo:
The non-DA eng test was completed successfully on wcoss2.

chan-hoo (approved):

PR #142, opened 2023-12-04
updates from Cactus and Hera for v0.7.7

hu5970:
1) update compilingsetups for Hera. Using spack-stack 1.5.0.
2) change recenter vaiable list to 19.
3) turn off GSI code start run for Hera/Jet.
4) change fit files name from rrfs_a to rrfs.

Tested on Cactus for v0.7.7 and Hera for v0.7.7


scripts/exrrfs_run_analysis.sh
MatthewPyle-NOAA, line 1: With this change, isn't the comment valid for non-WCOSS2 runs as well?

scripts/exrrfs_run_recenter.sh
MatthewPyle-NOAA, line 1: I feel like this list keeps getting changed - are we confident in the variable list?

ush/sample_configs/RRFS_A/config.sh_rrfs_a_enkf_n3
MatthewPyle-NOAA, line 1: One comment - we don't need to use the quilted restart functionality within the DA system if we don't want to.  I don't think any benefit is gained unless the forecast extends beyond the time the restart file is written.

MatthewPyle-NOAA (commented):
A few comments and questions - most important might be trying to clarify if we want to be using the quilted restart functionality with the EnKF forecasts.  Seems like it is impacting the required resources, and shouldn't be necessary.

scripts/exrrfs_run_analysis.sh
hu5970: This is a but fix. the CALC_ENSMEAN_TN is using the same executable as recentering. So it should use the same module as recentering.

scripts/exrrfs_run_recenter.sh
hu5970: This change is request from Shun based on v0.7.5 on dogwood. The rrfs should be official name for the system.

ush/sample_configs/RRFS_A/config.sh_rrfs_a_enkf_n3
hu5970: Agree. Seems the quilted restart is turned on automatically. It does not impact anything so far.

guoqing-noaa: @hu5970 we may turn off quilting in config_defaults.sh
Currently it is set as `QUILTING="TRUE"`

MatthewPyle-NOAA: @guoqing-noaa I think that is for history output (which we do want quilted).  Guessing we need a fresh config option to handle quilted restarts.  @hu5970 if you want to wait to clean up this issue later, I'll approve so we can move on with the multi-phys ensemble.

hu5970: We need to add QUILTING_RESTART="FALSE" to model configure. But I would like to clean up this issue later.

MatthewPyle-NOAA (approved):
Approving to keep v0.7.7 going - will clean up quilted restart issue later.

hu5970:
@MatthewPyle-NOAA Yes, Please squash merge it.

Issue #150, opened 2023-12-06
Launch log file becomes too big in multiple-day run


chan-hoo:
- Currently, the launch log file "log.launch_FV3LAM_wflow" stores the whole history of the launch commands.
- In a multiple-day run, its size becomes huge unnecessarily.

PR #151, opened 2023-12-06
Change append to replace for launch log file

chan-hoo:
## DESCRIPTION OF CHANGES: 

- The log file for the launch script `launch_FV3LAM_wflow.sh` does not append the new log to the existing log file. Instead, the latest log is only stored in the log file.

## TESTS CONDUCTED: 

- On machines/platforms:

- [ ] WCOSS2
- [x] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [x] Non-DA engineering test
- [ ] DA engineering test
- [ ] Other sample scripts:

## ISSUE: 

- Fixes the issue(s) mentioned in #150

### DIFF
--- a/ush/launch_FV3LAM_wflow.sh
+++ b/ush/launch_FV3LAM_wflow.sh
@@ -228,7 +228,7 @@ Output of rocotostat_cmd is:
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ${rocotostat_output}
-" >> "${WFLOW_LAUNCH_LOG_FN}" 2>&1
+" > "${WFLOW_LAUNCH_LOG_FN}" 2>&1
 #
 #-----------------------------------------------------------------------
 #


MatthewPyle-NOAA (commented):
Will there ever be a desire to see more than the most recent rocoto item in the log file?  Was wondering if it might be better to hold on to the most recent 1000 (or 10000) lines by appending as was done before, but then doing a "tail -n $lines logfile" to prevent it from getting incredibly large.

guoqing-noaa:
> @guoqing-noaa, can you test this PR? If you like this change, we can remove `run_rocoto.sh`. If you want to have the file though, it would be fine.

@chan-hoo If possible, could we keep "run_rocoto.sh"?

Regarding the heavy launch script, I have alternate thoughts:

- Users can always run rocotostat to check task status manually. So running rocotostat twice in each crob job is very unefficient. If the cycledef spans a few months (for example, 4 months or 1 year), it will add lots of unnecessary load burden to corn jobs.

- var_defns.sh is a huge file with 2991 lines. Sourcing it in each cron job but only using 3 or 4 variables is not that efficient for a cron job as well

- I understand that rocotorun may fail to submit some tasks due to various issues. However, the FV3LAM_wflow_*.log file may already provide some useful information for debugging. If an issue is transient, we may not need to or may not be able to address it. If an issue is persistent, we can always debug based on the current cycle runs.

Just my two cents. Thanks!

chan-hoo:
@guoqing-noaa, no objection at all.

chan-hoo:
@ShunLiu-NOAA, the first rocotostat is to find a "DEAD" task. Once it is detected, the crontab command will be removed automatically.

MatthewPyle-NOAA:
Thanks for that explanation, @chan-hoo.  Seems like the lightweight one is best for real experiments, while the heavy weight one using rocotostat is good for regression testing.  Is that fair?

chan-hoo:
@MatthewPyle-NOAA, I agree. I'll close this PR.

PR #163, opened 2023-12-13
Update scripts for running retros based on v0.7.7 on Hera and Jet.

hu5970:
## DESCRIPTION OF CHANGES: 
1) reset data paths to point to new retro data
2) Use the latest GSI
3) Turn on smoke/dust and GLM on retro cases
4) Turn on multiphysics  ensemble 
5) Update workflow setup on Jet.

## TESTS CONDUCTED: 
- [x ] WCOSS2
- [ x] Hera
- [ ] Orion
- [ x] Jet

- Test cases: 
retrospective cycle


sorc/CMakeLists.txt
```diff
@@ -252,6 +252,7 @@ if (BUILD_UPP)
     "-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}"
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
     "-Dexec_dir=${CMAKE_INSTALL_BINDIR}"
+#    "-DINTERNAL_IFI=ON"
```
chan-hoo, line 255: @hu5970, this line is commented out. Do we need to keep this line? If not, please remove this line.

MatthewPyle-NOAA: @chan-hoo I think it is there to uncomment in case the UPP needs to be rebuilt with IFI enabled.

ush/sample_configs/others/config.sh.RRFS_CONUS_3km_ens
MatthewPyle-NOAA, line 1: Should the TAG definition reflect that it is v0.7.7?

`TAG="c3enfcst76"`

should it be?

`TAG="c3enfcst77"`

MatthewPyle-NOAA (commented):
One small question, but think we can get this one in quickly and tag it as v0.7.7

sorc/CMakeLists.txt
hu5970: I need to have this line here but commented out because I do not want to find this line from UPP building scripts each time when we compile the RRFS with IFI on. It will be uncommented when it goes to operation.

ush/sample_configs/others/config.sh.RRFS_CONUS_3km_ens
MatthewPyle-NOAA: Got it...thanks.

MatthewPyle-NOAA (approved):

sorc/CMakeLists.txt
chan-hoo: Got it.

chan-hoo (approved):

PR #173, opened 2023-12-20
Update model hash and workflow to match the model change in output setup

hu5970:
Update configure files for RRFS_A/Retro/RRFS_B to match the new model hash.
    Update version number to 0.7.9
    Fixed a bug in setup configure.

Tested on WCOSS2 with v0.7.9 and on Hera/Jet with winter retros.


MatthewPyle-NOAA (approved):
Looks good to me - will merge later today if there are no objections.

ush/set_rrfs_config_general.sh
```diff
@@ -1,7 +1,8 @@
 
 if [[ $MACHINE == "jet" ]] ; then
+  ACCOUNT=nrtrr
   if [[ -n $RESERVATION ]] ; then
-    ACCOUNT="${ACCOUNT:-rtrr}"
+    ACCOUNT=rtrr
```
guoqing-noaa, line 5: Why do we change from `ACCOUNT="${ACCOUNT:-rtrr}` to a hardwired one? The previous one will be more flexible.

ush/set_rrfs_config_general.sh
```diff
@@ -28,7 +29,7 @@ if [[ $MACHINE == "jet" ]] ; then
   fi
 
 elif [[ $MACHINE == "hera" ]] ; then
-  ACCOUNT="${ACCOUNT:-wrfruc}"
+  ACCOUNT="zrtrr"
```
guoqing-noaa, line 32: similar question here

ush/set_rrfs_config_general.sh
hu5970: It can not get the right account if the configure set it wrong. This will use the right account for now. Need a better way to set the account.

hu5970:
@MatthewPyle-NOAA Could you merge this PR if there no other questions? Ming

PR #177, opened 2023-12-21
Update model hash to 991d652 (12/21/2023).

hu5970:
This hash has several bug fixes for RRFS model. It is tested in RRFS_A on Dogwood in past 18-h cycles.


MatthewPyle-NOAA (approved):
I know there is an eagerness to update the model, so approving instantly.

PR #179, opened 2023-12-22
Update retro configure to use the GDAS 6-h ensemble to initialize the ensemble DA

hu5970:
We should set the following in all retro configure files:
1) use the GDAS 6-h ensemble to initialize the ensemble DA
2) update the soil surgery setup
3) DO_POST_PROD set to true for ensemble DA retro


chunhuazhou (approved):
Looks good. Approving.

MatthewPyle-NOAA (approved):
Looks safe to me...approving.

PR #197, opened 2024-01-17
update model hash and compilers to Spack stack 1.5.1

hu5970:
The model is updated to hash use e7380dd (01/17/2024).
Update compilers to use Spack-stack 1.5.1 for Herea/Jet/Orion/Hercules

This version was tested on Hera and Jet with several cycles from winter 2022 retro.

### DIFF
--- a/modulefiles/build_hera_intel.lua
+++ b/modulefiles/build_hera_intel.lua
@@ -12,6 +12,9 @@ load(pathJoin("cmake", os.getenv("cmake_ver") or "3.23.1"))
 
 load("rrfs_common")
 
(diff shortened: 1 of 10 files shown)


MatthewPyle-NOAA (approved):
Nice on the addition of a rrfs_common that can be used by many builds.  And also see that you updated the model to the latest, including Sam's hailcast bugfix.  So approving.

guoqing-noaa:
@hu5970 FYI, if we move to spack-stack 1.6.0 in the future, we don't need the separate prod_util package as its changes were already included in v1.6.0. 
`/scratch1/NCEPDEV/nems/role.epic/spack-stack/spack-stack-1.6.0/envs/gsi-addon-dev/install/modulefiles/Core`

PR #209, opened 2024-01-26
Update model hash

hu5970:
This model hash includes supplementary physics updates (MYNN, GF,RUC_LSM, smoke and dust) for RRFS code freeze.


MatthewPyle-NOAA (approved):
Always happy to approve hash updates...

PR #210, opened 2024-01-29
Extend the forecast from 12h to 42h for 00/06/12/18 cycle.

hu5970:
For Beta evaluation retros with CONUS 3km grid, extend the forecast from 12h to 42h for 00/06/12/18 cycle.


MatthewPyle-NOAA (approved):
Won't hold this one up...approving.

Issue #222, opened 2024-02-05
The variable name DO_ENVAR_RADAR_REF_ONCE does not hint at its use


AmandaBack-NOAA:
The variable DO_ENVAR_RADAR_REF_ONCE is used to control whether radar reflectivity is assimilated simultaneously with other observations through SDL/VDL in GSI, but the variable name does not mean that. "Do EnVar radar ref once" means "do EnVar radar ref one time," not "do EnVar radar ref simultaneously with other observations." The variable currently doesn't even appear in the RRFS CONUS config.sh, so the user has to do a lot of digging in the workflow generation and other scripts to find that this variable exists and then to discover its meaning.

PR #223, opened 2024-02-05
Add variable DO_ENVAR_RADAR_REF_ONCE with brief description to RRFS CONUS config file

AmandaBack-NOAA:
## DESCRIPTION OF CHANGES: 
Add a line
DO_ENVAR_RADAR_REF_ONCE="TRUE" (its current default) to the RRFS_CONUS_3km config.sh with a comment explaining what it is for, so people building retros know how to turn SDL/VDL and hybrid radar ref task on and off

## TESTS CONDUCTED: 
Built and ran successfully with the variable set to both TRUE and FALSE

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ X] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
- Addresses the issue(s) mentioned in #222


MatthewPyle-NOAA (dismissed):
Looks good from my perspective, but if @hu5970 or @ShunLiu-NOAA could confirm from a DA perspective that would be great.

ShunLiu-NOAA:
Looks good to me.

shoyokota:
@AmandaBack-NOAA Originally I introduced DO_ENVAR_RADAR_REF_ONCE as the variable to be set indirectly depending on ngvarloc in ush/set_rrfs_config_SDL_VDL_MixEn.sh, so in my opinion, we don't necessarily need this modification. If it is not understandable, it may be good to add the comment in ush/set_rrfs_config_SDL_VDL_MixEn.sh.

AmandaBack-NOAA:
@shoyokota Is it your position that someone who wants to run the RRFS workflow without SDL/VDL and have the hybrid radar reflectivity run as a separate task, should do so by going into set_rrfs_config_SDL_VDL_MixEn.sh and deleting or changing the values of ngvarloc and nsclgrp? If so, how is that communicated to the user? It seems esoteric to me.

shoyokota:
@AmandaBack-NOAA Yes. We can set 2-scale SDL by setting nsclgrp=2 and 2-variable VDL by setting ngvarloc=2 in set_rrfs_config_SDL_VDL_MixEn.sh. If we don't want to apply neither SDL not VDL, we set nsclgrp=1 and ngvarloc=1. In this implementation, DO_ENVAR_RADAR_REF_ONCE is set automatically in set_rrfs_config_SDL_VDL_MixEn.sh. This 2-scale VDL is only for simultaneous radar reflectivity DA, so I don't think we need to set DO_ENVAR_RADAR_REF_ONCE independently from ngvarloc. 

To set nsclgrp and ngvarloc in set_rrfs_config_SDL_VDL_MixEn.sh is based on @hu5970's [suggestion](https://github.com/NOAA-GSL/regional_workflow/pull/451#discussion_r1099404014), so @hu5970 may also have some opinions, but if we should set them in config.sh, it may be good to move nsclgrp and ngvarloc to just before calling set_rrfs_config_SDL_VDL_MixEn.sh in config.sh.

AmandaBack-NOAA:
@chunhuazhou set_rrfs_config_SDL_VDL_MixEn.sh does not need to be added to config.sh for RRFS_CONUS_3km. It is there by default in the sample config and has been since before the GSL retros switched to using SDL/VDL.
If the intended use is to comment out or delete set_rrfs_config_SDL_VDL_MixEn.sh to run without SDL/VDL, I would be fine with a line indicating that in the RRFS_CONUS_3km sample config.
My solution was to add the line DO_ENVAR_RADAR_REF_ONCE="TRUE" to the config so that by default the setting doesn't change, but it's there if someone wants to change it. Either is fine with me. But the method to toggle SDL/VDL on and off does not need to be obscure, as far as I know.

ush/sample_configs/others/config.sh.RRFS_CONUS_3km
```diff
@@ -20,6 +20,7 @@ DO_POST_PROD="TRUE"
 DO_RETRO="TRUE"
 DO_NONVAR_CLDANAL="TRUE"
 DO_ENVAR_RADAR_REF="TRUE"
+DO_ENVAR_RADAR_REF_ONCE="TRUE" # set this to TRUE to turn on SDL/VDL and assimilate radar ref with other obs
```
chunhuazhou, line 23: Setting this to TRUE doesn't turn on SDLVDL, but to assimilate radar ref with other obs when SDLVDL is on

AmandaBack-NOAA:
Good point @chunhuazhou. Actually I forgot I made this yesterday and made other changes in my fork today, so I need to close this anyway. I hope someone else will find a solution to issue #222 because it still stands.

Issue #149, opened 2023-12-06
[main] date should be replaced with NDATE of prod_util


chan-hoo:
The bash `date` call in the scripts should be replaced with the `$NDATE` call of `prod_util` for EE2 compliance.

MatthewPyle-NOAA:
Just to clarify this issue, it really only applies to date definitions that are +/- a certain amount of time from a certain time.  This really is the job of $NDATE.    Converting some items in the RRFS ex-scripts from date to $NDATE will add numerous lines to the script (date has some slick abilities for formatting of character strings that include date/time information).

BenjaminBlake-NOAA:
@MatthewPyle-NOAA Does $NDATE only work if the date format is YYYYMMDDHH?  That is what it seems like to me.  If that is the case, then a lot of the instances of date definitions +/- a certain amount of time from a certain time in the ex-scripts can't use $NDATE instead of date, because many of them use dates in different formats (e.g. YYYY-MM-DD, YYYYMMDD, YYYYJJJDD, etc.).  I can change the ones in YYYYMMDDHH format if that is desirable.

MatthewPyle-NOAA:
@BenjaminBlake-NOAA Yes, pretty sure that is the case.  That was what I meant with my Feb 22 comments that it would add lines bringing the date into a format useable by $NDATE, and then more lines to pull out the date information from what $NDATE provided into the form needed for the rest of the script.

BenjaminBlake-NOAA:
Understood, thanks!  There are a few instances where $NDATE can be used instead of date, so I'll go ahead and modify those.

guoqing-noaa:
Sorry for my jumping into this discussion. I am curious when we can use the `date` shell command.

I met a similar situation when adding the MPAS+JEDI component into the workflow.
  
For example:
The MPAS model namelist will need the following time string:
`config_start_time = '2024-05-27_03:00:00'`

My current solution is to use `${NDATE}` for all situations where we need to find previous or next cycles and only use `date` to generate a formatted time string (treating `date` as a USH utility). For example:  
```
...
CDATElbc=$($NDATE -${offset} ${CDATE})
...
...
start_time=$(date -d "${CDATE:0:8} ${CDATE:8:2}" +%Y-%m-%d_%H:%M:%S) 
...
...
# copy output to COMOUT
CDATEp1=$($NDATE 1 ${CDATE})
[...]

PR #248, opened 2024-02-22
Change date to ndate in all J-jobs

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR partially addresses issue #149 .  All instances of date in the J-jobs were changed to $NDATE, and some irrelevant comments were removed from the top of JRRFS_BUFRSND.  The ex-scripts also invoke "date" in many different ways, and eventually we may want to modify some of those formatting calls to use $PDY and $cyc.  But I think that is outside the scope of this PR, and @MatthewPyle-NOAA recommended leaving those instances of date alone for now.  No tests were run with these changes because the prod_util module is loaded for all jobs, which means the ndate function will always be available.  Unless there is a RDHPCS machine listed below where ndate is not available, and that may break functionality there.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Partially fixes the issue(s) mentioned in #149


MatthewPyle-NOAA:
Hi Ben - I didn't realize these were simple date executions within the J-job.  Looking at a variety of operational systems, it does seem to be the norm for there to be "date" calls at the beginning and end of the script.   date will provide more time information than $NDATE for this use case.

I think Chan-Hoo's issue was more about the ex-scripts, which will be a lot tougher (and some won't make sense to change).  Long ago NCO seemed okay with some use of the date utility as long as it was defining or using more typical NCO date variables, but the AQM ex-scripts appear to be using $NDATE for their much more limited date manipulation.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA  Ahh ok that's fine, yeah modifying the ex-scripts will be a much more involved effort.  At some point I can look at how the AQM ex-scripts are handling things and see if we can adapt any of that logic to RRFS.

PR #249, opened 2024-02-23
Updates for  v0.8.3

hu5970:
1) Update the GSI hash to the current version
2) Update model hash to point to release branch
3) Fix small bugs in v0.8.3 test:
    1) update UPP control and NT file
    2) Use TPP_RUN_ANALYSIS in exrrfs_run_analysis.sh

Tested with retro and v0.8.3 on Dogwood.


sorc/Externals.cfg
MatthewPyle-NOAA, line 1: Only thing I was unsure of was the hash being specified for the weather model.  Seems to be going away from the level that the release branch was established (10974d9) - and I'm not finding the aeea708 hash among recent commits.

MatthewPyle-NOAA (commented):
Will be ready to approve once the question about the weather model hash is clarified.

sorc/Externals.cfg
MatthewPyle-NOAA: @hu5970 Making another comment in case you don't see these unless your name is mentioned....

hu5970: @MatthewPyle-NOAA The release branch has its own Hash now.
Screenshot 2024-02-23 at 10 03 10 AM

This is the hash match the develop branch (10974d9). We will update hash number based on release branch. I prefer to use hash for better control of version used in the operation. We can use the release branch name when we do not need to tune model (only bug fix).

hu5970: I checkout develop branch (10974d9) and hash aeea708. They have identical code.

hu5970:
@MatthewPyle-NOAA Please let me know if you still cannot find hash aeea708. Thanks, Ming

sorc/Externals.cfg
MatthewPyle-NOAA: Understood - didn't think that you were pointing at the release branch hash.

MatthewPyle-NOAA (approved):
Looks good...approving.

PR #250, opened 2024-02-26
Fix several bugs in renaming anal to analysis, and to point at local FMS build

hu5970:
Fix several bugs in renaming anal to analysis.
Update configure for v0.8.3 RRFS_A.

Tested with the RRFS_A v0.8.3 on Dogwood.


MatthewPyle-NOAA (approved):
Want to confirm that you want the IFI and GTG builds uncommented within CMakeLists.txt before merging, but looks good to me.

BenjaminBlake-NOAA (approved):
@hu5970  Thanks for catching the couple of instances I missed of JRRFS_RUN_ANAL in the ex-script and workflow xml files.

hu5970:
@MatthewPyle-NOAA Just want to enforce the IFI and GTG in real-time runs. It may cause trouble for some retro runs. But we will tell our colleagues what to do with retros.

PR #253, opened 2024-02-27
Turn off IFI/GTG by default to fix UPP compilation

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- Building the UPP source code with rrfs-workflow currently fails with -DINTERNAL_IFI and -DBUILD_WITH_GTG set to "ON" by default.  This change was recently made in PR #250, and it also causes the compilation to fail for many other source codes.  We do not want these compilations to fail by default, so I recommend commenting out these two settings.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [x] Others: Built the source code successfully on Dogwood.


MatthewPyle-NOAA:
@BenjaminBlake-NOAA I'm pondering this one a bit.  Seems like there is no option that makes it easy for people without IFI/GTG access, but also prevents accidental omission of compiling these things for real-time or retro experiments.  The recent commit made things better for the latter, but worse for the former.

hu5970:
Yes, there is no easy solution for this one. Someone has to manually turn on/off those in CMakefile. I would like to make it clear requirement that anyone use the rrfs-workflow to manually turn off (comment out) those two lines.

MatthewPyle-NOAA:
Would it be crazy to have two versions of the CMakeFiles.txt file, and have an initial step forcing people to select the correct one to use?

BenjaminBlake-NOAA:
I like that idea.  That would force the user to select one option or the other, adding an extra step, but it would ensure that the compilation wouldn't fail if someone forgot to comment out those lines.  @hu5970 what do you think?

hu5970:
Agree. Or should we add options to appbuild.sh to turn off the GTG and IFI? UPP build has such options.

BenjaminBlake-NOAA:
@hu5970  That's a good idea too.  I'll take a look at how the UPP build does that.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA @hu5970  I added an option to app_build.sh to turn off the GTG/IFI build.  It is set to true by default, so invoking this option turns off building GTG/IFI.  Though I wonder if it should be the other way around, because intuitively a --gtg option makes me think one would want to build with GTG/IFI.  I'm happy to modify this.

sorc/app_build.sh
MatthewPyle-NOAA, line 1: Could the --gtg option be changed to --nogtg?  That seems more intuitive to me.

MatthewPyle-NOAA (changes requested):
One small request to change if you agree, then will approve and merge.

sorc/app_build.sh
BenjaminBlake-NOAA: Good idea.  Let me do that now.

MatthewPyle-NOAA (approved):
Thanks for this fix, @BenjaminBlake-NOAA.  Approving now...

PR #260, opened 2024-03-01
Add the following functions for RRFS_A v0.8.5:

hu5970:
1) Use fvcom data directory
2) Write restart file every 12-h in long forecast
3) turn on GLM in GSI
4) update configure to run blending and GLM.

Those changes are tested with v0.8.5 on Cactus/RRFS_B/retro.


modulefiles/tasks/wcoss2/save_restart.local.lua
MatthewPyle-NOAA, line 1: Is there a reason why this one goes away from using the _ver definitions to hardcoded versions for this task lua file?  run_prepstart.local.lua went in the other direction.  For WCOSS2, would definitely like them using the values pulled from versions/run.ver.

MatthewPyle-NOAA (commented):
One concern with a modulefile...otherwise should be able to get it merged this weekend.

modulefiles/tasks/wcoss2/save_restart.local.lua
hu5970: The run_prepstart.local.lua use the _ver define because it does not preporcess FVCOM using ncap2 now. But the save_restart now needs to run ncap2. To solve this problem, the nco need to be updated to include ncap2.

MatthewPyle-NOAA (approved):
Folding it in...

PR #273, opened 2024-03-12
Update workflow to do retros on WCOSS2 with RRFS_A configuration.

hu5970:
Tested with Spring 2023 retro on Cactus.


MatthewPyle-NOAA (approved):
Looks safe to me...approving.  

But I was curious about the 
`elif [[ $MACHINETYPE == "person" ]] ; then` 
block added in set_rrfs_config_general.sh.  What is the person machinetype?

Issue #97, opened 2023-11-09
extrn_ics/lbcs tasks need to be integrated into make_ics/lbcs


chan-hoo:
- The two tasks `extrn_ics` and `extrn_lbcs` should be removed and their functions should be integrated into the `make_ics` and `make_lbcs` tasks, respectively.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA @ShunLiu-NOAA  I am currently working on this issue for the RRFS workflow.  For the GET_GEFS_LBCS_FILES task, which is used by the RRFS ensemble, I will try to incorporate it into the make_lbcs task with an if statement like this one:
if [ extrn_mdl_name != GEFS ]; then
  retrieve LBCs as normal
elif [ extrn_mdl_name = GEFS ]; then
  retrieve GEFS files and do time interpolation using wgrib2, which is currently done as part of GET_GEFS_LBCS_FILES
fi

I think this should work because most of the script logic in GET_EXTRN_LBCS and GET_GEFS_LBCS_FILES is the same except for the wgrib2 interpolation stuff.

BenjaminBlake-NOAA:
This issue was completed with PRs #278 , #284 , and #285.

Issue #195, opened 2024-01-17
GET_GEFS_LBCS_FILES task loads modules in ex-script


chan-hoo:
Currently, the `GET_GEFS_LBCS_FILES` task loads the necessary modules in its ex-script. The modules should be loaded in the job-card level. (job card > J-job > ex-script)

BenjaminBlake-NOAA:
Closing this issue because the GET_GEFS_LBCS_FILES task no longer exists.  This was resolved with PR #278

PR #278, opened 2024-03-20
Remove get_extrn_ics and get_extrn_lbcs tasks

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- In this PR, the get_extrn_ics and get_extrn_lbcs tasks have been removed from the rrfs-workflow.  All functionality from these tasks has been incorporated into the make_ics and make_lbcs tasks.  In addition, all processes for retrieving files from HPSS have been removed, so the only option is to run the workflow with IC/LBC files on disk, which is how the RRFS will run in operations.  I confirmed with Matt and Ming that the HPSS retrieval option can be removed.  Successful make_ics and make_lbcs jobs were completed with the non-DA engineering and DA engineering workflows.  
- In FV3LAM_wflow.xml, the dependencies from get_extrn_ics are added to make_ics, and the dependencies from get_extrn_lbcs, get_gefs_lbcs, and get_extrn_lbcs_long are added to make_lbcs.  In addition, the make_lbcs task is broken into two separate tasks within the larger make_lbcs metatask, one for cycledef=boundary and another for cycledef=boundary_long.  This is because the Rocoto dependencies are different for these cycle definition types.  The boundary_long cycle definition is for hourly cycles that need to generate boundaries for each cycle, and some cycles need long boundaries, like the HRRR.  boundary_long is not needed for RRFS, so we should consider removing this eventually.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [x] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro - Ran tests for a cycle that was a day or two behind the real-time parallel.
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Resolves issues #97 and #195


MatthewPyle-NOAA (approved):
I will approve and get this one merged, but the MAKE_ICS and MAKE_LBCS jobs will need a lot of future cleanup for production (will be pulling in specific files directly, without the kind of flexibility currently in the system).  Starting to think about when we should spin off a wcoss2_production branch of the workflow to make the needed changes that won't be good for RRFS developers.

PR #290, opened 2024-03-28
Change JET compile and run environment setups for Rocky 8

hu5970:
Those are environmental changes for RRFS running on JET with Rocky 8 OS update.
Tested with retro on JET.

To use crontab,  need to change "run_rocoto.sh" to:
```
#!/bin/bash
# module load rocoto
/apps/rocoto/1.3.6/bin/rocotorun -w FV3LAM_wflow.xml -d FV3LAM_wflow.db
```

Before more JET changes coming.


MatthewPyle-NOAA (approved):
Since only touches Jet, which approve it without a worry.  Thanks for helping with the Rocky 8 transition.

PR #296, opened 2024-03-30
1) Cycle smoke/dust at 06/18Z spinup cycle

hu5970:
1) Cycle smoke/dust at 06/18Z spinup cycle
     if 04/16Z smoke/dust cycle is skipped.
2) Bug fix boundary condition dependence (change or to and) 
3) Use rrfs_utl release branch.
4) Set UPP/prdgen for 15-min output.

Tested with RRFS_A v0.9.1 on Dogwood.

### DIFF
--- a/jobs/JRRFS_RUN_PREPSTART
+++ b/jobs/JRRFS_RUN_PREPSTART
@@ -93,6 +93,19 @@ export MAILTO="${MAILTO:-}"
 #
 #-----------------------------------------------------------------------
 #
+# create COMOUT directory
+#
+#-----------------------------------------------------------------------
+#
+if [ "${CYCLE_TYPE}" = "spinup" ]; then
+  export COMOUT="${COMOUT_BASEDIR}/$RUN.$PDY/${HH}_spinup"
+else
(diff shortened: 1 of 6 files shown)


MatthewPyle-NOAA (commented):
Looks okay to me at first glance, but will wait until tomorrow morning to approve/merge.

BenjaminBlake-NOAA (commented):
@hu5970  Is the change in FV3LAM_wflow.xml needed?  The dependencies are already inside one `<and>` tag so I didn't think a second `<and>` tag would be needed.  Otherwise these changes look good to me, thanks

hu5970:
> @hu5970 Is the change in FV3LAM_wflow.xml needed? The dependencies are already inside one `<and>` tag so I didn't think a second `<and>` tag would be needed. Otherwise these changes look good to me, thanks

Those <and> block is just for marking the whole boundary files as a group. We can remove it or keep it. It will not impact any thing.

BenjaminBlake-NOAA (approved):
@hu5970 Thanks, the changes look good to me.

MatthewPyle-NOAA (approved):
Will get this one merged...

PR #302, opened 2024-04-03
Update Hera compile and run environments for Rocky 8

hu5970:
All of the Hera is Rocky 8 now.

Tested with RRFS retro on Hera.


MatthewPyle-NOAA (approved):
Thanks for this contribution, @hu5970!

PR #303, opened 2024-04-03
Update model hash to b4a006e (April 2, 2024).

hu5970:
The bug fix committed by this PR is tested in RRFS_A and retro on Hera.


BenjaminBlake-NOAA (approved):
@hu5970  Thanks for testing, moving to the newer version of model code sounds good.

PR #307, opened 2024-04-04
Use hash number instead of branch name for the rrfs_utl release.

hu5970:
We need to use hash number to control which version of the code is used. When we make the tag, the same tag will always checkout the same code. We cannot control this with the branch name.


ShunLiu-NOAA (commented):
should you uncomment 
#branch = production/RRFS.v1

MatthewPyle-NOAA (approved):

MatthewPyle-NOAA:
> should you uncomment #branch = production/RRFS.v1

@ShunLiu-NOAA This change is to point at a specific hash rather than the branch name (which is being done for other components where we are using release branches).

PR #308, opened 2024-04-04
Use GSI release branch for RRFS.

hu5970:
Now we can use the GSI release branch for the RRFS application.
We use hash number for GSI release branch and also update the rrfs_utl hash to machine the GSI.
The branch names are also changed to reflect the release branch but we are not using the branch name because we want to tag certain hash for retro runs.


MatthewPyle-NOAA (approved):
Thanks for thanking things more clear, @hu5970

PR #264, opened 2024-03-05
Update UFS_UTILS for RRFSFW, remove unneeded modules from run.ver

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR contains two main updates (see below):  
- The UFS_UTILS code is updated for the RRFS fire weather nest to include the chgres_cube code changes needed to support rotated lat-lon grids with the WMO template.  This is what RRFS uses, and the fire weather grid is initialized using the RRFS rotated lat-lon data.  With this update, I've also changed RRFSFW to use the native level (natlev) files for ICs/LBCs instead of the pressure level (prslev) files.  The make_orog ex-script had to be updated to work with the newer version of UFS_UTILS, and some RRFS settings in the make_ics ex-script were modified to not use climatological values.
- The run.ver (and build.ver files) were updated to remove unneeded modules.  I looked at build.ver and run.ver of AQM as a starting point.  These changes are needed for EE2 compliance.  When running specific tasks, NCO only wants the modules loaded that are needed at runtime.  I tested all the jobs in the non-DA workflow with RRFSFW, and I also ran part of a DA engineering test to make sure the processing and analysis jobs would complete successfully.  I did not test the run_bufrsnd task, but the only change there is removing the PrgEnv-Intel module and that one was not needed by any other tasks.  Perhaps these module changes should also be tested in the v0.8.5 real-time runs.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [x] Parallel
- [x] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## CONTRIBUTORS (optional): 

@LarissaReames-NOAA and @GeorgeGayno-NOAA assisted with the UFS_UTILS code changes.

PR #321, opened 2024-04-12
Add PrgEnv-intel and bacio modules back to run.ver

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- @aerorahul noted that PrgEnv-intel needs to be loaded at runtime in order to use the intel module, and bacio is a prerequisite for w3emc.  In this PR, these two modules are added back to run.ver and the applicable modulefiles - they were previously removed in PR #264 .
- Removed nco and wgrib2 from build.ver because they are not needed at compile time.  Also switched the order of two listed modules in build.ver because they were not in alphabetical order.
- It is unclear to me why modulefiles/tasks/wcoss2/save_restart.local.lua is using some different versions of modules than the other tasks.  I would like to know why this is the case and if it is possible to use the same versions of modules as other tasks.

## TESTS CONDUCTED: 

No tests were conducted.  These modules were previously being loaded in the workflow so these changes should not affect anything.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:


MatthewPyle-NOAA (approved):
Will approve this one somewhat on faith...

PR #324, opened 2024-04-16
1) add function in BUFRSND to handle 15-min output

hu5970:
1) add function in BUFRSND to handle 15-min output
2) add TPP to workflow for BUFRSND.
3) update model hash.

Those changes are tested with RRFS_A v0.9.1 on Dogwood.
There are still some work needed for BUFRSND.


MatthewPyle-NOAA (approved):

MatthewPyle-NOAA:
> @hu5970 These changes look good to me, thanks.
> 
> One thought I have (unrelated to this PR) is if we want to name the first subhourly file -00-36.nc. I know it is after the first timestep of 36 seconds, so not truly a f00 file, but I don't really like the idea of labeling it as a 36 second file. If others disagree I'm happy to be overruled. :)

I think changing that file name coming out of the model would require model code changes.  I agree it is a bit awkward/ugly, but think we need to live with it.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA  Ahh okay, yeah if the logic for doing that is hardwired into the model and would be a hassle to modify, then I agree and let's leave it as is.

hu5970:
@BenjaminBlake-NOAA @MatthewPyle-NOAA  I agree the first (00-h) output file name is awkward. It is hardwired inside the model and reflect the output fields are true 36 second model forecast results.

PR #339, opened 2024-04-19
adding RTMA-specific conditions for BUFRSND workflow component.

EdwardColon-NOAA:
The BUFRSND workflow component will not be included in FV3LAM_wflow.xml without a conditional check in /parm/FV3LAM_wflow.xml. Furthermore, the time increment setting is adjusted for RTMA in scripts/exrrfs_bufrsnd.sh.


BenjaminBlake-NOAA (approved):
Looks like a safe change to me, approving.

MatthewPyle-NOAA (approved):
Looks good to me as well...

PR #342, opened 2024-04-22
Update fix/lam and fix/gsi, and configure files for CONUS 13km grid.

hu5970:
Update fix/lam and fix/gsi, and configure files for CONUS 13km grid.

Tested on Hera.


MatthewPyle-NOAA (approved):
Looks safe to me...approving.

PR #347, opened 2024-04-24
cycle the smoke at 18Z during product machine switch

hu5970:
1) cycle the smoke at 18Z based on the tracer file during product machine switch.
2) point to gsi fix for mesonet uselist and aircraft obs rejection list. 
3) update configure file for RRFS_A

Those changes are used in RRFS_A real-time and retros.


BenjaminBlake-NOAA (approved):
@hu5970  The changes look good to me.  If any or all of these changes are needed in operations then we'll want to add them to the production/RRFS.v1 branch too.  Thanks

MatthewPyle-NOAA (approved):
Wasn't sure about the forecast length for the retro configuration, but will approve.

PR #377, opened 2024-05-22
Update modules

hu5970:
Update to use the following modules for compiling on Hera and Jet:
   fms.2024.01   : for new model code
   g2tmpl/1.12.0 : for UPP

Tested on Hera and Jet with retros.


ShunLiu-NOAA (approved):
Looks good to me.

PR #392, opened 2024-06-12
Update test period in real-time configure files.

hu5970:
Tested on WCOSS2 with retros


MatthewPyle-NOAA (approved):
Config changes look safe, but was curious about the removal of g2tmpl loads from the two tasks.  Guessing it isn't needed for those, but wanted to make sure that removal was intentional

hu5970:
@MatthewPyle-NOAA Yes. Those two tasks do not need g2tmpl and they were using wrong version.

PR #438, opened 2024-08-16
Fixed a bug in FV3LAM_wflow.xml and high cost IO in make_lbcs

hu5970:
Fixed a bug in FV3LAM_wflow.xml.
Use CAT to combine two grib2 files in make_lbcs
Reduce the parallel runs of make_lbcs as much as possible. Parallel make_lbcs for GEFS waste lots of IO for now.

Those fixes are used on WCOSS2 for retros.


ShunLiu-NOAA:
This is detected as an issue that caused the I/O spike on WCOSS2. There are impact for both realtime and retros.

MatthewPyle-NOAA:
Thanks @ShunLiu-NOAA - sounds like a good fix.

Issue #394, opened 2024-06-13
Meet NCO Standard - Remove hardwired environment variable


lgannoaa:
Please review j-job, ex-script, configuration (var_defns.sh) and remove hardwired environment variable. Using the following solution:
- Taking the following parameter value from workflow level:
   PACKAGEHOME, HOMErrfs, PDY, cyc, FHR, MEMBER_NAME, BCGRP, MEM_TYPE, GSI_TYPE, BCGUP
- Modify application logic to dynamically calculate or assign the following environment variables:
   CYCLE_DIR, GLOBAL_VAR_DEFNS_FP, SLASH_ENSMEM_SUBDIR, ENSMEM_INDX, fhr, TMMARK, NWGES_BASEDIR, LOGDIR, nens.
- All environment variable assignment (include var_defns.sh) should take default value if set already from workflow level. 
   For example, export USHrrfs=${USHrrfs:-${HOMErrfs}/ush}

MatthewPyle-NOAA:
@BenjaminBlake-NOAA Could this issue be closed?

BenjaminBlake-NOAA:
@MatthewPyle-NOAA I think this one should remain open for now.  A future workflow item will be to remove var_defns.sh which will remove more hardwired variables.  @lgannoaa is more familiar with how this was done for AQM.

Issue #238, opened 2024-02-21
Use AQMv7 for guidance


MatthewPyle-NOAA:
AQMv7 (also based on the SRW app) had to make numerous changes to satisfy NCO.  Get a copy of their final version to use as a template on what changes we may need beyond what we've currently considered.

JacobCarley-NOAA:
For reference, a copy NCO's compliance issue spreadsheet can be found here: https://docs.google.com/spreadsheets/d/159-IhX7smtWjggP2l2bvnNHR10QRoW0g2AgwaqbMry8/edit?usp=sharing

MatthewPyle-NOAA:
Lin is very familiar with AQMv7, and certain things that were okay for that implementation are no longer acceptable to NCO.

PR #445, opened 2024-08-26
[production/RRFS.v1] Add HOMErrfs (and related envars) and FIX directory definitions to J-jobs

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- In this PR, the following standard environment variables are added to the J-jobs in the workflow: `$HOMErrfs, $EXECrrfs, $FIXrrfs, $PARMrrfs, $USHrrfs`.  `$HOMErrfs` is defined at the job card level for now.  Eventually we may want to define it at the J-job level using the `$PACKAGEROOT` environment variable in the implementation standards document, which would be defined at the job card level, so I added commented out versions of `$HOMErrfs` to the J-jobs in case we want to go this route.
- With these new variables being defined, the following environment variables have been removed from the workflow: `$JOBSdir, $EXECdir, $PARMdir, $USHdir, $SCRIPTSdir, $SORCdir.`
- `FIX_GSI, FIXLAM, FIX_BUFRSND, FIX_SMOKE_DUST, FIX_CRTM, FIX_UPP`, and `FIX_UPP_CRTM` have been removed from the workflow generation part and have instead been added to the J-jobs, which is where they should be defined using `$FIXrrfs`.  A future PR could do more of this for the other fix directories.  I did not want to include that in this PR because this one already has enough changes in it.
- With `FIX_CRTM` and `FIX_UPP_CRTM` now defined at the J-job level using the `$CRTM_FIX` path from the crtm module, the crtm module needs to be loaded for the ANALYSIS_GSI and POST tasks.
- The `$COMnam` path is now defined in JRRFS_MAKE_GRID for the RRFS fire weather nest, which currently needs to read in the `/lfs/h1/ops/prod/com/nam/v4.2/input/nam_firewx_loc` file.  `nam.ver` is also added to `versions/run.ver`.  Previously this path was hardwired in the `exrrfs_make_grid.sh` script and that is not desirable.

## TESTS CONDUCTED: 

Successful tests were completed with the DA engineering test (spinup cycle) and the fire weather nest.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [x] Parallel
- [x] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Partly related to issues #238 and #394


jobs/JRRFS_ANALYSIS_ENKF
MatthewPyle-NOAA, line 1: Should the commented out HOMErrfs defining line be removed?  Applies to all J-jobs, I think.  Since that is a job card (ecflow) level defined item, it probably doesn't need to be here in any fashion.

scripts/exrrfs_analysis_enkf.sh
MatthewPyle-NOAA, line 1: [note to Matt: ] Double check if FIX_GSI was an item that was included in this or not.

scripts/exrrfs_make_grid.sh
MatthewPyle-NOAA, line 1: Should this COMnam be COMINnam?  (bringing in information from the NAM)

MatthewPyle-NOAA (commented):
I need to think more about some of the FIX redefining (think saw FIX_GSI being used even though isn't being defined any more).  But overall makes things look much more production like - nicely done!

scripts/exrrfs_make_grid.sh
BenjaminBlake-NOAA: Yes, good catch.  When RRFS is implemented and the NAM fire weather nest is turned off, the RRFS should have its own firewx_loc file with lat/lon coordinates, and this NAM file won't be used anymore.

jobs/JRRFS_ANALYSIS_ENKF
BenjaminBlake-NOAA: Agreed, I'll get that removed.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA Thanks for the review!  Let me know if you want anything modified related to the FIX directories, and I'm willing to do so.

MatthewPyle-NOAA (approved):
Looks good to me now - thanks for the updates.  And now understand what you did with the FIX definitions.

PR #455, opened 2024-09-05
Add tasks - ioda_bufr, ic, lbc

guoqing-noaa:
Add tasks ioda_bufr, ic and lbc

Make changes as suggested in previous PR reviews


scripts/exrrfs_lbc.sh
MatthewPyle-NOAA, line 1: The norm is for the 

```
export err=$?
err_chk
```
kinds of lines to immediately follow the executable execution.  I understand that checking on whether the expected output file probably achieves the same aim, I'd recommend pulling the err from the executable execution.

MatthewPyle-NOAA (commented):
Mostly looks good to me, but a few concerns/problems that I commented on.

BenjaminBlake-NOAA (commented):
@guoqing-noaa Thanks for putting this together.  It mostly looks good to me.  I agree with @MatthewPyle-NOAA 's comments and I have one additional suggestion.  In the J-jobs, I would use ${HOMErrfs}/scripts instead of ${SCRIPTSrrfs} to point to the ex-scripts.  In operations and in the implementation standards they don't use a ${SCRIPTS} variable for whatever reason.

scripts/exrrfs_ic.sh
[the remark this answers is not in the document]
guoqing-noaa: Thanks for the good suggestion! I will add a variable `MPI_RUN_CMD` so that we can define different mpi run commands for different scheduler systems.

scripts/exrrfs_ioda_bufr.sh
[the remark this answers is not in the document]
guoqing-noaa: Yes, this is a tricky situation. Because we have to decode the prepbufr file first in order to know what types of observation are available in a prepbufr. We can either add an extra step to query prepbufr. Or We can modify the bufr2ioda.x source code to just output FYI messages without exiting with an error when given types of observation are not available. 

At the same time, EMC is planning to replace bufr2ioda.x with bufr-query (see https://github.com/NOAA-EMC/RDASApp/issues/149). So I would suggest we live with this part until things are sorted out. I can create an issue at rrfs-workflow for awareness.

scripts/exrrfs_lbc.sh
guoqing-noaa: @MatthewPyle-NOAA I agree with your comments!  
Normally we should do that. But as stated in my reply in the above, we don't have a final solution for bufr2ioda.x yet. 
I will modify these lines as follows: 
check how many ioda*nc files we have, if zero, then err_exit.

jobs/JRRFS_IC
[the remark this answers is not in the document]
guoqing-noaa: To confirm, 
we want to change
`$SCRIPTSrrfs/exrrfs_${task_id}.sh`  
to 
`$SCRIPTSrrfs/exrrfs_ic.sh`  

Do I understand it right? Thanks!

scripts/exrrfs_lbc.sh
guoqing-noaa: Now I think differently: 
if zero ioda*nc files, we may not want to block the whole workflow because of this.
if we don't have observations, the cycling should still move forward.

guoqing-noaa:
Thanks a lot for the review, @MatthewPyle-NOAA and @BenjaminBlake-NOAA 

I will go ahead to make changes as suggested. 
I will change `${SCRIPTSrrfs}` to `${HOMErrfs}/scripts`

guoqing-noaa:
@MatthewPyle-NOAA @BenjaminBlake-NOAA 
I made changes based on your review. Tests ran correctly on Hera/Jet. Thanks!

BenjaminBlake-NOAA (commented):
@guoqing-noaa Thanks for your work on addressing our suggestions.  I am wondering if the ${task_id} variable should be removed entirely.  I don't think it is used in operations.  Those "OUTPUT.${task_id}" files are typically just "OUTPUT.$$", where $$ is a job identifier.  @MatthewPyle-NOAA what do you think?

MatthewPyle-NOAA:
> @guoqing-noaa Thanks for your work on addressing our suggestions. I am wondering if the ${task_id} variable should be removed entirely. I don't think it is used in operations. Those "OUTPUT.${task_id}" files are typically just "OUTPUT.$$", where $$ is a job identifier. @MatthewPyle-NOAA what do you think?

@BenjaminBlake-NOAA @guoqing-noaa I personally like the define it once and use it approach (less error prone), but it isn't the production norm.  So I agree the ${task_id} concept should go.  Having things explicitly defined makes it more straightforward to grep or search for within a script.

guoqing-noaa:
@BenjaminBlake-NOAA @MatthewPyle-NOAA Thanks a lot for the further review!
To confirm, we want to remove all ${task_id} in J-jobs and ex-scripts? Or only the `export pgmout="${DATA}/OUTPUT.${task_id}` part?    

I would hope we can have a hybrid solution: use explicit task_id where NCO has a requirement or preference while using the implicit variables elsewhere (this may make the comparison of different j-jobs/ex-scripts easier and more meaningful).  

But it is fine to use explicit task_id in all places if it is preferred.

guoqing-noaa:
@BenjaminBlake-NOAA 
Just curious, `$$`  returns the Process ID (PID) of the current shell or script. Theoretically, there is a possibility that we may get the same number for different runs in some rare situations. Is there a concern about this?

BenjaminBlake-NOAA:
@guoqing-noaa I'm inclined to get rid of ${task_id} entirely but I could also see where a hybrid approach would be useful.  I'll defer to @MatthewPyle-NOAA on the final decision.  :)

I hadn't considered that possibility for $$.  NCO may not be concerned about it though because the production J-jobs use it, so I think it is acceptable.

MatthewPyle-NOAA:
I think that OUTPUT.$$ only exists within a particular run directory.  They don't get collected somewhere permanent as OUTPUT.$$.  

In the implementation standard, it seems to get dumped into an overall job log file (OUTPUT.$$ generally is produced by an executable).

```
if [ -e "$pgmout" ]; then
cat $pgmout
fi
```

guoqing-noaa:
@BenjaminBlake-NOAA @MatthewPyle-NOAA 
Thanks a lot for your review. I have made changes as suggested:
1. remove all task_id variables but use their explicit literal values in j-jobs and ex-scripts.
2. change to OUTPUT.$$

Sorry for the weird commit message in one of the latest commits:
`change OUTPUT. to OUTPUT.202287 per the review feedback`   
It should be:
`change OUTPUT. to OUTPUT.$$ per the review feedback` 
I forgot that the bash shell will expand `$$` in the command line, :)

MatthewPyle-NOAA:
Sounds like you're ready for another look at your changes, @guoqing-noaa ?  Let me know if you're done updating for now and I can look at it again tomorrow.  Thanks!

guoqing-noaa:
> Sounds like you're ready for another look at your changes, @guoqing-noaa ? Let me know if you're done updating for now and I can look at it again tomorrow. Thanks!


Yes, it is ready for a new round of review. Thanks, @MatthewPyle-NOAA !

BenjaminBlake-NOAA (approved):
@guoqing-noaa Thank you for addressing my comments!  I approve of these changes.

guoqing-noaa:
Thank you, @BenjaminBlake-NOAA!

jobs/JRRFS_IODA_BUFR
MatthewPyle-NOAA, line 1: One small item - the "official" standard production variable for deciding on whether or not to delete $DATA at the end of a job is "KEEPDATA".   Would this kind of approach work?  Pull in a task specific KEEPDATA variable to define KEEPDATA, and then use that in decision making.

```
KEEPDATA=${KEEPDATA_ioda_bufr:-${KEEPDATA}}
[[ "${KEEPDATA}" == "NO" ]] && rm -rf ${DATA}
```
Also move it up to where things like DATA are being defined.   Also make change in other scripts utilizing KEEPDATA.

MatthewPyle-NOAA (commented):
One relatively small request, but definitely looking a lot better.  Thanks for the updates you've already made.

jobs/JRRFS_IODA_BUFR
guoqing-noaa: Thanks, @MatthewPyle-NOAA!

My preliminary thoughts on this: 
Our experience finds that sometimes we want to **only** keep some specific task run directories, not all run directories. This setting will facilitate that and it does not break NCO standards since if you set KEEPDATA=YES, it will function as expected in NCO.  This as well gives us flexibilities to keep different task run directories in our research if we define a KEEPDATA_task_id variable.

I am fine with either of the options:
1. Keep this strategy, but move the assignment up as suggested.
2. Remove task-specific KEEPDATA_task_id variables

Let me you what do you prefer and I can go ahead to make changes accordingly. Thanks!

guoqing-noaa: Ah, I understand what this means now. Thanks, @BenjaminBlake-NOAA and @MatthewPyle-NOAA .

MatthewPyle-NOAA (approved):
Thanks again for addressing concerns, @guoqing-noaa.  Approving now.

guoqing-noaa:
Thanks, @MatthewPyle-NOAA and @BenjaminBlake-NOAA

PR #485, opened 2024-09-15
create a stage_data/ directory to store relevant scripts

guoqing-noaa:


ShunLiu-NOAA:
Here is an example of util directory used in GFS. 
image
Let's discuss how to handle these scripts. I think we can borrow some idea you used for .agent to handle staged data.

guoqing-noaa:
@ShunLiu-NOAA 
Thanks for sharing the GFS information.
To clarify, we don't track staged data in rrfs-workflow as those data can be easily regenerated using the stage_data scripts.
This PR is to move those existing scripts into a subdirectory to make the util/ directory more organized.

BenjaminBlake-NOAA:
Hi @guoqing-noaa , I agree it is a good idea to move those staging data scripts into a subdirectory to make the main util/ directory more organized.  I think it would be good to use one of the operational directory names if possible, like the ones @ShunLiu-NOAA  listed for GFS.  It looks like all the files in util are utility scripts, so my recommendation would be to place everything there in util/ush (including what's in the sideload subdirectory).  What do you and others think?  I'm open to keeping things in separate directories too if that makes more sense!

guoqing-noaa:
@BenjaminBlake-NOAA Thanks for the discussion.
I think Shun's screenshot is from GFSv16.x.  
**GFSv17** has moved one step forward to have a separate [UFS_UTILS ](https://github.com/ufs-community/UFS_UTILS)repo for all utilities required at the workflow runtime, its repo is at: https://github.com/ufs-community/UFS_UTILS
   
So the global workflow itself no longer has a util/ directory any more. See below,  see attached figure below:
https://github.com/NOAA-EMC/global-workflow
So the general idea is to put all source codes (including utility source codes) under sorc/.

I think we agreed before that we only put runtime scripts in `ush/` and non-runtime (or non-NCO) scripts under `util/`. And in the implementation stage, we can simply remove the whole `util/` directory without any impact on operation.
For rrfs utilities, we already have a dedicated repo: https://github.com/NOAA-GSL/rrfs_utl, so we will NOT put any source code into `util/`

If we think the current `util/` does not serve its literal meaning,  we can change it to `ush2/`, `xsh`, or any other better names.

image

BenjaminBlake-NOAA (approved):
@guoqing-noaa Thanks for the additional explanation.  Since these scripts aren't needed for implementation your solution is good for me!

guoqing-noaa:
Thanks, @BenjaminBlake-NOAA and @ShunLiu-NOAA!

I had a conversation with @ShunLiu-NOAA and we agreed that we may NOT need the `util` directory at the moment as all rrfs utilities (source codes inside the rrfs_utl repo) will go under `sorc/`; and all scripts under current `util` will not be needed by the operation and should go into `workflow` and `workflow/ush`. I will close this PR and create a new PR to make the proposed changes.

PR #511, opened 2024-10-03
Add the fcst task

guoqing-noaa:
We will add the `fcst` task with 3 PRs to facilitate the review process and keep a change record.

1. commit what we have right now (a `fcst` task as well as "pass-through" `da` task for the stop and start fcst capability)
2. update to use the latest MPAS v8.2.2 code and the mpasout files
3. complete the left-over part to enable the `forecast_only` capability in the workflow 

This PR is the first one.
The workflow has been tested on Hera and Jet

### DIFF
--- a/doc/build_and_run.md
+++ b/doc/build_and_run.md
@@ -48,7 +48,7 @@ Refer to [this guide](https://github.com/NOAA-EMC/rrfs-workflow/wiki/deploy-a-re
     
 This Python script creates an experiment directory (i.e. `EXPDIR`), writes out a runtime version of `exp.setup` under EXPDIR, and  then copies runtime config files from `HOMErrfs` to `EXPDIR`.
        
-Users usually need to set up `ACCOUT`, `QUEUE`, `PARTITION`, or `RESERVATION` by modifying `resources/config.${machine}` or you may export those variables in the current environment before running setp_exp.py, or export those variables in `exp.setup`.  
+Users usually need to set up `ACCOUT`, `QUEUE`, `PARTITION`, or `RESERVATION` by modifying `config_resources/config.${machine}` or you may export those variables in the current environment before running setp_exp.py, or export those variables in `exp.setup`.  
     
 The workflow uses a cascade config structure to separate concerns so that a task/job/application/function_group only defines relevant environmental variables required in runtime. Refer to [this guide](https://github.com/NOAA-EMC/rrfs-workflow/wiki/The-cascade-config-structure) for more information.
 
--- a/doc/build_and_run.md
+++ b/doc/build_and_run.md
@@ -48,7 +48,7 @@ Refer to [this guide](https://github.com/NOAA-EMC/rrfs-workflow/wiki/deploy-a-re
     
 This Python script creates an experiment directory (i.e. `EXPDIR`), writes out a runtime version of `exp.setup` under EXPDIR, and  then copies runtime config files from `HOMErrfs` to `EXPDIR`.
        
-Users usually need to set up `ACCOUT`, `QUEUE`, `PARTITION`, or `RESERVATION` by modifying `config_resources/config.${machine}` or you may export those variables in the current environment before running setp_exp.py, or export those variables in `exp.setup`.  
+Users usually need to set up `ACCOUNT`, `QUEUE`, `PARTITION`, or `RESERVATION` by modifying `config_resources/config.${machine}` or you may export those variables in the current environment before running setp_exp.py, or export those variables in `exp.setup`.  
     
 The workflow uses a cascade config structure to separate concerns so that a task/job/application/function_group only defines relevant environmental variables required in runtime. Refer to [this guide](https://github.com/NOAA-EMC/rrfs-workflow/wiki/The-cascade-config-structure) for more information.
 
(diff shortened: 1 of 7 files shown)


doc/build_and_run.md
MatthewPyle-NOAA, line 1: Small item, but guessing `ACCOUT` should be `ACCOUNT`?

jobs/JRRFS_FCST
MatthewPyle-NOAA, line 1: Just a comment for now, but eventually we'll need an ability to restart a model run from a midpoint (like in production if there is a hardware problem).  Deleting the run directory if it exists is bad in this case (we'll need to add checks on the existence of restart files - assuming MPAS has something like that - to see if we're restarting a previously started run.  Nothing you need to do now - just planting the idea for the future.

scripts/exrrfs_da.sh
MatthewPyle-NOAA, line 1: Should the block from lines 33 to 38 be modified to only execute when needed?  I feel like doing nothing within an if block can be problematic on some systems.   Would this logic do what you need?

if [[ "${begin}" != "YES" ]]; then
  cpfs ${DATAROOT}/${NET}/${rrfs_ver}/${RUN}.${CDATEm1:0:8}/${CDATEm1:8:2}/fcst/restart.${timestr}.nc .
fi

MatthewPyle-NOAA (commented):
A few relatively minor comments for you - mostly looks very good to me.

doc/build_and_run.md
guoqing-noaa: Yes, thanks a lot for catching this typo!

jobs/JRRFS_FCST
guoqing-noaa: Thanks for the input!  The good thing is that MPAS is very similar to WRF, we can "restart" easily from the `mpasout` (similar as `wrfout`) files. Yes, we will keep this in mind and sort out things later.

scripts/exrrfs_da.sh
guoqing-noaa: Good catch!  A few months ago when I developed this workflow, we were NOT able to do data assimilation on cold start. But now we figured out a way to do DA on cold start. We will modify this when we update the DA task. For now, we can leave it as-is and update later. Do this work for you? Thanks!

guoqing-noaa: Just removed the do-nothing line as suggested. Thanks!

hu5970:
As we discussed, we want to add the model parts, including ungrid, ics/lbcs, MPAS, MPASSIT, and UPP first and make sure those components can be run on Hera/JET and meet the operation standard as much as we can. Let's focus on those parts first. I will say remove all the DA parts or leave it as is. We will come back to DA later.

I will comment the model parts soon after I get time to test them.

guoqing-noaa:
The DA task here is to facilitate the stop and restart capability. No any real DA task at the moment.

BenjaminBlake-NOAA (commented):
@guoqing-noaa The changes look good to me - thanks!  One suggestion I have - could the JRRFS_DA and exrrfs_da.sh files be renamed to something more specific?  When I see DA by itself, it makes me think this task is actually doing DA when in this case it is not.

guoqing-noaa:
@BenjaminBlake-NOAA Thanks!
Sorry that this is confusing. 

To clarify, your understanding is right. These tasks are for real "DA". It is that at this moment, we don't modify the background in the DA step. We want to test whether the workflow work as expected if we do 1 hr cycling by stopping the model and running it again.  I think we don't want to add a temporary task for the "stop the model and run it again" capability.

BenjaminBlake-NOAA:
@guoqing-noaa That makes sense, thanks!  My suggestion would be to rename it to JRRFS_RESTART/exrrfs_restart.sh, or JRRFS_DA_RESTART/exrrfs_da_restart.sh.  But I am flexible.  @MatthewPyle-NOAA is on leave today but we can wait and see what he thinks.

guoqing-noaa:
@BenjaminBlake-NOAA 
In MPAS, we don't need "restart". It is similar like RAP/HRRR which can start forecasts from `wrfout` files,  MPAS can start forecasts directly on `mpasout` files.

BenjaminBlake-NOAA (approved):
@guoqing-noaa Okay sounds good, I'll go ahead and approve the PR now and we can always revisit renaming this later on if necessary.

guoqing-noaa:
Thanks, @BenjaminBlake-NOAA!  Sounds a good plan.

To add a little more information, I commit this DA task so that others can also contribute here to update the DA task making it assimilate more and more observations. Otherwise, people will need to use my fork to make contributions.

edwardhartnett:
Is there an issue associated with this PR?

guoqing-noaa:
> Is there an issue associated with this PR?

I think this PR is good to go. But @MatthewPyle-NOAA was on leave yesterday. He should be able to look at this on Monday.

scripts/exrrfs_da.sh
MatthewPyle-NOAA: It seems like the block from lines 33-38 is the same.  Did you make a change elsewhere in the script?

MatthewPyle-NOAA (commented):
Still a question about scripts/exrrfs_da.sh, but thanks for fixing the other issues.

guoqing-noaa:
> Still a question about scripts/exrrfs_da.sh, but thanks for fixing the other issues.

Ah, I forgot there are two parts of "do-nothing" lines. Just fixed it. Thanks, @MatthewPyle-NOAA

Issue #467, opened 2024-09-08
all python code should have docstrings


edwardhartnett:
To document Python modules, classes, and subprograms should be documented with pythin docstrings. 

See https://peps.python.org/pep-0257/ for details about docstrings.

PR #517, opened 2024-10-08
documentation for blending_fv3.py

edwardhartnett:
## DESCRIPTION OF CHANGES: 
documentation for blending_fv3.py

## TESTS CONDUCTED: 
N/A

### Machines/Platforms:
N/A

### Test cases: 
N/A

## ISSUE: 
Part of #467

### DIFF
--- a/ush/blending_fv3.py
+++ b/ush/blending_fv3.py
@@ -1,9 +1,22 @@
+"""
+This file handles blending of FV3 tiles???
+
+"""
 import numpy as np
 from netCDF4 import Dataset
 import raymond
 import sys
 
 def check_file_nans(test_nc, vars_fg, vars_bg, name):
+    """
+    Check file NANS?
+
+    test_nc: Test netCDF file.
+    vars_fg: Vars?
+    vars_bg: Vars?
+    name: Name of file?
+
+    """
     nans = False
     for (var_fg, var_bg) in zip(vars_fg, vars_bg):
         i = vars_fg.index(var_fg)
--- a/ush/blending_fv3.py
+++ b/ush/blending_fv3.py
@@ -33,6 +33,11 @@ def check_file_nans(test_nc, vars_fg, vars_bg, name):
         return nans
 
 def err_check(err):
+    """
+    Check for error.
+
+    err: Error code.
+    """
     if err > 0:
         print(f"An error ocurred in {sys.argv[0]}. Blending failed!!!")
         print(f"err={err}")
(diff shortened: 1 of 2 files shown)


edwardhartnett:
@delippi can you please help me fill in the missing documentation in this code?

Wherever ??? appears in the source file, I need documentation. Add it in the review and I will put it in the file.

ush/blending_fv3.py
```diff
@@ -1,9 +1,22 @@
+"""
+This file handles blending of FV3 tiles???
```
edwardhartnett, line 2: @delippi what does this file do?

ush/blending_fv3.py
```diff
@@ -1,9 +1,22 @@
...
 import sys
 
 def check_file_nans(test_nc, vars_fg, vars_bg, name):
+    """
+    Check file NANS?
+
+    test_nc: Test netCDF file.
+    vars_fg: Vars?
+    vars_bg: Vars?
+    name: Name of file?
```
edwardhartnett, line 17: @delippi what are the meanings of these parameters?

ush/blending_fv3.py
delippi: """
This script performs blending between regional and global weather forecast model restarts using the Fortran module raymond. The Raymond filter is a sixth-order tangent low-pass implicit filter and can be controlled via the cutoff length scale (Lx).
"""

ush/blending_fv3.py
delippi: """
Check for NaN values in specified variables of a netCDF file.

This function iterates over a list of variables and checks for NaN values in the provided 
netCDF file. It prints the count of NaNs found for each variable and indicates whether 
any NaNs were detected.

Parameters:
    test_nc: Dataset
        The test netCDF file containing the variables to be checked for NaN values.
    vars_fg: list of str
        A list of variable names from the regional model (foreground) to check.
    vars_bg: list of str
        A corresponding list of variable names from the global model (background) to check.
    name: str
        A string representing the context (e.g., 'glb' for global or 'reg' for regional) 
        to identify the source of the variables being checked.

Returns:
    bool
        Returns True if any NaN values are found in the specified variables; 
        otherwise, returns False.
"""

Again, if there are any NaNs found, I wanted to catch that here instead of later when the model is running. I don't think there is any reason to expect NaNs.

MatthewPyle-NOAA (approved):
Looks mergeable....thanks for another documentation contribution.

MatthewPyle-NOAA (approved):
Reapproving...

PR #531, opened 2024-10-16
allow fcst-only runs every  a few hours

guoqing-noaa:
This PR enable only running MPAS model forecasts (no DA) every a few hours (for example every 6 hours).
It uses GFS grib2 files as an example.

Top level configurations in exp.setup to run fcst_only:
```
export FCST_ONLY=${FCST_ONLY:-true}
export FCST_ONLY_FREQ=6 # every FREQ hours
export IC_OFFSET=6
``` 
One can overwrite other default rrfs-workflow settings in this file.    
Default settings can be found under `parm/config`.  
An example `exp.fcst_only` is provided under `exp/`.  And one can `cp exp/exp.fcst_only exp.setup` to get a quick start.
Modify other settings in exp.setup accordingly (such as directories).

Other related code changes:
1. `workflow/rocoto_funcs/smart_cycledefs.py`
This is to setup fcst_only cycledefs for the rocoto workflow.

2. `workflow/rocoto_funcs/tasks1.py`
This is to apply slightly different dependencies  for  the `fcst` task if fcst_only

3. `workflow/setup_rocoto.py`
This is to preempt the `PROD_BGN_AT_HRS` array to tell exrrfs_fcst.sh that each cycle will start from a cold start instead of a warm start.


MatthewPyle-NOAA (approved):
Approving, but didn't scrutinize very carefully since this PR is only touching the workflow/rocoto setup.

PR #534, opened 2024-10-18
Fine-tune the experiment setup process

guoqing-noaa:
Per the discussions with @hu5970 @chunhuazhou @Junjun-NOAA @spanNOAA, we made the following changes:

- fix the misue of the reserved `NET` variable
- let each experiment have its own `COM` and `DATA` directories
- add `get_required_env()` which stop the experiment setup process if a given environmental variable is NOT found
- Split functions in `tasks*.py` into individual files, each corresponding to one task.


MatthewPyle-NOAA (approved):
Looks good to me based on a quick look

Issue #525, opened 2024-10-11
clean up jobid definitions in J-jobs


MatthewPyle-NOAA:
jobid/taskid defining has yet to be done consistently throughout the system.  This issue is to remind us to consistently apply a solution for all J-jobs.

hu5970:
@BenjaminBlake-NOAA Thanks for working on this. I will run test after your make your PR.

BenjaminBlake-NOAA:
@hu5970 Thank you, that sounds good to me.  I should have a PR ready later today or tomorrow.

BenjaminBlake-NOAA:
@hu5970 I just created a draft PR with these changes here: #535 
Are you still able to help with testing?  I ran a quick DA engineering test and it completed successfully.

MatthewPyle-NOAA:
@BenjaminBlake-NOAA can this issue be closed now?

PR #524, opened 2024-10-11
Add taskid and run deterministic and EnKF DA cycles.

hu5970:
Add taskid (jobid) to each task and run deterministic and EnKF DA cycles.

Tests:
- WCOSS2
  - [x ] Cactus/Dogwood
- [X] RRFS_A:

Issue #498, opened 2024-09-24
directory structures:  `com`, `umbrella`, `$DATA`


guoqing-noaa:
Based on today's discussion, we will have three different types of major directories depending on the lifetime of output files in the production:

1. `$DATA`:  this is the run directory of a task. Files under this directory will be removed upon the completion of a task

2. `umbrella`:  this directory holds output files to be shared inside one cycle and will be purged upon the completion of a cycle

3. `com`: this directory holds output files to be shared across cycles and will be purged after ~6 days in production.

NOTE: both $DATA and umbrella are under `$DATAROOT`

In the workflow, we can set the default behavior to use the operational settings while in the develop platforms, users can have the flexibility to set the actual locations of the above three types of directories.

@MatthewPyle-NOAA @BenjaminBlake-NOAA @ShunLiu-NOAA @hu5970 @lgannoaa Let me know if there is anything I missed and feel free to edit/comment. Thanks!

lgannoaa:
The document for today's demo can be found [here](https://docs.google.com/document/d/1He_WMbebndQ9IyUC_2D7P80X9oFP_gnqaagYE7CN6jU/edit?usp=sharing).

Issue #36, opened 2023-08-17
Confirm that compath.py is being used properly, and check COM type defs


MatthewPyle-NOAA:
Look to see if the compath.py utility is being used within scripts to define certain operational directory paths.  Add or modify as needed.

Related, make sure that items like COMIN, COMOUT, COMINgefs are being defined and used properly.
jobid - Should be available, defined in the job card (ecf in operations) that executes the J-job ($job.$PBS_JOBID, $job is the job name)

Is compath.py now available on other platforms?  Guessing not, or that it won't function similarly beyond WCOSS2.  So will need to be tested carefully beyond WCOSS2.

BenjaminBlake-NOAA:
@MatthewPyle-NOAA @lgannoaa  I think this issue is mostly complete, however $jobid may need to be updated in the workflow.  

Currently I define $jobid in the J-jobs as ${RUN}_${task}_${envir}_${cyc}, and it is the name of the run directory.  Do we need to add ${PBS_JOBID} to the end of that, and also define it at the job card level instead of within the J-job?  It looks like in AQM, $jobid is defined at the job card level:
export DATA=${DATA:-${DATAROOT}/${jobid}}

lgannoaa:
Hi @BenjaminBlake-NOAA, thank you for your help. This is an important step to bring RRFS package a step closer to follow the NCO implementation standard. Using AQM code as guidance is an excellent approach.
We set the pid in J-Job for jobs with unique DATA as:
export pid=${pid:-$$}
export outid=${outid:-"LL$job"}
export DATA=${DATA:-${DATAROOT}/${jobid}}
mkdir -p ${DATA}
cd ${DATA}

Please do not define ${PBS_JOBID} in j-job or ex-script.

BenjaminBlake-NOAA:
Hi @lgannoaa , thanks for your reply!  So it sounds like I should define `${job}` and `${PBS_JOBID}` at the job card level, then pass those two variables into the J-jobs and use it for the `${jobid}` variable, which is defined as `${job}.${PBS_JOBID}`.  I can do this soon.

MatthewPyle-NOAA:
@BenjaminBlake-NOAA Is this one also complete now?

BenjaminBlake-NOAA:
[...]

PR #535, opened 2024-10-18
Update jobid definition for all J-jobs

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR builds off of #512 and #524 
- In the NCO implementation standards, `$jobid` needs to be used when setting the run directory names. `$jobid` is defined at the job card level as `$job.$PBS_JOBID`. `$job` is the `<jobname>` tag in Rocoto (and whatever the equivalent is in ecflow), and `$PBS_JOBID` is a sequence of letters/numbers that is specific to each individual task. In this PR, we introduce a temporary variable called `$taskid` which is defined at the job card level in the Rocoto xml, and is passed into the J-jobs where `$jobid` is set and used to define the run directory (`$DATA=$DATAROOT/$jobid`).
- Note: At the moment we shouldn't switch entirely to using `$jobid = $job.$PBS_JOBID` because then other jobs in the workflow cannot access the run directories for other tasks. Once workflow changes are made to utilize the umbrella data directory structure described in issue #498  then we can shift to using `$jobid = $job.$PBS_JOBID` everywhere, and the temporary `$taskid` variable will be removed.
- The following line is added to all J-jobs but is commented out for now, because we do not want to switch to the standard `$jobid` definition just yet (see note above): `export jobid=${jobid:-${taskid}}`

## TESTS CONDUCTED: 

Completed a successful spinup cycle with the DA engineering test.  @hu5970 has offered to assist with testing these changes.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [x] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Continues work on issues #24 , #36 , and #525

### DIFF
--- a/jobs/JRRFS_ANALYSIS_ENKF
+++ b/jobs/JRRFS_ANALYSIS_ENKF
@@ -61,13 +61,13 @@ specified cycle.
 
 export pid=${pid:-$$}
 export RUN=${RUN:-rrfs}
+#export jobid=${jobid:-${taskid}}
+export jobid=${taskid}
 
 export CYCLE_TYPE=${CYCLE_TYPE:-prod}
 if [ "${CYCLE_TYPE}" = "spinup" ]; then
-  export jobid=${RUN}_enkfupdt_${OB_TYPE}_spinup_${envir}_${cyc}
   export COMOUT=${COMOUT:-$(compath.py -o ${NET}/${rrfs_ver}/${RUN}.${PDY}/${cyc}_spinup)}
 else
-  export jobid=${RUN}_enkfupdt_${OB_TYPE}_${envir}_${cyc}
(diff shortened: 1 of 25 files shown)


MatthewPyle-NOAA (approved):
Passes my eyeball test, so will get this merged so it can be tested.

PR #536, opened 2024-10-22
update doc to match the latest situation and add load_prod_util.sh

guoqing-noaa:


MatthewPyle-NOAA (approved):
Looks good to me...

BenjaminBlake-NOAA (approved):
Looks good to me as well.  I've never seen wushdir used before (workflow ush) but seems fine to use.

PR #551, opened 2024-11-19
minor fix to set up GWD links when using RRFS_sas.

hu5970:
Seems we need to add RRFS_sas option in two locations for setting GWD. I do not know if this is matter but does not hurt to set it with reasonable values.

This is compared with setups for winter retro v1.0.3 on Dogwood.


MatthewPyle-NOAA (approved):

BenjaminBlake-NOAA (approved):
Thanks @hu5970 .  These changes should also be made to the main branch since dev-sci development should be mostly frozen at this point.

hu5970:
@BenjaminBlake-NOAA @MatthewPyle-NOAA  I am checking real-time run (v0.9.1) and retro (v1.0.3) to see if there are any local changes need to be committed to dev-sci branch. Then I will commit all recent changes to main. Hopefully this is our latest round of change to the dev-sci.

PR #557, opened 2024-11-27
Run mpassit and UPP task with groups

hu5970:
Run mpassit and UPP task with groups
Change the method used in save_fcst to find history files. 

The system is tested with real-time CONUS 3km MPAS runs.


ush/find_fcst_length.sh
```diff
@@ -10,9 +10,9 @@ array=($1)
 cyc=$2
 fcst_length=$3
 num_fhrs=${#array[@]}
-if [[ $num_fhrs -eq 24 ]]; then
+if (( $num_fhrs == 24 )); then
   icyc=$((10#${cyc}))
-  if [ $icyc -lt $num_fhrs ]; then
+  if (( $icyc < $num_fhrs )); then
```
guoqing-noaa, line 15: Line 13 and 15,  suggest using `${num_fhrs}` and `${icyc}` instead of `$num_fhrs` or `$icyc` to be consistent with other parts on the variable reference style.

scripts/exrrfs_upp.sh
```diff
@@ -1,25 +1,78 @@
...
+${cpreq} ${EXECrrfs}/upp.x .
 ${cpreq} ${FIXrrfs}/upp/* .
 FIXcrtm=${FIXrrfs}/crtm/2.4.0
 while read line; do
   ln -snf ${FIXcrtm}/${line} .
 done < crtmfiles.upp
+#
+#  find cycle time
+#
+YYYYMMDDHH=${CDATE}
```
guoqing-noaa, line 23: 1. Why do we re-define `CDATE` to `YYYYMMDDHH`?
2. `YYYYMMDDHH` is not used anywhere

scripts/exrrfs_upp.sh
```diff
@@ -1,25 +1,78 @@
...
+YYYYMMDD=${CDATE:0:8}
```
guoqing-noaa, line 24: YYYYMMDD is not used

scripts/exrrfs_mpassit.sh
```diff
@@ -37,20 +26,78 @@ elif [[ "${MESH_NAME}" == "conus3km" ]]; then
...
+num_fhrs=${#fhr_all[@]}
+group_total_num=$((10#${GROUP_TOTAL_NUM}))
+group_index=$((10#${GROUP_INDEX}))
+
+for (( ii=0; ii<${num_fhrs}; ii=ii+${group_total_num} )); do
+    i=$(( ii + ${group_index} - 1 ))
+    if (( $i >= ${num_fhrs} )); then
+      break
+    fi
+# get forecast hour and string
```
guoqing-noaa, line 65: Can we indent all the comment lines to align with their corresponding codes?

scripts/exrrfs_upp.sh
hu5970: This is the format used by RRFSv1. It is more clear what the time looks like. I just added those and think about to use them in later update.

scripts/exrrfs_mpassit.sh
```diff
@@ -37,20 +26,78 @@ elif [[ "${MESH_NAME}" == "conus3km" ]]; then
...
+      if [[ -s "./mpassit.${timestr}.nc" ]]; then
+        mv ./mpassit.${timestr}.nc ${UMBRELLA_DATA}${ensindexstr}/mpassit/.
+        mv namelist.mpassit namelist.mpassit_${fhr}
+      else
+        echo "FATAL ERROR: failed to genereate mpassit.${timestr}.nc"
+        err_exit
+      fi
+    else
+      echo "FATAL ERROR: cannot find history file at ${timestr}"
+      err_exit
```
guoqing-noaa, line 101: Discussion on this:  Now one MPASSIT task will process multiple history files, do we want to generate `FATAL ERROR` and exit for the lack of one history file?

scripts/exrrfs_save_fcst.sh
```diff
@@ -45,53 +33,47 @@ else
...
-      mv ${workdir}/diag.${timestr}.nc    ${umbrelladir}/.
-      if (( ${fhr} == ${fcst_len_hrs_thiscyc} )); then
-	echo "Moved the last history file diag.${timestr}.nc. ALL DONE"
-	exit 0
+      mv ${history_file} ${umbrelladir}/.
+      mv ${diag_file}    ${umbrelladir}/.
+      # save to com
+      if (( ${ii} == 1 )); then
+        ${cpreq} ${umbrelladir}/history.${timestr}.nc ${comoutdir}/.
+        ${cpreq} ${umbrelladir}/diag.${timestr}.nc ${comoutdir}/.
```
guoqing-noaa, line 71: Line 69-71: why do we save some history and diag files into COM? Should we save mpasout files instead?

scripts/exrrfs_save_fcst.sh
```diff
@@ -45,53 +33,47 @@ else
...
       fi
+    else
+      echo "ERROR, cannot find diag.${timestr}.nc"
+      err_exit
```
guoqing-noaa, line 75: Same here, do we want to abort the workflow due to the lack of one history file?

scripts/exrrfs_upp.sh
```diff
@@ -38,45 +91,33 @@ numx=2
...
-${cpreq} ${wrftwo} ${COMOUT}${ensindexstr}/upp/${RUN}_two_${CDATE}_f${fhr}.grib2
-${cpreq} ${wrfprs} ${COMOUT}${ensindexstr}/upp/${YYJJJHH}0000${fhr}
+      ${cpreq} ${wrfprs} ${COMOUT}${ensindexstr}/upp/${RUN}_prs_${CDATE}_f${fhr2}.grib2
+      ${cpreq} ${wrfnat} ${COMOUT}${ensindexstr}/upp/${RUN}_nat_${CDATE}_f${fhr2}.grib2
+      ${cpreq} ${wrftwo} ${COMOUT}${ensindexstr}/upp/${RUN}_two_${CDATE}_f${fhr2}.grib2
+      ln -snf  ${COMOUT}${ensindexstr}/upp/${RUN}_prs_${CDATE}_f${fhr2}.grib2 ${COMOUT}${ensindexstr}/upp/${YYJJJHH}0000${fhr2}
+
+    else
+      echo "FATAL ERROR: cannot find mpass file at ${timestr}"
+      err_exit
```
guoqing-noaa, line 121: Line 119-121: same here, we may keep the workflow running even if we miss one mpassit file.

scripts/exrrfs_upp.sh
guoqing-noaa: I don't think it is a good idea to define a new variable that is already defined by the NCO standard.

scripts/exrrfs_upp.sh
hu5970: May use it later.

scripts/exrrfs_upp.sh
```diff
@@ -1,25 +1,78 @@
...
 FIXcrtm=${FIXrrfs}/crtm/2.4.0
 while read line; do
   ln -snf ${FIXcrtm}/${line} .
 done < crtmfiles.upp
+#
+#  find cycle time
+#
+YYYYMMDDHH=${CDATE}
+YYYYMMDD=${CDATE:0:8}
+HH=${CDATE:8:2}
```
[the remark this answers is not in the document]
hu5970: save it for later use

scripts/exrrfs_mpassit.sh
hu5970: Right now, everything is set based on the normal runs. History should available in order. If anyone missing, that could be a problem from model. We will see if we need to add more flexibility during the test.

scripts/exrrfs_save_fcst.sh
hu5970: Right now, everything is set based on the normal runs. History should available in order. If anyone missing, that could be a problem from model. We will see if we need to add more flexibility during the test.

scripts/exrrfs_upp.sh
hu5970: Right now, everything is set based on the normal runs. Mpassit should available in order. If anyone missing, that could be a problem from model. We will see if we need to add more flexibility during the test.

scripts/exrrfs_save_fcst.sh
hu5970: This is testing the function to save mpasout and maybe other files to COM directory. Will replace with MPASSOUT when it is available.

ush/find_fcst_length.sh
hu5970: Done

scripts/exrrfs_mpassit.sh
hu5970: Changed for exrrfs_mpassit.sh, exrrfs_save_fcst.sh, exrrfs_upp.sh

scripts/exrrfs_upp.sh
guoqing-noaa: In the previous RAP/HRRR/RRFSv1 era,  we defined lots of YYYYMMDDHH-ish variables. In my view, these are not good practices because the [NCO standard ](https://www.nco.ncep.noaa.gov/idsb/implementation_standards/ImplementationStandards.v11.0.0.pdf) has defined those built-in variables:
```
CDATE=YYYYMMDDHH
PDY=YYYYMMDD
cyc=HH
cycle=tHHz or tHHMMz
subcycle=MM
```
I don't think we need to introduce new variables for those existing things.

If we want to extract some parts from the date string, such as `2024112700`,  temporarily, we can always use the BASH substring functionality easily, for example, ${CDATE:0:6} will be `202411`.  This will help us stick with the NCO standards. Just my two cents.  

With that said, I am okay with the majority vote, no matter if it is `CDATE` or `YYYYMMDDHH`.

@MatthewPyle-NOAA @BenjaminBlake-NOAA What do you think? Thanks!

hu5970: We can discuss this later and make decision how to update all of the scripts if needed.

guoqing-noaa (approved):
We can merge this to move forward and revisit the `CDATE, PDY...` and `YYYYMMDDHH, YYYYMMDD` issue later.

ShunLiu-NOAA (approved):

scripts/exrrfs_upp.sh
guoqing-noaa: To add to this discussion, `CDATE, PDY, cyc, cycle, subcyc`  have strict definitions and they hold and can only hold the values for the current cycle. 
On the other hand, `YYYYMMDDHH`-ish variables are not strictly tied to the current cycle, people can use them for any previous or future cycles and that will confuse users when reading the scripts.

PR #574, opened 2024-12-06
Update CLEAN and ARCHIVE task in FV3LAM_wflow.xml 

hu5970:
Update CLEAN and ARCHIVE task in FV3LAM_wflow.xml so they will be turned off for retros.
This also adds more dependents for those two tasks. This is important to avoid accidentally running clean task.

### DIFF
--- a/parm/FV3LAM_wflow.xml
+++ b/parm/FV3LAM_wflow.xml
@@ -3749,7 +3749,7 @@ MODULES_RUN_TASK_FP script.
 ************************************************************************
 ************************************************************************
 -->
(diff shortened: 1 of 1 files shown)


ShunLiu-NOAA (approved):

MatthewPyle-NOAA:
Does scripts/exrrfs_archive_retro.ksh exist?  A bit confused if this one is mostly preventing cleaning of old cycles prematurely, or if it is really shutting down the archiving of retro output also.

ShunLiu-NOAA:
There is archive retro job yet. We manually archive data to HPSS daily. But we are discussing if we should add retro-archiving job to workflow.

PR #576, opened 2024-12-09
Changes for using convection_permitting physics for conus12km and fix fcst job err information

chunhuazhou:
## DESCRIPTION OF CHANGES: 
- Set namelist options according to grid selection for convection_permitting physics
- Fix fcst job err information

### DIFF
--- a/scripts/exrrfs_fcst.sh
+++ b/scripts/exrrfs_fcst.sh
@@ -15,6 +15,23 @@ else
   lbc_interval=${ENS_LBC_INTERVAL:-3}
 fi
 #
+# determine time steps and etc according to the mesh
+#
+if [[ ${MESH_NAME} == "conus12km" ]]; then
+  dt=60
+  substeps=2
+  disp=12000.0
+  radt=30
+elif [[ ${MESH_NAME} == "conus3km" ]]; then
+  dt=20
+  substeps=4
+  disp=3000.0
+  radt=15
+else
+  echo "Unknown MESH_NAME, exit!"
+  err_exit
+fi
+#
 # find forecst length for this cycle
 #
 fcst_length=${FCST_LENGTH:-1}
@@ -81,7 +98,5 @@ source prep_step
 ${cpreq} ${EXECrrfs}/atmosphere_model.x .
 ${MPI_RUN_CMD} ./atmosphere_model.x 
 # check the status
-if [[ -f './log.atmosphere.0000.err' ]]; then # has to use '-f" as the 0000 err file may be size 0
-  echo "FATAL ERROR: MPAS model run failed"
-  err_exit
-fi
+ls log.atmosphere.*.err 1>/deve/null 2>&1
+export err=$?; err_chk
--- a/scripts/exrrfs_fcst.sh
+++ b/scripts/exrrfs_fcst.sh
@@ -98,5 +98,8 @@ source prep_step
 ${cpreq} ${EXECrrfs}/atmosphere_model.x .
 ${MPI_RUN_CMD} ./atmosphere_model.x 
 # check the status
-ls log.atmosphere.*.err 1>/deve/null 2>&1
-export err=$?; err_chk
+errfiles=(./log.atmosphere.*.err)
+if [ -e "${errfiles[0]-}" ]; then
+  echo "FATAL ERROR: MPAS model run failed"
+  err_exit
+fi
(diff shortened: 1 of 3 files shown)


hu5970 (approved):

scripts/exrrfs_fcst.sh
```diff
@@ -81,7 +98,8 @@ source prep_step
 ${cpreq} ${EXECrrfs}/atmosphere_model.x .
 ${MPI_RUN_CMD} ./atmosphere_model.x 
 # check the status
-if [[ -f './log.atmosphere.0000.err' ]]; then # has to use '-f" as the 0000 err file may be size 0
+errfiles=(./log.atmosphere.*.err)
+if [ -e "${errfiles[0]-}" ]; then
```
guoqing-noaa, line 102: I did not test this but I think this will work.

I would prefer the other way as we discussed, i.e. using "ls" and `err_chk` which may be a little bit easier to read.

chunhuazhou: Using "ls" returned err code 2, when the forecast jobs ran successfully. I tested this and it is working fine so let's keep it, at least for now.

guoqing-noaa: Suggest replacing lines 101-105 with the follows:

```
if ls ./log.atmosphere.*.err 1> /dev/null 2>&1; then
  echo "FATAL ERROR: MPAS model run failed"
  err_exit
else
  exit 0
fi
```

chunhuazhou: Changed and tested. Thanks!

guoqing-noaa (approved):

scripts/exrrfs_fcst.sh
guoqing-noaa: Thanks, @chunhuazhou!

hu5970:
@MatthewPyle-NOAA Please check this PR. Thanks,Ming

scripts/exrrfs_fcst.sh
MatthewPyle-NOAA: Jumping in on this one a bit late.  I find the "if ls" syntax a bit odd.  

How about something like:

```
numerr=`ls ./log.atmosphere*.err | wc -l`

if [ $numerr -gt 0 ] ; then
echo "FATAL ERROR: MPAS model run failed"
  err_exit
fi 
```

hu5970: I will have a PR right after this one and I can test it with the following a little changed version:

numerr=$(ls ./log.atmosphere*.err | wc -l)

if (( $numerr > 0 )) ; then
echo "FATAL ERROR: MPAS model run failed"
  err_exit
fi

guoqing-noaa: Here is the output from the suggested change `numerr=$(ls ./log.atmosphere*.err | wc -l)`
```
ls: cannot access './log.atmosphere*.err': No such file or directory
```
You still need to add 2>/dev/null, like this `numerr=$(ls ./log.atmosphere*.err 2>/dev/null | wc -l)`

guoqing-noaa: also, if we go this route,  suggest changing `numerr` to `num_log_err` to be a little bit more accurate. Thanks!

hu5970: It works without "2>/dev/null":

```
++ exrrfs_fcst.sh[103]: wc -l
++ exrrfs_fcst.sh[103]: ls './log.atmosphere*.err'
ls: cannot access './log.atmosphere*.err': No such file or directory
+ exrrfs_fcst.sh[103]: numerr=0
+ exrrfs_fcst.sh[104]: ((  0 > 0  ))
+ JRRFS_FCST[53]: export err=0
+ JRRFS_FCST[53]: err=0
+ JRRFS_FCST[53]: err_chk
 completed cleanly
```

Adding "2>/dev/null" should be a little more clean.

hu5970: I have included this change in my PR.

hu5970: The current check is:

```
numerr=$(ls ./log.atmosphere*.err 2>/dev/null | wc -l)
if (( $numerr > 0 )) ; then
  echo "FATAL ERROR: MPAS model run failed"
  err_exit
fi
```

guoqing-noaa: Thanks, Ming!

Sorry for some further picky comments, but I would suggest using `${numerr}` instead of `$numerr`

Also, I am fine with `numerr`, but my first impression will interpret it as "num of errors" or "error number". At the same time, `num_errfiles` or `num_err_log` may be more clear.

hu5970: Those are good improvements. We can make such improvement in the next PR and start to review and test mroe important changes.

guoqing-noaa: I agree, these discussions are not specific to this PR. I don't have objections to merge this PR.

PR #581, opened 2024-12-12
Enhance the MPASSIT task so it works with both CONUS3km and CONUS12km domains

hu5970:
1) Update MPASSIT fix files so all domains use the same list files.
2) Update namelist to use history file for grid so it works with both CONUS3km and CONUS12km domain and different levels. 
3) Save the 1-h MPASOUT file for cycling.

Those changes are tested with CONUS3km and CONUS12km real-time MPAS runs on JET.


chunhuazhou (approved):

BenjaminBlake-NOAA (commented):
@hu5970 The changes look good to me, I just have one question.  Why are the history and diag files no longer copied into COMOUT in exrrfs_save_fcst.sh?

hu5970:
@BenjaminBlake-NOAA The history and diag files are copied to umbrella directory for post process. This change is to copy 1-h MPASOUT instead of history and diag to COMOUT directory for cycling. The original copy of history and diag is a space holder. We need to use 1-h MPASOUT for next cycle.

BenjaminBlake-NOAA (approved):
@hu5970 Thanks, looks good to me.

parm/namelist.mpassit
MatthewPyle-NOAA, line 1: If grid_file_input_grid nd hist_file_input_grid are defined the same, do they need to be two distinct variables?

MatthewPyle-NOAA (commented):
One small question, but can definitely merge it tomorrow.

parm/namelist.mpassit
guoqing-noaa: @MatthewPyle-NOAA That's required by the MPASSIT program.

guoqing-noaa (approved):

parm/namelist.mpassit
hu5970: "grid_file_input_grid" is the file provides grid information. All those grid information are included in history file, which is the file defined by "hist_file_input_grid".  We need to set both the same file for now. But will make a separate "grid_file_input_grid" in the future if we want to reduce the output file size by removing grid information from it.

hu5970:
@MatthewPyle-NOAA Only you can change the PR title, right?

MatthewPyle-NOAA:
> @MatthewPyle-NOAA Only you can change the PR title, right?

I don't know if only I can.  Do you see an "Edit" button off to the right side near the top where the title is in a large font?   I do, and that is how I edit titles.

PR #584, opened 2024-12-13
Use the RRFSv1 method to set up CYCLEDEF

hu5970:
Update the cycledef based on the method from RRFSv1.

This is because it is quite complicated to set up the cycledef in rocoto. We use the same method as RRFSv1 to explicitly define the format for CYCLEDEF. So, both experiment period (set up RUN_PERIOD for both real-time and retro) and CYCLEDEF need to set up explicitly now.

This is tested with real-time MPAS run on JET.


guoqing-noaa (approved):

workflow/exp/exp.ens_conus12km
```diff
@@ -48,7 +47,10 @@ export LBC_EXTRN_MDL_BASEDIR="${COMINgefs}/gefs.@Y@m@d/@H"
 export LBC_EXTRN_MDL_NAME_PATTERN="pgrb2ap5/gep#gmem#.t@Hz.pgrb2a.0p50.ffHHH"
 export LBC_EXTRN_MDL_NAME_PATTERN_b="pgrb2bp5/gep#gmem#.t@Hz.pgrb2b.0p50.ffHHH"
 #
-export RETRO_PERIOD='2024052623-2024052623'
+export RUN_PERIOD="2024052623-2024052623"
+export CYCLEDEF_IC="   &STARTYEAR;&STARTMONTH;&STARTDAY;2300 &ENDYEAR;&ENDMONTH;&ENDDAY;2300 06:00:00 "
```
guoqing-noaa, line 51: Just think of this, do we really need individual year, month, and day?

Can we make it simple using something as follows?
```
export CYCLEDEF_IC="   &START_PDY;2300 &END_PDY;2300 06:00:00 "
```

hu5970: We have another way to set up cycledef which needs to know year, month, day separately.

PR #585, opened 2024-12-18
More changes for running ensemble forecasts

chunhuazhou:
## DESCRIPTION OF CHANGES: 
More changes for running the MPAS ensemble, including modifications to ungrib_ic and ungrib_lbc, adding ensemble for the ic, lbc, prep_ic, prep_lbc, fcst, save_fcst.

## TESTS CONDUCTED: 
Tested conus-12km ensemble runs on Jet.


scripts/exrrfs_lbc.sh
```diff
@@ -76,9 +71,9 @@ knt=0
 for fhr in  ${fhr_all}; do
   EDATE=$($NDATE ${fhr} ${CDATEin})
   timestring=$(date -d "${EDATE:0:8} ${EDATE:8:2}" +%Y-%m-%d_%H:%M:%S)
-  ln -snf ${UMBRELLA_DATA}${ensindexstr}/ungrib_lbc/${prefix}:${timestring:0:13} .
+  ln -snf ${UMBRELLA_DATA}${MEMDIR}/ungrib_lbc${MEMID}/${prefix}:${timestring:0:13} .
 done
-ln -snf ${COMINrrfs}/${RUN}.${PDY}/${cyc}${ensindexstr}/ic/init.nc .
+ln -snf ${COMINrrfs}/rrfs.${PDY}/${cyc}${MEMDIR}/ic/init.nc .
```
guoqing-noaa, line 76: Why we hardcode `${RUN}` to `rrfs`?

scripts/exrrfs_save_fcst.sh
```diff
@@ -26,13 +18,8 @@ echo "forecast length for this cycle is ${fcst_len_hrs_thiscyc}"
...
 # decide the location of run and umbrella
 #
-if [[ -z "${ENS_INDEX}" ]]; then
-  ensindexstr=""
-else
-  ensindexstr="/mem${ENS_INDEX}"
-fi
-workdir="${DATAROOT}${ensindexstr}/${RUN}_fcst_${cyc}"
-umbrelladir="${UMBRELLA_DATA}/${jobid}"
+workdir="${DATAROOT}/${RUN}_fcst${MEMID}_${cyc}"
```
guoqing-noaa, line 21: This was not introduced by this PR but it will be good if we can fix this as early as possible.

We should not define a new variable `workdir`. The NCO implementation standard already has a variable `DATA` which means the same thing as `workdir`.
Also, `DATA` is and should be defined in the J-job:
`export DATA=${DATA:-${DATAROOT}/${jobid}}`
In NCO, it takes the form of `${DATAROOT}/${jobid}`.

scripts/exrrfs_save_fcst.sh
```diff
@@ -26,13 +18,8 @@ echo "forecast length for this cycle is ${fcst_len_hrs_thiscyc}"
...
+umbrelladir="${UMBRELLA_DATA}${MEMDIR}/${jobid}"
```
guoqing-noaa, line 22: Similarly as above, `UMBRELLA_DATA` should be `UMBRELLA_ROOT` and `umbrelladir` should be `UMBRELLA_DATA` which should be defined in the J-Job.

workflow/rocoto_funcs/base.py
```diff
@@ -86,7 +86,6 @@ def header_entities(xmlFile,expdir):
 "\n\
   <envar><name>HOMErrfs</name><value>&HOMErrfs;</value></envar>\n\
   <envar><name>EXPDIR</name><value>&EXPDIR;</value></envar>\n\
-  <envar><name>DATAROOT</name><value><cyclestr>&DATAROOT;/{net}/{rrfs_ver}/{run}.@Y@m@d/@H</cyclestr></value></envar>\n\
```
guoqing-noaa, line 89: We don't need to pass `DATAROOT` to the runtime environment?

workflow/rocoto_funcs/base.py
```diff
@@ -268,13 +267,13 @@ def xml_task(xmlFile,expdir,task_id,cycledefs,dcTaskEnv={},dependencies="",metat
...
-    'nodes': get_cascade_env(f"NODES_{task_id}".upper()),
-    'reservation': get_cascade_env(f"RESERVATION_{task_id}".upper()),
-    'native': get_cascade_env(f"NATIVE_{task_id}".upper())
+    'account': get_cascade_env(f'ACCOUNT_{meta_id}'.upper()),
+    'queue': get_cascade_env(f'QUEUE_{meta_id}'.upper()),
+    'partition': get_cascade_env(f"PARTITION_{meta_id}".upper()),
+    'walltime': get_cascade_env(f"WALLTIME_{meta_id}".upper()),
+    'nodes': get_cascade_env(f"NODES_{meta_id}".upper()),
+    'reservation': get_cascade_env(f"RESERVATION_{meta_id}".upper()),
+    'native': get_cascade_env(f"NATIVE_{meta_id}".upper())
```
guoqing-noaa, line 276: I think we don't need to make changes in lines 270-276.  Have you run into any problems with the original `task_id`?

workflow/rocoto_funcs/fcst.py
```diff
@@ -5,26 +5,37 @@
 ### begin of fcst --------------------------------------------------------
 def fcst(xmlFile, expdir, do_ensemble=False):
   # Task-specific EnVars beyond the task_common_vars
-  dcTaskEnv={}
+  meta_id='fcst'
+  cycledefs='prod'
+  hrs=os.getenv('PROD_BGN_AT_HRS', '3 15')
+  fcst_len_hrs_cycls=os.getenv('FCST_LEN_HRS_CYCLES', '03 03')
```
guoqing-noaa, line 11: Should `fcst_len_hrs_cycls` be `fcst_len_hrs_cycles`?

workflow/rocoto_funcs/base.py
guoqing-noaa: OK, I got it. Now we want to define `DATAROOT` at each task.

workflow/rocoto_funcs/base.py
chunhuazhou: It was using "task_id" but for tasks with longer strings such as ic_m001, it can't find the right resources. On the other hand, we can use "meta_id" which is exactly the task name, such as "ic" and "lbc".

guoqing-noaa: @chunhuazhou Whether you specify `NODES_IC_M001` or not, the `get_cascade_env` function will make sure we will get the correct environmental setting in a reverse cascade way, i.e. if `NODES_IC_M001` is not available, the function will then check `NODES_IC`. The function is designed to have the maximum granularity for potential fine tuning. We don't have to use it at this stage, but let's keep this capability.

scripts/exrrfs_save_fcst.sh
hu5970: Can change it to:
workdir="${DATA}"
for now.

scripts/exrrfs_save_fcst.sh
hu5970: There is no UMBRELLA_ROOT. UMBRELLA space is in DATAROOT. Need to think about how to use the umbrella more. We can leave it as is for now.

workflow/rocoto_funcs/setup_xml.py_my
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
```
guoqing-noaa, line 1: @chunhuazhou 
This file `setup_xml.py_my` is an extra file. Please delete it.

guoqing-noaa (approved):
LGTM. Thanks for all the changes.

hu5970 (approved):

PR #590, opened 2025-01-07
Fix dependencies for warm start

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Minor fixes for warm start runs:
- Fix checkfile for prep_lbc
- Add data dependencies for prep_ic

### DIFF
--- a/scripts/exrrfs_prep_lbc.sh
+++ b/scripts/exrrfs_prep_lbc.sh
@@ -35,7 +35,7 @@ while [[ $n -le 12 ]]; do
   YYYYMMDDlbc=${CDATElbc:0:8}
   HHlbc=${CDATElbc:8:2}
   checkfile=${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}${MEMDIR}/lbc/${last_bdyfile}
-  if [[ -r "${checkfile}" ]];then
+  if [[ -r "${checkfile}" ]]; then
      echo "Found ${checkfile}; Use it as boundary for forecast "
      break
   else
--- a/scripts/exrrfs_prep_lbc.sh
+++ b/scripts/exrrfs_prep_lbc.sh
@@ -35,7 +35,7 @@ while [[ $n -le 12 ]]; do
   YYYYMMDDlbc=${CDATElbc:0:8}
   HHlbc=${CDATElbc:8:2}
   checkfile=${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}${MEMDIR}/lbc/${last_bdyfile}
-  if [ -r ${checkfile} ];then
+  if [[ -r "${checkfile}" ]];then
      echo "Found ${checkfile}; Use it as boundary for forecast "
      break
   else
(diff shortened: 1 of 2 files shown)


scripts/exrrfs_prep_lbc.sh
```diff
@@ -35,7 +35,7 @@ while [[ $n -le 12 ]]; do
   YYYYMMDDlbc=${CDATElbc:0:8}
   HHlbc=${CDATElbc:8:2}
   checkfile=${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}${MEMDIR}/lbc/${last_bdyfile}
-  if [ -r "${checkfile}" ];then
+  if [ -r ${checkfile} ];then
```
guoqing-noaa, line 38: suggest using double brackets. Add one space before `then`

```
if [[ -r ${checkfile} ]]; then
```

Also, I am curious why to remove the double quotes since using them is safer than without them.  I just did a quick test and it worked without a problem.

workflow/rocoto_funcs/prep_ic.py
```diff
@@ -71,6 +72,7 @@ def prep_ic(xmlFile, expdir, do_ensemble=False):
     <and>
       <and>
 {strneqs}
+<datadep age="00:03:00"><cyclestr offset="-{cyc_interval}:00:00">&COMROOT;/rrfs/v2.0.2/rrfsdet.@Y@m@d/@H/fcst/</cyclestr><cyclestr>mpasout.@Y-@m-@d_@H.00.00.nc</cyclestr></datadep>
```
guoqing-noaa, line 75: We may want to make this dependency more flexible as follows:

```
&COMROOT;/&NET;/&rrfs_ver;/&RUN;&WGF;.@Y@m@d/@H/fcst/
```

MatthewPyle-NOAA: I was going to make the same comment as @guoqing-noaa about making it more flexible with regard to version number.

scripts/exrrfs_prep_lbc.sh
chunhuazhou: The old line ` if [ -r "${checkfile}" ];  then` didn't work but if it works with double brackets let's do what you suggested.

scripts/exrrfs_prep_lbc.sh
```diff
@@ -35,7 +35,7 @@ while [[ $n -le 12 ]]; do
   YYYYMMDDlbc=${CDATElbc:0:8}
   HHlbc=${CDATElbc:8:2}
   checkfile=${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}${MEMDIR}/lbc/${last_bdyfile}
-  if [ -r "${checkfile}" ];then
+  if [[ -r "${checkfile}" ]];then
```
guoqing-noaa, line 38: @chunhuazhou We need to add one space before `then`, otherwise it will not work

chunhuazhou: Done. Thanks!

guoqing-noaa (approved):
LGTM. Thanks, @chunhuazhou!

PR #591, opened 2025-01-10
Reduce nodes used for CONUS 12km and update UPP grib2 file name

hu5970:
1) The resources used for running CONUS12km are tuned. All the tasks can be run with only 1 node (ppn=40) on JET.
2) The UPP grib2 file name is updated to match the RRFS v1, which is more close to operation.

Those changes are tested on JET with CONUS12km real-time runs.


MatthewPyle-NOAA (approved):
Looks okay to me, but will let others chime in before merging.

guoqing-noaa (approved):

PR #598, opened 2025-01-23
Update ioda task, RDASApp hash

hu5970:
1) Update J-job, scripts, and python for IODA convertor.
2) Update RDASApp to the latest hash and compile option. 
3) Add option in configure to turn on ioda task
4) add configure for Hera CONUS 12km run

## TESTS CONDUCTED: 
Tested on JET conus 12km real-time runs.


scripts/exrrfs_ioda_bufr.sh
MatthewPyle-NOAA, line 1: It seems like the bufr2ioda.x executable is copied into the run directory, but then actually executes a copy from within the RDASapp.  This is a bit confusing.  Long term the RDASapp executables need to be integrated into $HOMErrfs/exec/

MatthewPyle-NOAA (commented):
One comment about the exrrfs_ioda_bufr.sh script...

scripts/exrrfs_ioda_bufr.sh
hu5970: Agree. right now, bufr2ioda.x cannot be executed from the run directory because of missing link to a dynamic lib. It can be executed if executable within the RDASApp. I do not have time to figure out how to solve this problem but certainly to put it in todo list.

guoqing-noaa (approved):
Thanks for the changes. 

I think we don't need separate "DO_IODA", "DO_JEDI", "DO_CYC" options going forward (current stop-and-start experiment is just an engineering test), one option `DO_DA` or `DO_CYC` will be enough. But we may leave that for future PRs.

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -2,31 +2,37 @@
...
+#"prepbufr_aircraft.yaml" 
+#"prepbufr_ascatw.yaml" 
+#"prepbufr_gpsipw.yaml" 
+#"prepbufr_mesonet.yaml" 
+#"prepbufr_profiler.yaml" 
+#"prepbufr_rassda.yaml" 
+#"prepbufr_satwnd.yaml" 
+#"prepbufr_adpsfc.yaml" 
+#"prepbufr_adpupa.yaml"
+#)
```
delippi, line 29: Apologies for a picky comment. I just wanted to say that I *think* you can just comment out the specific lines you don't want in the yaml_list array instead of commenting the whole thing and creating a separate list. I also opened the file up and noticed there is trailing white space at the end of each of those lines. In fact there is a lot of trailing white space in several other files too. What are the thoughts on trailing white space? It might seem trivial, but seems like not quite the best practice to me. If using vim, you can configure your .vimrc to highlight trailing white space.

MatthewPyle-NOAA (approved):
Thanks for the explanation about the executable.  Will wait to see if you want to address the @delippi comment before merging.

scripts/exrrfs_ioda_bufr.sh
hu5970: Thanks, We can address those issues in the next PR after we can test the changes.

hu5970:
@MatthewPyle-NOAA please merge this one. I will make another one soon. Thanks, Ming

PR #599, opened 2025-01-23
Fix for real-time MPAS runs

chunhuazhou:
## DESCRIPTION OF CHANGES: 
- Changes to have same mpassit and upp outputs as in dev2
-  Add data dependency for mpassit
-  Minor fix to find the lbc files


hu5970 (approved):

workflow/rocoto_funcs/prep_ic.py
```diff
@@ -72,7 +72,7 @@ def prep_ic(xmlFile, expdir, do_ensemble=False):
     <and>
       <and>
 {strneqs}
-<datadep age="00:03:00"><cyclestr offset="-{cyc_interval}:00:00">&COMROOT;/&NET;/&rrfs_ver;/&RUN;&WGF;.@Y@m@d/@H{ensdirstr}/fcst/</cyclestr><cyclestr>mpasout.@Y-@m-@d_@H.00.00.nc</cyclestr></datadep>
+      <datadep age="00:05:00"><cyclestr offset="-{cyc_interval}:00:00">&COMROOT;/&NET;/&rrfs_ver;/&RUN;&WGF;.@Y@m@d/@H/fcst/</cyclestr><cyclestr>mpasout.@Y-@m-@d_@H.00.00.nc</cyclestr></datadep>
```
guoqing-noaa, line 75: I did not check the final XML file, Just curious whether we need the extra empty spaces on line 75?

guoqing-noaa: I would think we need that as we previously used a Python variable `datadep`  and now <datadep> is written out directly.

PR #600, opened 2025-01-24
Add logic for running offline IODA tools.

delippi:
## DESCRIPTION OF CHANGES: 
- Added logic to `exrrfs_ioda_bufr.sh` to run an offline tool to process IODA files. The offline tool adds `MetaData/longitude_latitude_pressure` to the IODA file which is used as the group category for duplicate checking based on [RDASApp#259](https://github.com/NOAA-EMC/RDASApp/pull/259)
- Also cleans up some a few things in the edited file including extra white space.

## TESTS CONDUCTED: 

## ISSUE: 
- https://github.com/NOAA-EMC/RDASApp/issues/258

## Dependencies:
- https://github.com/NOAA-EMC/RDASApp/pull/259
- The above PR needs merged and the submodule hash to RDASApp updated in this repo in order to have access to the offline tool.


MatthewPyle-NOAA (approved):
Looks good to me, but will wait to hear from others before merging...

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -35,9 +31,19 @@ for yaml in ${yaml_list[@]}; do
  ${MPI_RUN_CMD} ${execfile} ${yaml}
  # some data may not be available at all cycles, so we don't check whether bufr2ioda.x runs successfully
 done
-ls ./ioda*nc  
+
+# run offline IODA tools
+${cpreq} ${HOMErrfs}/sorc/RDASApp/rrfs-test/IODA/offline_add_var_to_ioda.py .
+ioda_files=$(ls ioda*nc)
+for ioda_file in ${ioda_files[@]}; do
+  python offline_add_var_to_ioda.py -o ${ioda_file}
```
guoqing-noaa, line 39: I think NCO wants the script (including *.py) to run from the command line directly without the leading "python".

guoqing-noaa: Also, has the Python environment been correctly loaded for this tool?

MatthewPyle-NOAA:
@delippi Did you test within the leading python to make sure it all still works?  If so, think we can merge it.

scripts/exrrfs_ioda_bufr.sh
delippi: @guoqing-noaa, thanks for the comments. The tool already had `#!/usr/bin/env python`  at the top. Its just habit to run it with `python` anyway. I just removed it in my most recent commit and it still works. As for the python environment it seems everything that is needed is already loaded so I didn't have to load anything else.

delippi:
I also just noticed that this `exrrfs_ioda_bufr.sh` is not using the most up to date version of the bufr2ioda yamls. I actually have those saved in RDASApp. I think we ought to have those in a single location, but which repo? @hu5970 @MatthewPyle-NOAA @ShunLiu-NOAA

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -35,9 +31,19 @@ for yaml in ${yaml_list[@]}; do
  ${MPI_RUN_CMD} ${execfile} ${yaml}
  # some data may not be available at all cycles, so we don't check whether bufr2ioda.x runs successfully
 done
-ls ./ioda*nc  
+
+# run offline IODA tools
+${cpreq} ${HOMErrfs}/sorc/RDASApp/rrfs-test/IODA/offline_add_var_to_ioda.py .
+ioda_files=$(ls ioda*nc)
+for ioda_file in ${ioda_files[@]}; do
+  offline_add_var_to_ioda.py -o ${ioda_file}
```
guoqing-noaa, line 39: Sorry for another comment, could you add `./` before `offline_add_var_to_ioda.py`? 
Not all environments put the current directory in the $PATH variable.
_(for example, our Jet role account cron jobs set $PATH to empty)_

guoqing-noaa:
> I also just noticed that this `exrrfs_ioda_bufr.sh` is not using the most up to date version of the bufr2ioda yamls. I actually have those saved in RDASApp. I think we ought to have those in a single location, but which repo? @hu5970 @MatthewPyle-NOAA @ShunLiu-NOAA

@delippi Thanks for noticing this! Yes, we will need to update the bufr2ioda yaml files.

For all PARM files, rrfs-workflow will be self-contained.  That means, RDASApp manages its parm/test files while rrfs-workflow adapts a given version from RDASApp. This adheres to the NCO standard generating PARM files on the fly from the `PARMrrfs`  directory and has the benefit that RDASApp can move forward with any new experimental changes without worrying about affecting rrfs-workflow (similar to our previous practice that the gsiparm.* files are in both the GSI repo and the rrfsv1 repo).

guoqing-noaa:
I would suggest we hold this PR until the discussion in https://github.com/NOAA-EMC/RDASApp/pull/259 is resolved.

delippi:
RDASApp was updated yesterday with https://github.com/NOAA-EMC/RDASApp/pull/259, but in order for this PR to work, the version of RDASApp needs updated to the latest. Should I add that to this PR or should that be done separately?

MatthewPyle-NOAA:
@delippi I would say yes to updating the hash of RDASApp to pick up what is needed within this PR.  Unless others object.

guoqing-noaa:
This PR can be merged after https://github.com/NOAA-EMC/rrfs-workflow/pull/618 is merged,

guoqing-noaa (approved):

MatthewPyle-NOAA:
All good on your end, @delippi ?

delippi:
Wow, thanks for updating the hash to RDASApp @guoqing-noaa. Yeah, I think we are good to go then.

PR #603, opened 2025-01-27
organize EXP files and update build/run instructions

guoqing-noaa:
combine similar exp files into one file
add local files so that one exp file can run on multiple platforms with minimum user edits
add "rt" (realtime) to jet realtime exp file names
remove outdated exp files
update build_and_run.md to match the latest changes


guoqing-noaa:
I would like to provide an example of using the optional "local“ file:

Let's say we stage May 2024 retro data on Jet/Hera/Gaea/Orion for conus12km experiments.

We can gather all data paths in different platforms in `local.conus12km` and keep experiment options in `exp.conus12km`.
At the experiment setup time, we can `cat exp/exp.conus12km exp/local.conus12km > exp.setup` and then only need to modify the run directory, which greatly reduces manual edits.

On the other hand, one can skip the "local" file, just `cp exp/exp.conus12km exp.setup` and then manually edit data paths as well as other local options based on the working platform. 

So there are two ways to prepare the "exp.setup" file. Users can pick either one depending on their situation.

workflow/exp/exp.conus12km
MatthewPyle-NOAA, line 1: Is there a reason why some definitions have 'local.setup' (single quotes) and others have "local.setup" (double quotes)?

MatthewPyle-NOAA (commented):
Mostly looked good to me - did have one question.

workflow/exp/exp.conus12km
guoqing-noaa: Great question! I inherited the original forms and did not change the single/double quotes. Yes, it will be great to be consistent. I can do that in another PR to make the use of single/double quotes consistent in all EXP files. Thanks!

PR #604, opened 2025-01-27
update bufr2 ioda yamls

delippi:
## DESCRIPTION OF CHANGES: 
- Update the prepbufr2ioda yamls to be consistent with what is used in RDASApp (with a few minor differences). The differences include only renaming placeholder(s) to what is expected in the workflow and the path to the prepbufr file as expected in the workflow.
- A few of the files were renamed to the 6 character string mnemonics (for example: `prepbufr_mesonet.yaml` --> `prepbufr_msonet.yaml`)
- A few files were separated out for clarity (for example: `prepbufr_aircraft.yaml` --> 1) `prepbufr_aircar.yaml`, 2) `prepbufr_aircft.yaml`. This will just make future debugging easier.

## TESTS CONDUCTED: 

## ISSUE: 
- Follows https://github.com/NOAA-EMC/RDASApp/pull/183

### DIFF
--- a/parm/prepbufr_adpsfc.yaml
+++ b/parm/prepbufr_adpsfc.yaml
@@ -6,7 +6,7 @@
 observations:
   - obs space:
       name: bufr
-      obsdatain: "./bufr/prepbufr"
+      obsdatain: "./prepbufr"
 
       exports:
         #group_by_variable: prepbufrDataLvlCat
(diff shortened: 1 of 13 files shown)


guoqing-noaa (approved):
LGTM. Thanks for updating these bufr2ioda yaml files!

PR #617, opened 2025-02-06
copy bufr2ioda.x instead of a link to meet the NCO standards

guoqing-noaa:


scripts/exrrfs_ioda_bufr.sh
MatthewPyle-NOAA, line 1: This approach (loading the library in the ex-script) is problematic in the long term.  All modules/libraries should be loaded up front as a job is launched.

MatthewPyle-NOAA (commented):
I appreciate trying to get rid of the linking, but where the library is loaded needs to be done at a higher level in the workflow.

scripts/exrrfs_ioda_bufr.sh
guoqing-noaa: @MatthewPyle-NOAA Thanks a lot for cathing this!  Modified as suggested.

MatthewPyle-NOAA: Thanks @guoqing-noaa.

MatthewPyle-NOAA (approved):
Approving now that my concern has been addressed.

Issue #629, opened 2025-02-10
Generating jedivar yamls


delippi:
For running DA cycled experiments, we need to implement a method to build the jedivar.yaml from the validated yamls in RDASApp. We will eventually want to do this with JCB, but not right now.

It seems the best approach is to generate a template at the time of configuring the workflow, then update that template for each cycle replacing date strings. We shouldn't have to rebuild the jedivar.yaml each cycle... although we could if we needed to for some reason.

PR #630, opened 2025-02-10
Add nonjcb configuration builder 

delippi:
## DESCRIPTION OF CHANGES: 
- Added a tool (`workflow/gen_jedivar_yaml_nonjcb.sh`) to build the jedivar.yaml. This is built prior to running the workflow.
- Minor updates to `scripts/exrrfs_jedivar.sh` for using the preconfigured jedivar.yaml

## TESTS CONDUCTED: 

## ISSUE: 
- https://github.com/NOAA-EMC/rrfs-workflow/issues/629


delippi:
@guoqing-noaa, I made a template PR for building the jedivar yaml from the yamls in RDASApp once we sort out the problem with using all the QC filters. I added a tool to generate the yamls (very similar to RDASApp's gen_yaml.sh). I was also wondering if this is supposed to be python instead of bash?

guoqing-noaa:
@delippi  Thanks for developing this.
However, at this moment, we want to track the jedivar.yaml directly from the parm/ directory.
We still have a lot to go on the workflow side and this is not a top priority at the moment. We may revisit this later. Thanks!

scripts/exrrfs_jedivar.sh
```diff
@@ -101,16 +101,29 @@ fi
...
+
+# Fill in global placeholders
+sed -i \
+    -e "s/@analysisDate@/${analysisDate}/" \
+    -e "s/@beginDate@/${beginDate}/" \
+    -e "s/@HYB_WGT_STATIC@/${HYB_WGT_STATIC}/" \
+    -e "s/@HYB_WGT_ENS@/${HYB_WGT_ENS}/" \
+    -e "s/@length@/${length}/" \
+    -e "s/@DISTRIBUTION@/$distribution/" \
+    ./jedivar.yaml
```
guoqing-noaa, line 126: The changes here break the existing hybrid settings. i.e. We cannot do pure 3DVAR or pure EnVar after this change.

delippi:
> @delippi Thanks for developing this. However, at this moment, we want to track the jedivar.yaml directly from the parm/ directory. We still have a lot to go on the workflow side and this is not a top priority at the moment. We may revisit this later. Thanks!

I figured you would still use that yaml for some time. This is just a draft to get it ready to go when we are ready for it.

workflow/gen_jedivar_yaml_nonjcb.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
delippi: I will move it.

scripts/exrrfs_jedivar.sh
delippi: Do you mean because I removed this part?
```diff
- if [[ "${HYB_WGT_ENS}" == "0" ]]; then # pure 3DVAR
-   sed -i '88,113d' ./jedivar.yaml
- elif [[ "${HYB_WGT_STATIC}" == "0" ]]; then # pure 3DEnVar
-   sed -i '46,87d' ./jedivar.yaml
- fi
```
We will have (and do have in RDASApp) a 3dvar and pure envar template instead of deleting these sections. You would specify this via
```
basic_config="mpasjedi_hyb3denvar.yaml"
```

MatthewPyle-NOAA:
@delippi Should this original version draft PR be deleted now?

PR #632, opened 2025-02-12
Implement the Umbrella directory structure

guoqing-noaa:
- Implement the umbrella directory structure as summarized in the comment below.
- align the dependency indentation in the generated XML files
- always use mapsin.nc as the input stream
- set RESERVATION/CLUSTER/NATIVE to empty strings by default
- Use GNU for JEDIVAR on Hera as GNU-compiled JEDI uses less memory and requests fewer nodes
- update build_and_run.md


guoqing-noaa:
Copy the summary about **UMBRELLA** here:
1. Every task (at each cycle) has an umbrella directory under `${DATAROOT}`, with `${DATA}` located further within the umbrella directory.
2. The umbrella directory consists of two parts:
    2.1). **Task-specific Unique Run Directory**: A subdirectory serving as the unique job directory for that task, referenced by the `${DATA}` variable per NCO standards. This directory is removed upon task completion.

     2.2) **Shared Files and Directories**: Other files and subdirectories under UMBRELLA can be shared across multiple tasks within that cycle. These are removed along with the umbrella directory when the cycle completes.
3. The umbrella directory is referenced by a variable following this naming convention:
           `UMBRELLA_@TASK@_DATA       (e.g. UMBRELLA_IC_DATA)`
4. The umbrella directory name itself follows this naming convention:
           `${DATAROOT}/{RUN}_@task@_${cyc}_${rrfs_ver}/${WGF}${MEMID}`
5. In the development stage, we want to include `PDY` in the umbrella directory name. To do this, we redefine the `DATAROOT` as `DATAROOT/PDY` at the workflow level (`envar` inside a rocoto XML task)

Junjun-NOAA (approved):

chunhuazhou (approved):

scripts/exrrfs_save_fcst.sh
MatthewPyle-NOAA, line 1: The comment about a 10 second wait for files to finish writing - I think eventually we'll want some type of file written in MPAS when the write is done.

MatthewPyle-NOAA (approved):
Looks good to me.

PR #646, opened 2025-02-20
get the version number from the 'VERSION' file under workflow/ so that we don't need to change each exp file separately

guoqing-noaa:


MatthewPyle-NOAA:
@guoqing-noaa My initial thought is whether this could be done using the existing ./versions/run.ver file?

guoqing-noaa:
> @guoqing-noaa My initial thought is whether this could be done using the existing ./versions/run.ver file?

@MatthewPyle-NOAA Thanks for the feedback. To my preliminary understanding, the version/ directory is only used on WCOSS2. `run.ver` mostly contains information for all relevant packages instead of the workflow version. 

I don't have a great solution at the moment. That's why I put the 'VERSION' file under the workflow/ directory, which will be removed once going into the final implementation stage. 

As we previously discussed, the workflow version does not mean the RRFS system version (e.g. we may use the v2.6.8 version of workflow to implement RRFSv2.0).

MatthewPyle-NOAA (approved):
Thanks for the explanation, @guoqing-noaa.  Approving now

guoqing-noaa:
Thanks a lot for your quick review and merge! @MatthewPyle-NOAA

PR #648, opened 2025-02-21
err_exit if no ioda files generated

guoqing-noaa:


scripts/exrrfs_ioda_bufr.sh
MatthewPyle-NOAA, line 1: I would change the WARNING to FATAL ERROR if it is going to trigger an err_exit.

MatthewPyle-NOAA (approved):
Thanks for updating it...

PR #668, opened 2025-03-04
hotfix to assign all missing QualityMarker values to 15

guoqing-noaa:
This PR is to address the real-time MPAS model crash issue due to unreasonably large increments. 
It is caused by observations with missing QualityMarker values, which can not be handled by our current rrfs-workflow MPASJEDI settings.  

We will try to make a permanent solution, but it will take a while.

For now, we will put all missing QualityMarker values to 15 so they can be filtered out by specific obs filters.

The added Python script is based on `offline_add_var_to_ioda.py` developed by Donnie.

Thanks @chunhuazhou, @spanNOAA and @hu5970  for quickly identifying the cause.

### DIFF
--- a/scripts/exrrfs_ioda_bufr.sh
+++ b/scripts/exrrfs_ioda_bufr.sh
@@ -35,10 +35,10 @@ for yaml in ${yaml_list[@]}; do
 done
 
 # run offline IODA tools
-${cpreq} ${HOMErrfs}/sorc/RDASApp/rrfs-test/IODA/offline_add_var_to_ioda.py .
+${cpreq} ${USHrrfs}/offline_ioda_tweak.py .
 ioda_files=$(ls ioda*nc)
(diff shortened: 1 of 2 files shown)


chunhuazhou (approved):

MatthewPyle-NOAA (approved):

delippi:
I see now, the missing QMs are causing those unreasonably large increments. Nice job finding the problem.

I also preferred the word "modify" over "tweak" in the name but not such a big deal.

Can someone send me the path to a run which has the missing QualityMarks for the data? I would like to double check the yaml and also try some modifications that I have in mind for a more robust fix. @guoqing-noaa @chunhuazhou @HuiLiu-NOAA

HuiLiu-NOAA:
@guoqing-noaa :  This can be fixed easily in prepbufr2ioda.yaml by setting the QualityMarker to 15 as default value.

PR #669, opened 2025-03-06
Fix data dependency

spanNOAA:
This PR addresses the issue where the real-time workflow cannot correctly check dependencies due to an incorrect path.
Modify the data dependency path in jedivar, and change IC_OFFSET to 30 in ensemble conus 12km setup.


MatthewPyle-NOAA (approved):

guoqing-noaa (approved):

workflow/exp/rt_jet/exp.rt_jet_ens_conus12km
```diff
@@ -32,7 +32,7 @@ export HISTORY_INTERVAL=1
 export MPASSIT_GROUP_TOTAL_NUM=1
 export UPP_GROUP_TOTAL_NUM=1
 
-export IC_OFFSET=36
+export IC_OFFSET=30
```
guoqing-noaa, line 35: One question, why we want to use 30h fcst instead of 36h now?

chunhuazhou: That is what we used in RRFSv1

chunhuazhou (approved):

workflow/exp/rt_jet/exp.rt_jet_ens_conus12km
chunhuazhou: GEFS foreasts 30h and after has better ensemble spread than shorter forecasts, I believe that was the conclusion.

MatthewPyle-NOAA: A reminder that we did go away from using longer GEFS forecasts to initialize things in RRFSv1 (back in Nov 2023):

"Updated the twice per day ensemble re-initialization to use 7-h GDAS EnKF instead of 30h GEFS EnKF."

guoqing-noaa: Thanks @MatthewPyle-NOAA for the notes!
We cannot use GDAS ensembles at the moment. We will work toward that. Thanks!

PR #672, opened 2025-03-06
Add nonjcb configuration builder (take 2)

delippi:
## DESCRIPTION OF CHANGES: 
- This PR follows what I did in https://github.com/NOAA-EMC/rrfs-workflow/pull/630 several weeks ago, but I tested with updated version of the workflow.
- Added a tool (`workflow/ush/gen_jedivar_yaml_nonjcb.sh`) to build the jedivar.yaml. This is built prior to running the workflow.
  - Very easy to use! just un/comment the obs types you want to be in jedivar.yaml. It will replace the file in parm.
- Minor updates to `scripts/exrrfs_jedivar.sh` for using the preconfigured jedivar.yaml
  - The changes here need to go elsewhere, I just wasn't sure where to put them.
- Minor update to `scripts/exrrfs_ioda_bufr.sh` to copy ioda_empty.nc to data/obs before bufr2ioda step.
  - I need to add the ioda_empty.nc file to the PR. Is it okay to add this binary? It won't ever change. I'm not sure what to do with it otherwise.
- Updates to `ush/yaml_from_parm.sh`. Removes the hardcoded line numbers for deleting parts of the yaml to be a little more robust and not break current functionality. Some reorganization and extra replacement was needed too.

## TESTS CONDUCTED: 

## ISSUE: 
- https://github.com/NOAA-EMC/rrfs-workflow/issues/629
- This replaces https://github.com/NOAA-EMC/rrfs-workflow/pull/630


MatthewPyle-NOAA:
@delippi Would it make sense to store the ioda_empty.nc file in the RRFSv2 fixed files?  A small binary file isn't a huge issue on github, but it isn't considered a best practice with version control systems.  The fact that it won't ever change makes it more acceptable, though.

guoqing-noaa:
I think we don't need the "ioda_empty.nc" files, at least not as the default option. 
We can handle the missing observations easily in the script level.
The JEDI design is that the JEDIVAR.yaml should contain and only contain observers that will be involved in the given DA cycle. We don't want to have an "idle" observer in the super yaml file and it does nothing, which will confuse most people.

delippi:
@MatthewPyle-NOAA, right. I knew it wasn't best practice, but it won't ever change. It could be put in RRFSv2 fixed files. I'm actually not sure how to do that.

@guoqing-noaa, thanks for chiming in. I was just going to mention your thoughts. I was under the impression that we would want a static jedi.yaml with all the obs spaces and only modify things such as analysisDate. For example, we wouldn't want to remove obs spaces just because the ioda file doesn't exist for that ob type. I think this is a good topic to discuss. @ShunLiu-NOAA @hu5970 @SamuelDegelia-NOAA

guoqing-noaa:
@delippi I think we agreed that we have three `YAML_GEN_METHOD`, 1 is for GSL real-time rrfs-workflow, 2 is for your utility and 3 for JCB. It looks like this PR tries to modify `YAML_GEN_METHOD=1`.

delippi:
> @delippi I think we agreed that we have three `YAML_GEN_METHOD`, 1 is for GSL real-time rrfs-workflow, 2 is for your utility and 3 for JCB. It looks like this PR tries to modify `YAML_GEN_METHOD=1`.

Oh, for some reason I though option 2 was with your python tool. Nonetheless, I left the default parm/jedivar.yaml unmodified. My gen_yaml script overwrites the parm/jedivar.yaml, but the user has to actually do that otherwise it uses the default one that you made. I did modify the yaml_from_parm.sh which I don't think breaks anything for the default yaml. I will see what I can do try to use option 2. Also, do you just specify YAML_GEN_METHOD=2 in exp.setup?

guoqing-noaa:
Yes, if you want to use `YAML_GEN_METHOD=2`, just add this to any line of exp.setup.

guoqing-noaa:
> Yes, if you want to use `YAML_GEN_METHOD=2`, just add this to any line of exp.setup.

Sorry I just double checked and this part is somehow NOT in the repo. I will fix this right now.

delippi:
> > Yes, if you want to use `YAML_GEN_METHOD=2`, just add this to any line of exp.setup.
> 
> Sorry I just double checked and this part is somehow NOT in the repo. I will fix this right now.

@guoqing-noaa, should length (e.g., `length: 'PT@length@H'`) and distribution be specified in the exp.setup as well? Right now, I just had them hard coded in the jedivar script--I know that isn't right. I wasn't sure how to handle these two setting. I guess length could be hardcoded in my gen_yamls. The obs distribution probably can't though. What do you think?

guoqing-noaa:
`length` and `distribution` are both fixed in YAML_GEN_METHOD=1.
I think you can hardcode length to 4 as in the method 1, at least for now and a not-short period.

It will also be helpful if you can hardcode `distribution` as well

delippi:
> `length` and `distribution` are both fixed in YAML_GEN_METHOD=1. I think you can hardcode length to 4 as in the method 1, at least for now and a not-short period.
> 
> It will also be helpful if you can hardcode `distribution` as well

Thanks for the suggestions. I will convert these to hard coded values for now. We can always come back and change it later.

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -26,6 +26,22 @@ yaml_list=(
...
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_aircar.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_aircft.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_ascatw.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_gpsipw.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_msonet.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_proflr.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_rassda.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_satwnd.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_sfcshp.nc
+${cpreq} ${PARMrrfs}/ioda_empty.nc ioda_vadwnd.nc
```
guoqing-noaa, line 43: As @MatthewPyle-NOAA mentioned, we need to put these fixed ioda_empty.nc files into ${FIXrrfs}

You can change `${PARMrrfs}` to `${FIXrrfs}/jedi`.  

Also, share me the ioda_empty.nc file and I will take care of the rest part.

delippi: @guoqing-noaa, I appreciate you offering to handle putting the empty ioda file in fix location. I currently have a copy of it on Hera here:
`/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/workflow/v2.0.9.5/parm/ioda_empty.nc`

guoqing-noaa: I staged it into FIX_RRFS2. 

You may do as follows to add it to the workflow repo:
```
cd ${HOMErrfs}/fix/jedi
cp -rp /scratch1/BMC/wrfruc/gge/rrfs-workflow/fix/jedi/ioda_empty.nc .   # need -rp as it is a link in the repo
git add ioda_empty.nc
git commit -m ‘add ioda_empty.nc into FIXrrfs'
```

delippi: I think I got it. Thank you very much for those instructions!

delippi:
@guoqing-noaa or @MatthewPyle-NOAA do either of you have any other comments on this? 

Since the workflow uses YAML_GEN_METHOD=1 as default, this PR won't affect any current configurations. We should consider updating the RDASApp hash once a few PRs are merged. This PR uses the file added in #278, but I've just manually copied it there for my own testing.

- https://github.com/NOAA-EMC/RDASApp/pull/278
- https://github.com/NOAA-EMC/RDASApp/pull/314

ShunLiu-NOAA (approved):
Looks good to me.

guoqing-noaa:
> @guoqing-noaa Are you okay with this one being merged now?

It looks fine to my visual inspection. Thanks!

PR #674, opened 2025-03-07
enable Radiosonde data assimilation and allow users to configure obs types to be assimilated in exp setup files

guoqing-noaa:
Add options `OBS_TYPE_USE` and `OBS_TYPE_REMOVE` to the exp setup files so that users can easily configure what observation types to be assimilated by jedivar and getkf.


guoqing-noaa:
Tested 24 cycles, and it worked as expected.

MatthewPyle-NOAA (approved):
No red flags for me, but will let the DA experts chime in before I merge it.

Junjun-NOAA (approved):

chunhuazhou (approved):

PR #685, opened 2025-03-14
Changes to get dev3 real-time runs close to dev1 real-time runs

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Changes to get dev3 real-time runs close to dev1 real-time runs, including
- adding two physics fix files
- updating mpassit fix files
- include aerosol initialization 
- change ic_offset to 0


fix/mpassit/diaglist
```diff
@@ -1,4 +1,5 @@
 rainnc                   RAINNC
+frainnc                  ACFRAIN
```
guoqing-noaa, line 2: I think this needs the correct UPP version. Can we do it now? @EricJames-NOAA

fix/mpassit/histlist_2d
```diff
@@ -9,12 +9,13 @@ lh				LH
...
 snowfallac                      SNOWFALLAC
 vegfra                          VEGFRA
 shdmin                          SHDMIN
 shdmax                          SHDMAX
 canwat                          CANWAT
 snowc                           SNOWC
-snownc                          SNOWNC
-graupelnc                       GRAUPELNC
+snownc                          ACSNOW
+graupelnc                       ACGRAUP
```
guoqing-noaa, line 20: @EricJames-NOAA Could you double check these two line changes? Thanks!

guoqing-noaa: @chunhuazhou Check this PR:
https://github.com/NOAA-EMC/rrfs-workflow/pull/624/files

It will be good to check with @EricJames-NOAA whether changes to Lines 19-20 are needed. Thanks!

ShunLiu-NOAA (approved):

fix/mpassit/histlist_2d
guoqing-noaa: Thanks, @EricJames-NOAA

fix/mpassit/diaglist
EricJames-NOAA: @guoqing-noaa if the var 'frainnc' is in the MPAS diag files, it should be fine to add this line here.

chunhuazhou: > Thanks @EricJames-NOAA.
> 
> @chunhuazhou If dev1 has this line and you have updated the MPAS-Model to the same as dev1, we should be able to see `frainnc` in the MPAS diag files, right?

Actually mpassit in dev3 couldn't handle frainnc and failed, probably due to different mpassit versions in dev1 and dev3. I reverted the change to mpassit/diaglist so we have no changes to diaglist for now.

hu5970 (approved):

guoqing-noaa (approved):

MatthewPyle-NOAA (approved):

PR #687, opened 2025-03-19
hotfix - getkf uses the background stream to read background state variables

guoqing-noaa:
This is a further hotfix for the GETKF crash issue which has been "temporarily" fixed by PR #686 

To summarize, GETKF needs to read more state variables from the background ensembles besides `rho`.


guoqing-noaa:
Tested using the retro case, running from 2024052700z to 2024052706z

MatthewPyle-NOAA (approved):

SamuelDegelia-NOAA (approved):
Good change, thanks for taking care of this in the workflow.

ShunLiu-NOAA (approved):

PR #698, opened 2025-03-26
add a linter for Shell scripts

guoqing-noaa:
Thank @JacobCarley-NOAA for the great suggestions on adding linters for rrfs-workflow.

This PR is the first step to integrate standard industry language linters for Shell scripts, Python codes and YAML files.

As a starting point, this PR adds a linter for Shell scripts. It works very similarly to what has been implemented in the global workflow. For reference, I have tested this on a PR in my fork: https://github.com/guoqing-noaa/rrfs-workflow/pull/3

`workflow/ush/linter_shellcheck.sh` is added so one can use it as an offline linter before creating a PR to reduce multiple rounds of commits.    
&emsp;&emsp; `usage:  workflow/ush/linter_shellcheck.sh <shell_file>`    

`workflow/ush/linter_shellcheck_all.sh`  is added to facilitate us to modify existing shell scripts to make them compliant with the [ShellCheck standards](https://www.shellcheck.net/wiki/). 
&emsp;&emsp; `usage:  workflow/ush/linter_shellcheck_all.sh`   
I will modify those existing scripts gradually. Anyone interested can join this process.

`workflow/ush/gfs.link_to_nco_namespace.sh` is an example of changes suggested by ShellCheck

I will add Python Linters and YAML linters in the next few PRs.


MatthewPyle-NOAA (approved):
Won't claim to understand all of it, but looks good to me...

PR #700, opened 2025-03-26
enable JEDIVAR for cold-start cycles

guoqing-noaa:
The background files from cold start cycles are `init.nc`, which are different from `restart.nc` or `mpasout.nc` in warm start cycles. mpasedi will crash if we write out the analysis to `init.nc` directly. So we first write out the analysis variables separately and then `ncks` them into `init.nc`

I have some conversations with NCAR DA folks about this crash. A fix will NOT be related to JEDI, but related to the MPAS-Model I/O interface, which will need helps from the NCAR model team.


MatthewPyle-NOAA (approved):

guoqing-noaa:
Tested 24 cycles (2024052700-2024052800), worked as expected.

chunhuazhou (approved):

PR #707, opened 2025-03-27
enable getkf for cold start cycles

guoqing-noaa:
`init.nc` misses a few DA variables needed by GETKF.
We have updated `parm/streams.init_atmosphere` to include those variables. 
Now, GETKF can run on cold start cycles.


SamuelDegelia-NOAA (approved):
Looks good, thanks for updating this @guoqing-noaa.

MatthewPyle-NOAA (approved):

parm/streams.init_atmosphere
```diff
@@ -6,9 +6,24 @@
 
 <immutable_stream name="output"
                   type="output"
-                  filename_template="@output_stream@"
+                  filename_template="none.nc"
```
TingLei-NOAA, line 9: Just curious. what is this change for?

guoqing-noaa:
@MatthewPyle-NOAA I am re-testing this PR after the linter-related changes and will update you once it is good on my side. Thanks!

parm/streams.init_atmosphere
guoqing-noaa: We will use the `output` stream for the next `jedi_ics` stream, but we don't want it to write out any files, so we set `output_interval="none"` for this stream.    
At the same time, we still need to give a unique name to `filename_template` as a placeholder. It can be anything. I just used `none.nc`.  You can understand `none.nc` means nothing, :)

guoqing-noaa:
@MatthewPyle-NOAA Tested 2024052700 to 2024052712. All good. This PR is ready for a merge. Thanks!

ShunLiu-NOAA (approved):

PR #713, opened 2025-03-31
hotfix - add module purge into build.mpassit to solve the compiling failure on Gaea

guoqing-noaa:
build.mpassit failed on Gaea with an Intel internal error.
It was found that it was due to a conflict with some loaded modules in the users' current shell environment.
`module purge` solved this issue.


ShunLiu-NOAA (approved):
adding a description to the PR will make it complete.

guoqing-noaa:
> adding a description to the PR will make it complete.

added. Thanks!

MatthewPyle-NOAA (approved):

PR #718, opened 2025-04-01
Use "Error inflation based on pressure check (setupq.f90)" in realtime runs

guoqing-noaa:
This is a sister PR to PR https://github.com/NOAA-EMC/rrfs-workflow/pull/716.
It uncomments the "Error inflation based on pressure check (setupq.f90)" sections in JEDI YAML files and hence enables us to correctly assimilate q observations in the GSL realtime runs.


ShunLiu-NOAA:
@guoqing-noaa Have we tested this with the RDASApp?

spanNOAA (approved):

HuiLiu-NOAA (approved):
Should work well in the realtime run. I tested it in the retro run.

PR #720, opened 2025-04-01
Enforce using the same MPAS-Model for the workflow and RDASApp by default

guoqing-noaa:
The MPAS-Model is very sensitive to the metadata in NetCDF files. If we use different versions of MPAS (hence different metadata) between the workflow and the RDASApp, most time we will get a crash even if the actual MPAS differences don't matter to DA at all. 

The crash information does not tell us this is because of the mismatch of metadata and we will take time to examine it until finally we find it is due to the mismatch of the MPAS-Model. I have experienced this a few times, and it took me quite a while to figure that out. 

Now that a few other developers have also experienced a similar situation. It is good to add a safeguard to ensure we use the same MPAS-Model commit for both the workflow and the RDASApp unless users want to explicitly use different MPAS versions for some testing (in which case, they can run `build.rdas  different_mpas` or build manually).

This is just a safeguard, and one still needs to make PRs to both rrfs-workflow and RDASApp when updating the MPAS-Model.


MatthewPyle-NOAA (approved):

chunhuazhou (approved):
This is nice!

PR #721, opened 2025-04-01
Add recentering code and task

hu5970:
1) Added the recentering code
2) Added recentering task in workflow.
This is controlled by configure option DO_RECENTER. The default is false.

## TESTS CONDUCTED: 
This is tested on JET with retro.


jobs/JRRFS_RECENTER
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+declare -rx PS4='+ $(basename ${BASH_SOURCE[0]:-${FUNCNAME[0]:-"Unknown"}})[${LINENO}]${id}: '
```
guoqing-noaa, line 2: add the following line after line 1 to suppress some ShellCheck warnings:
```
# shellcheck disable=SC1091,SC2153,SC2154
```

SC1091: ShellCheck cannot follow a "source" command. This is because the differentialShellCheck in the Github action cannot run this script to check whether the source files are available at runtime, so we have to disable this rule.

SC2153: Try to catch typos in variable names, we can add this in the last round of checks as some catches don't make sense.

SC2154: Complain about a variable referenced but not assigned. Such as EXECrrfs, which is from the calling parent script. We can add check this rule in the last round when all other concerns are addressed.

guoqing-noaa:
@hu5970 
You may run this command:
```
workflow/ush/linter_shellcheck.sh   jobs/JRRFS_RECENTER
workflow/ush/linter_shellcheck.sh   workflow/ush/linter_shellcheck.sh scripts/exrrfs_recenter.sh
...
```
to check ShellCheck complains offline

guoqing-noaa:
@hu5970 For Python scripts, you can run the following command
```
workflow/ush/linter_pythonmodify.sh    workflow/rocoto_funcs/recenter.py
```
to do some automatic correction first.

And then run
```
workflow/ush/linter_pythoncheck.sh    workflow/rocoto_funcs/recenter.py
```
to check whether there are any leftover lint complaints about the Python code. 

Unfortunately, there are no tools to help correct the ShellCheck complaints automatically.

Let me know if you want to have a discussion on the Linting. Thanks!

jobs/JRRFS_RECENTER
```diff
@@ -0,0 +1,59 @@
...
```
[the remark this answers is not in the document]
guoqing-noaa: `${id}`  is never used, so it is safe to remove it to address this "warning"

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+declare -rx PS4='+ $(basename ${BASH_SOURCE[0]:-${FUNCNAME[0]:-"Unknown"}})[${LINENO}]${id}: '
```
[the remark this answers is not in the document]
guoqing-noaa: `${id}`  is never used, so it is safe to remove it to address this "warning"

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,72 @@
...
+set -x
+
+cpreq=${cpreq:-cpreq}
+#prefix=${EXTRN_MDL_SOURCE%_NCO} # remove the trailing '_NCO' if any
+#
+# enter the run directory
+#
+cd "${DATA}"
```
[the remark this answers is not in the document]
guoqing-noaa: I used:
```
cd "${DATA}" || exit 1 
```

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,72 @@
...
+done
+
+#-----------------------------------------------------------------------
+#
+# link the control member 
+#
+#-----------------------------------------------------------------------
+#
+mpassinfile="${UMBRELLA_PREP_CONTROL_IC_DATA}/mpasin.nc"
+if [ -r "${mpassinfile}" ] ; then
```
guoqing-noaa, line 37: NCO recommends using  `-s` instead `-r`, i.e
 `if [ -s "${mpassinfile}" ]; then `

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,72 @@
...
+# enter the run directory
+#
+cd "${DATA}" || exit
+
+#start_time=$(date -d "${CDATE:0:8} ${CDATE:8:2}" +%Y-%m-%d_%H:%M:%S) 
+#timestr=$(date -d "${CDATE:0:8} ${CDATE:8:2}" +%Y-%m-%d_%H.%M.%S) 
+#
+# determine whether to begin new cycles and link correct ensembles
+#
+if [[ -r "${UMBRELLA_PREP_IC_DATA}/mem001/init.nc" ]]; then
```
guoqing-noaa, line 17: NCO recommends using  `-s` instead `-r`

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,72 @@
...
+  l_write_mean=.true.
+  l_recenter=.true.
+/
+EOF
+
+# run mpasjedi_enkf.x
+export OMP_NUM_THREADS=1
+ulimit -s unlimited
+ulimit -v unlimited
+ulimit -a
```
guoqing-noaa, line 63: Lines 61-63 may be removed as they have already been set in the top-level script `launch.sh`

scripts/exrrfs_recenter.sh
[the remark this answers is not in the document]
hu5970: Where is id coming from? Do we need it on WCOSS2? It is a warning. If we need id on WCOSS2, let's keep it.

scripts/exrrfs_recenter.sh
hu5970: Done

scripts/exrrfs_recenter.sh
hu5970: Done

scripts/exrrfs_recenter.sh
hu5970: Done

scripts/exrrfs_recenter.sh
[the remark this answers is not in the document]
guoqing-noaa: My bad. I copied this line from the global workflow. But we don't use any id in the RRFS workflow, and I forgot to remove it. `${id}` has been removed from all other j-jobs/ex-scripts.

scripts/exrrfs_recenter.sh
```diff
@@ -0,0 +1,66 @@
...
+  filetail(1)='.nc'
+  numvar(1)=6
+  varlist(1)="qv u w theta tslb q2"
+  l_write_mean=.true.
+  l_recenter=.true.
+/
+EOF
+
+# run mpasjedi_enkf.x
+export pgm="gen_ensmean_recenter.exe"
```
guoqing-noaa, line 60: Is `gen_ensmean_recenter.exe` a Fortran code?  If yes, NCO would like us to add `source prep_step` before running this exe.

guoqing-noaa: Does recenter need `export OMP_NUM_THREADS=1`?  If so, we can set it here.

workflow/rocoto_funcs/setup_xml.py
```diff
@@ -100,6 +101,8 @@ def setup_xml(HOMErrfs, expdir):
             lbc(xmlFile, expdir, do_ensemble=True)
             prep_ic(xmlFile, expdir, do_ensemble=True)
             prep_lbc(xmlFile, expdir, do_ensemble=True)
+            if os.getenv("DO_RECENTER", "FALSE").upper() == "TRUE":
+                recenter(xmlFile,expdir)
```
guoqing-noaa, line 105: add one space before `expdir`:
```
recenter(xmlFile, expdir)
```

hu5970: Done

scripts/exrrfs_recenter.sh
hu5970: added "source prep_step"

scripts/exrrfs_recenter.sh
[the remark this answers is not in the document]
hu5970: removed

jobs/JRRFS_RECENTER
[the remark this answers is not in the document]
hu5970: removed

MatthewPyle-NOAA (approved):
I'm good with it once it gets past the linters.

guoqing-noaa (approved):
The leftover lint comments are as follows:
```
Error: SHELLCHECK_WARNING:
./scripts/exrrfs_recenter.sh:61:11: warning[SC2154]: EXECrrfs is referenced but not assigned.

Error: SHELLCHECK_WARNING:
./scripts/exrrfs_recenter.sh:62:8: note[SC1091]: Not following: prep_step: openBinaryFile: does not exist (No such file or directory)
```
They cannot be avoided by our workflow design (j-job calls ex-scripts) and the use of the differentialShellCheck, which is not able to check whether the sourced file is available in the current environment.

The only way to pass the above two Linter complaints is to add
```
# shellcheck disable=SC1091,SC2154
```
at the top of the scripts after the shebang line.
I am fine with merging as-is or adding the above line.

guoqing-noaa (approved):
@MatthewPyle-NOAA FYI, Ming has resolved the linters' complaints, and this PR is ready for a merge. Thanks!

PR #727, opened 2025-04-03
Use ObsErrorModelHumidity in place of ObsErrorModelStepwiseLinear when processing specific humidity observations

spanNOAA:
This PR fixes an issue where specific humidity wasn't being properly assimilated due to incorrect settings in the specific humidity observation error. The updated jedivar.parm file now uses the ObsErrorModelHumidity function instead of ObsErrorModelStepwiseLinear for specific humidity observations. Additionally, the filter order has been rearranged to ensure the ObsErrorModelHumidity filter works correctly.


guoqing-noaa (approved):
@spanNOAA Thanks a lot for your fabulous work figuring out a fix!

@MatthewPyle-NOAA FYI, this is a hotfix for our real-time runs. Thanks!

SamuelDegelia-NOAA:
I am not aware of how `ObsErrorModelHumidity` differs from `ObsErrorModelStepwiseLinear`. @spanNOAA Would you mind also creating an issue in RDASApp with additional explanation of the problem?

spanNOAA:
> I am not aware of how `ObsErrorModelHumidity` differs from `ObsErrorModelStepwiseLinear`. @spanNOAA Would you mind also creating an issue in RDASApp with additional explanation of the problem?

@SamuelDegelia-NOAA I'm continuing to investigate how various functions impact our analysis outcomes. Our findings thus far indicate that using ObsErrorModelStepwiseLinear prevents us from obtaining meaningful increments/updates in specific humidity measurements. 

image
As you can see in this figure, the specific humidity fitting rate is approximately zero. indicating that humidity observations aren't effectively being assimilated into our analysis. It looks like we messed up by using the wrong error settings for specific humidity. 

image
So we basically used the wrong numbers (even though that's what we were supposed to do) - we put in the error values for relative humidity. When we apply ObsErrorModelStepwiseLinear, JEDI thinks those numbers are for specific humidity because we set "ObsType/specificHumidity" as the variable it should use for error assignment. That's why JEDI ends up having huge error values (like 170-200 g/kg) on the specific humidity observations from aircraft.

When we switch to using ObsErrorModelHumidity instead, the final errors come down to a more reasonable range of 0.03g/kg - 3g/kg (based on what I'm seeing in the JEDI logs, though I don't have plots yet), and we're getting normal increments again. I'll need to run a complete test between now and Monday to make a reliable conclusion on this issue.

delippi:
> > I am not aware of how `ObsErrorModelHumidity` differs from `ObsErrorModelStepwiseLinear`. @spanNOAA Would you mind also creating an issue in RDASApp with additional explanation of the problem?
> 
> @SamuelDegelia-NOAA I'm continuing to investigate how various functions impact our analysis outcomes. Our findings thus far indicate that using ObsErrorModelStepwiseLinear prevents us from obtaining meaningful increments/updates in specific humidity measurements.
> 
> image As you can see in this figure, the specific humidity fitting rate is approximately zero. indicating that humidity observations aren't effectively being assimilated into our analysis. It looks like we messed up by using the wrong error settings for specific humidity.
> 
> image So we basically used the wrong numbers (even though that's what we were supposed to do) - we put in the error values for relative humidity. When we apply ObsErrorModelStepwiseLinear, JEDI thinks those numbers are for specific humidity because we set "ObsType/specificHumidity" as the variable it should use for error assignment. That's why JEDI ends up having huge error values (like 170-200 g/kg) on the specific humidity observations from aircraft.
> 
> When we switch to using ObsErrorModelHumidity instead, the final errors come down to a more reasonable range of 0.03g/kg - 3g/kg (based on what I'm seeing in the JEDI logs, though I don't have plots yet), and we're getting normal increments again. I'll need to run a complete test between now and Monday to make a reliable conclusion on this issue.

Quick question about those results. Was the `ObsErrorFactorPressureCheck` uncommented or was it commented in these experiments?

spanNOAA:
No, the ObsErrorFactorPressureCheck was still applied.

spanNOAA:
> > No, the ObsErrorFactorPressureCheck was still applied.
> 
> Interesting. Based on your explanation I would have guessed that it wasn't used. That function also scales ObsError from RH to q. In all the phase 1-3 validation it seems to work fine. For example where all initial ob errors are specified to be `errors: [0.05912, 0.05912, ... at all levels]`: adpsfc_specificHumidity_187
> 
> I'm pretty sure humidity obs were not being used correctly in the real time since the ObsErrorFactorPressureCheck (OEFPC) was commented out for awhile. I had stated several times that we cannot use these obs until the qsat variable is available in MPAS-JEDI and that commenting the OEFPC section out was not a fix (yes, it will run, but ob errors will not be correct). I'm just trying to make that point clear. What I would suggest is making sure that OEFPC was indeed used as I'm fairly certain that is all that is needed. I don't know what would happen if you used both the ObsErrorModelHumidity and ObsErrorFactorPressureCheck.

You're right. The observation error for q appears appropriate in both the MPAS and FV3 models shown in your histogram. However, in my experiment, it's showing unusual patterns, which I'm currently investigating. Additionally, I couldn't locate any error conversion within the stepwiselinear function. Could you please indicate where this conversion is implemented?

guoqing-noaa:
> 
> You're right. The observation error for q appears appropriate in both the MPAS and FV3 models shown in your histogram. However, in my experiment, it's showing unusual patterns, which I'm currently investigating. Additionally, I couldn't locate any error conversion within the stepwiselinear function. Could you please indicate where this conversion is implemented?

@spanNOAA Could you double check whether you have turned on `ObsErrorFactorPressureCheck` when using Hui's PR?

guoqing-noaa:
@MatthewPyle-NOAA Let's wait for the test results from a few cycles.

delippi:
> Additionally, I couldn't locate any error conversion within the stepwiselinear function. Could you please indicate where this conversion is implemented?

It isn't in `stepwiselinear`. The conversion is in `ObsErrorFactorPressureCheck` [here](https://github.com/JCSDA/ufo/blob/0876a4ba773a3eedc660a7745cd03eb129f6f4af/src/ufo/filters/obsfunctions/ObsErrorFactorPressureCheck.cc#L350)

> @spanNOAA Could you double check whether you have turned on ObsErrorFactorPressureCheck when using Hui's PR (before the change to use ObsErrorModelHumidity)?

@spanNOAA yes, please double check this. And make sure that it isn't inadvertently being commented out by `commentQC.py`. That code wasn't meant to be used anywhere other than my original testing and some ctests "to just have it functioning".

PR #728, opened 2025-04-04
[rrfs-mpas-jedi] Add option to run posterior observer for GETKF and offline domain check

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR adds an option to enable the posterior observer for GETKF. This option allows us to output analysis `hofx` and `oma` statistics in the ensemble `jdiag` files. We can then use this data to compute obs space statistics and verify that the GETKF is working as expected in RRFSv2 (see follow up post).  

In order to use the posterior observer, we also need to use the offline domain check, so that is added and run through `exrrfs_ioda_bufr.sh`.  

Some background information: we currently split the GETKF into a prior observer and solver step. This allows us to use a more efficient observation distribution for the prior observer (`RoundRobin`). However, there is currently no way to split the posterior observer up from the solver and thus they must use the same observation distribution (`Halo`). The Halo distribution is much less efficient for hofx calculation. As such, enabling the posterior observer adds significant runtime to the GETKF solver. Here are comparisons for the RDASApp ctest case: 

```
Solver without the posterior observer: 467 s
Solver with the posterior observer: 866 s
```

I therefore choose to disable the posterior observer by default. 

Note: the Halo observation distribution currently does not work with the `reduce obs space` filter action, so a utility is added to comment out that action if the posterior observed is enabled. I also comment out the Temporal Thinning filter because the `longitude_latitude_pressure` variable seems to have some issues when read in by the solver. But the actual QC performed by these filter/filter actions is already handled by the prior observer. So these filters don’t really do anything for the solver anyways, and the ob counts match between the prior jdiag and posterior jdiag files. You could actually remove all filters for the GETKF solver and it would produce the same result (but I prefer to keep them by default). 

## TESTS CONDUCTED: 
Tested with the default `exp.conus_12km` setup file on Orion. 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [x] Orion
  - [ ] Hercules

## ISSUE: 

None


SamuelDegelia-NOAA:
Here are some diagnostics using the new data output from enabling the posterior observer. We can see that the GETKF is working well with decreases in bias and RMSE for each cycle. The background ensemble is overdispersive, especially near the surface. 

Sawtooth: 


RMSE profiles: 


Bias profiles:

guoqing-noaa:
@SamuelDegelia-NOAA Thank you for this PR! Very helpful!

Could you help me understand the post observer more?
This PR will run the post observer at the same time as the solver, right? Will it be an option to run the post observer in a separate task to save the solver run time as you demonstrated? Will we run the post observer in the final operation?

SamuelDegelia-NOAA:
> @SamuelDegelia-NOAA Thank you for this PR! Very helpful!
> 
> Could you help me understand the post observer more? This PR will run the post observer at the same time as the solver, right? Will it be an option to run the post observer in a separate task to save the solver run time as you demonstrated? Will we run the post observer in the final operation?

You are correct - this PR adds the option to run the posterior observer at the same time as the solver. 

If we wanted to run a separate task for the posterior observer, we would have to use the `mpasjedi_enshofx.x` executable. The problem with this option is that the inputs would have to be the analysis ensemble. As such, the QC would be slightly different from the QC used during DA and to compute the prior jdiag files (i.e., different ob counts between prior and posterior jdiag files). That is okay for statistics like I plotted above but it would cause some difficulties with computing things like obs space increments. So this option isn't necessarily desirable for me. 

Talking with some colleagues, I believe there is an option in development to split the posterior observer similar to how we split the prior observer and solver. With that method, we would use the prior `jdiag` and analysis ensemble as input to posterior observer to get the same ob counts for `ombg` and `oman`. If this actually gets done, we could use it final operations.

scripts/exrrfs_getkf.sh
```diff
@@ -98,6 +98,12 @@ case ${YAML_GEN_METHOD:-1} in
     ;;
 esac
 
+# If running posterior observer during the solver, edit the yaml
+if [[ "${TYPE}" == "solver" ]] && [[ "${GETKF_POST_OBSERVER}" == "TRUE" ]]; then
```
guoqing-noaa, line 102: suggest changing `${GETKF_POST_OBSERVER}`
to
`${GETKF_POST_OBSERVER:-FALSE}`
So we only need to set `GETKF_POST_OBSERVER` in the solver task if needed.

scripts/exrrfs_getkf.sh
```diff
@@ -98,6 +98,12 @@ case ${YAML_GEN_METHOD:-1} in
...
+  ${cpreq} "${USHrrfs}"/yaml_getkf_postobserver .
+  ./yaml_getkf_postobserver getkf.yaml
```
guoqing-noaa, line 104: lines 103 and 104 can be combined to:
`"${USHrrfs}"/yaml_getkf_postobserver getkf.yaml`

NCO allows us to call scripts from `USHrrfs` directly.

guoqing-noaa:
> > @SamuelDegelia-NOAA Thank you for this PR! Very helpful!
> > Could you help me understand the post observer more? This PR will run the post observer at the same time as the solver, right? Will it be an option to run the post observer in a separate task to save the solver run time as you demonstrated? Will we run the post observer in the final operation?
> 
> You are correct - this PR adds the option to run the posterior observer at the same time as the solver.
> 
> If we wanted to run a separate task for the posterior observer, we would have to use the `mpasjedi_enshofx.x` executable. The problem with this option is that the inputs would have to be the analysis ensemble. As such, the QC would be slightly different from the QC used during DA and to compute the prior jdiag files (i.e., different ob counts between prior and posterior jdiag files). That is okay for statistics like I plotted above but it would cause some difficulties with computing things like obs space increments. So this option isn't necessarily desirable for me.
> 
> Talking with some colleagues, I believe there is an option in development to split the posterior observer similar to how we split the prior observer and solver. With that method, we would use the prior `jdiag` and analysis ensemble as input to posterior observer to get the same ob counts for `ombg` and `oman`. If this actually gets done, we could use it final operations.

Thanks for the detailed information. 
Since we remove the `reduce obs space`, will we have a memory and running time issue when we assimilate satellite radiance data?

ush/yaml_finalize
```diff
@@ -99,8 +99,8 @@ if dcConvInfo:  # if non empty
         'ioda_aircar.nc': ['t133', 'q133', 'uv233'],
     }
     for iodafile, keys_to_remove in dcOBS.items():
-        if not os.path.exists(f"data/obs/{iodafile}"):
-            for key in keys_to_remove:
+        for key in keys_to_remove:
+            if not os.path.exists(f"data/obs/{iodafile}") and not os.path.exists(f"data/jdiag/jdiag_{iodafile[5:-3]}_{key}.nc"):
                 dcConvInfo.pop(key, None)  # `None` prevents KeyError if key doesn't exist
```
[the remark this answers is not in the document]
SamuelDegelia-NOAA: I am fine with waiting on #729. It would be good to double check that my additions here don't have issues after #729 is merged.

guoqing-noaa:
> > Thanks for the detailed information. Since we remove `reduce obs space`, will we have a memory and running time issue when we assimilate satellite radiance data?
> 
> The obs used in this combined solver and posterior observer task come from the jdiag files output by the prior observer. Those jdiag files already have obs removed by `reduce obs space` if they fail the QC checks during `getkf_observer`. Thus the set of obs run for the posterior observer should already be thinned and we should not have memory issues. This is also one benefit of the combined solver and posterior observer task because it ensures that the `omb` and `oma` stats are for the same set of obs and can be combined in one file.
> 
> Note: the prior filters are actually not needed at all for the GETKF solver since the prior observer should have already applied those to the `jdiag` files. They could technically be removed from `getkf_solver.yaml` but I perfer to keep a consistent yaml between the observer and solver.

Ah, good to know. Thanks!

scripts/exrrfs_getkf.sh
SamuelDegelia-NOAA: Just change it, thanks!

scripts/exrrfs_getkf.sh
SamuelDegelia-NOAA: Just change it, thanks!

workflow/rocoto_funcs/getkf.py
```diff
@@ -19,6 +19,7 @@ def getkf(xmlFile, expdir, taskType):
         'PHYSICS_SUITE': f'{physics_suite}',
         'REFERENCE_TIME': '@Y-@m-@dT@H:00:00Z',
         'YAML_GEN_METHOD': os.getenv('YAML_GEN_METHOD', '1'),
+        'GETKF_POST_OBSERVER': os.getenv('GETKF_POST_OBSERVER', 'FALSE').upper(),
```
guoqing-noaa, line 22: Do we want to add `GETKF_POST_OBSERVER` envar only in the solver task?

SamuelDegelia-NOAA: I think it is okay to set it for both the observer and solver. The [added code](https://github.com/NOAA-EMC/rrfs-workflow/blob/a3174aa2b65bfff091214dec1a2986a841fc1427/scripts/exrrfs_getkf.sh#L101-L104) in `exrrfs_getkf.sh` only runs if the task type is the solver. 

Also, if we only want to set it for the solver then we would have to add some more logic to `rocoto_funcs/getkf.py` now that we merged the rocoto funcs for the observer and solver.

ush/yaml_finalize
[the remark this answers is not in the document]
SamuelDegelia-NOAA: Just dealt with conflicts and tested this PR again. Everything works after #729 was merged and there are no longer any changes needed to `ush/yaml_finalize`.

workflow/rocoto_funcs/getkf.py
guoqing-noaa: @SamuelDegelia-NOAA 
When `GETKF_POST_OBSERVER` is `true`,  the getkf_observer task will be passed an environmental variable `GETKF_POST_OBSERVER=TRUE`, which actually means nothing to the getkf_observer task. The new users may get confused when they see this env variable for the getkf_observer task,  and may be puzzled whether it should be set to FALSE for the getkf_observer task. To avoid any of that kind of confusion,  the NCO standards recommend only passing the required variables into a task. But this is not strictly enforced. We may live with it or make changes in the future.

If you want to make a change, we dont' need to add extra logic, you may move lines 9-line12 to after line 25, remove line 22, and put it inside the `elif taskType.upper() == "SOLVER":` logic. Thanks!

SamuelDegelia-NOAA: Ah okay, I will set `GETKF_POST_OBSERVER` only for the solver if it is an NCO requirement. I also wasn't aware that you could append a single environment variable in `dcTaskEnv` like that. I thought the whole block of environment variables had to be set at once.

workflow/exp/exp.ens_conus12km
```diff
@@ -19,6 +19,7 @@ export ZETA_LEVELS="L60.txt"
 #
 export DO_IODA="false"
 #export DO_JEDI=true
+export GETKF_POST_OBSERVER="false"
```
guoqing-noaa, line 22: @SamuelDegelia-NOAA I am sorry I did not notice this earlier.

Could you change `"false"` to `false` without double-quoting for lines 20 and 22?

workflow/exp/rt_jet/exp.rt_jet_ens_conus12km
```diff
@@ -21,6 +21,7 @@ export ZETA_LEVELS="L60.txt"
 #
 export DO_IODA="false"
 export DO_JEDI=true
+export GETKF_POST_OBSERVER="false"
```
guoqing-noaa, line 24: similarly for lines 22 and 24

workflow/exp/exp.ens_conus12km
SamuelDegelia-NOAA: Done. I also removed the double quotes from the `DO_RECENTER` setting.

workflow/exp/rt_jet/exp.rt_jet_ens_conus12km
SamuelDegelia-NOAA: Done.

guoqing-noaa (approved):
LGTM. @SamuelDegelia-NOAA Thanks a lot for addressing my picky comments!

MatthewPyle-NOAA (approved):

PR #738, opened 2025-04-14
distinguish crtm used by UPP and JEDI

guoqing-noaa:
I don't know why but UPP and JEDI use different CRTM files even though they are both 2.4.0. @EricJames-NOAA @hu5970 Chime in if you have more information on this.

This PR is to rename the current `crtm/2.4.0` to `crtm/2.4.0_upp` and add `crtm/2.4.0_jedi` so that they don't affect each other.


MatthewPyle-NOAA (approved):
Hopefully this need for distinction between the JEDI and UPP versions of the fix/crtm files is temporary?

xyzemc (approved):
Thanks for correcting the crtm_jedi coefficient.

PR #745, opened 2025-04-17
allow running different YAML configurations with the same `HOMErrfs`

guoqing-noaa:
@HuiLiu-NOAA expressed a need to run different ztd operator while using the same `HOMErrfs`.

@spanNOAA and I also have the need to test different surface operator while using the same `HOMErrfs`

This PR is to add this capability.


MatthewPyle-NOAA (approved):
Looks okay to me....

spanNOAA (approved):

Junjun-NOAA (approved):

PR #746, opened 2025-04-17
add more granularity for generating selective workflow tasks

guoqing-noaa:
@barlage brought up a good workflow use scenario:

One may want to run the ioda/ic/lbc tasks only once and then use that same copy for multiple retro experiments.   
This happened to me when I did some retro experiments on Gaea using the same set of ioda observation files. At that time, I manually did it. 

Now, Model or Physics developers would like to run 20+ experiments on the same IC/LBC but with different physics.

This PR is to add more granularity for generating selective workflow tasks so as to meet the above needs.

By default, `DO_IC_LBC`, `DO_FCST`and `DO_POST` all take values of `true`, so this PR does not affect any existing workflow capabilities.

If one defines the following variables in the exp setup files, 
```
export DO_IODA=false
export DO_JEDI=false
export DO_POST=false
export DO_SPINUP=false
export DO_FCST=false
export DO_CYC=false
```
one will only run the IC/LBC tasks. The associated workflow may only need to run once to generate the needed IC/LBC data

One can then define the following variables in the exp setup files to only run the forecast related tasks (`prep_ic`, `prep_lbc`, `fcst`, `save_fcst`):
```
export DO_IODA=false
export DO_JEDI=false
export DO_POST=false
export DO_SPINUP=false
export DO_IC_LBC=false
export DO_CYC=false
```

### DIFF
--- a/workflow/config_resources/config.base
+++ b/workflow/config_resources/config.base
@@ -13,6 +13,9 @@ export CLUSTER=${CLUSTER:-""}
 export NATIVE=${NATIVE:-""}
 export MORE_XML_ENTITIES=${MORE_XML_ENTITIES:-false}
 export REALTIME_MONTHS=${REALTIME_MONTHS:-2}
+export DO_FCST=${DO_FCST:-true}
+export DO_POST=${DO_POST:-true}
+export DO_IC_LBC=${DO_IC_LBC:-true}
 
 # ioda_bufr
 export NODES_IODA_BUFR=${NODES_IODA_BUFR:-"<nodes>1:ppn=1</nodes>"}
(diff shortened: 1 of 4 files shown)


guoqing-noaa:
FYI, here is an example script to link staged ioda/ic/lbc data into the com/ directory to multiple RRFS experiments:
```
#!/bin/bash
src=/gpfs/f6/bil-fire10-oar/world-shared/my_staged_data

for pdy in $(seq -w 20240501 20240531); do
  for cyc in $(seq -w 00 23); do
    mkdir -p rrfs.${pdy}/${cyc}
    #ln -snf ${src}/rrfs.${pdy}/${cyc}/ioda_bufr    rrfs.${pdy}/${cyc}
    ln -snf ${src}/rrfs.${pdy}/${cyc}/ic    rrfs.${pdy}/${cyc}
    ln -snf ${src}/rrfs.${pdy}/${cyc}/lbc    rrfs.${pdy}/${cyc}
  done
done
```

Junjun-NOAA (approved):

chunhuazhou (approved):

MatthewPyle-NOAA (approved):

PR #749, opened 2025-04-18
Updates to match v1 benchmarks 00/12Z cold start without DA, but run jedi to monitor obs anyway

delippi:
## DESCRIPTION OF CHANGES: 
A follow up PR to better align the cycling strategy for v2 baselines with what was used for v1 benchmarks. This PR enables running 00/12Z cold starts as usual, but DA is no longer performed at those cycles (a new switch was added for this: `COLDSTART_CYCS_DO_DA`). JEDI is still run at 00/12Z but all obs usually set to assimilate are switched to passivate at these times. There is also a few updates to align with using jedivar.yaml from the exp/config location.

## TESTS CONDUCTED: 
A few cycles on Hera.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [x] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules


scripts/exrrfs_jedivar.sh
```diff
@@ -147,7 +147,7 @@ export err=$?
 err_chk
 #
 # ncks increments to cold_start IC
-if [[ ${start_type} == "cold" ]]; then
+if [[ ${start_type} == "cold" && ${COLDSTART_CYCS_DO_DA} == "true" ]]; then
   var_list="pressure_p,rho,qv,qc,qr,qi,qs,qg,ni,nr,ng,nc,nifa,nwfa,volg,surface_pressure,theta,u,uReconstructZonal,uReconstructMeridional"
   ncks -A -H -v "${var_list}" ana.nc mpasin.nc
   export err=$?
```
delippi, line 153: @ShunLiu-NOAA, this was the part I was referring to this morning. At cold start cycles the analysis file is called `ana.nc` in the jedivar.yaml and a other times the analysis file is `mpasin.nc`. I'm not sure on the exact reason for this, but with the extra check for `COLDSTART_CYCS_DO_DA == "true"` ensures that any changes to `ana.nc` do not get merged into the `mpasin.nc` file. 

In summary, at 00/12Z cold start times, airTemperature, specificHumidity, and winds from aircar (133/233) get set to passivate (aka monitor), JEDI is still run but nothing is impacting the analysis (`ana.nc`), and to make double sure, the result from `ana.nc` does not get applied to `mpasin.nc`. Therefore, `mpasin.nc` will be exactly as if JEDI had truly been skipped.

delippi: Just to follow up. I `nccmp -dm` (compares data and metadata) and `mpasin.nc` was identical before and after running JEDI.

delippi:
Converted to draft to make updates and test for recent PRs changing the exp.setup files.

ShunLiu-NOAA (approved):

PR #758, opened 2025-04-24
set reasonable default data retention hours for retros

guoqing-noaa:
One can set `KEEPDATA=no` so that task run directories can be purged upon the completion of the tasks.
But the umbrella directories are not removed automatically, and they may accumulate lots of disk space if running a long retro or running many retros.

In ecflow, we will have an FSM (File System Management) task that will purge the umbrella directories at each cycle. It will take a while to implement this FSM capability (as FSM will do much more beyond purging the umbrella directories) in the Rocoto workflow. At this time, this is NOT a priority.

But rrfs-workflow already has an alternate "clean" task for real-time runs on Jet, which can be turned on by adding `export DO_CLEAN=true` in the exp setup file. This will set the following default data retention hours for real-time runs.
```
STMP_CLEAN_HRS=24
COM_CLEAN_HRS=120 # 5 days
LOG_CLEAN_HRS=840 # 35 days
```

Apparently, these default values are not appropriate for retros.
This PR is to set reasonable data retention hours for retros as follows:
```
STMP_CLEAN_HRS=24
COM_CLEAN_HRS=2400  # 100 days
LOG_CLEAN_HRS=2400
```
This will reduce users' effort as one may only need to add one line, `export DO_CLEAN=true`, to the exp setup files.

### DIFF
--- a/workflow/config_resources/config.base
+++ b/workflow/config_resources/config.base
@@ -13,6 +13,7 @@ export CLUSTER=${CLUSTER:-""}
 export NATIVE=${NATIVE:-""}
 export MORE_XML_ENTITIES=${MORE_XML_ENTITIES:-false}
 export REALTIME_MONTHS=${REALTIME_MONTHS:-2}
+export FCST_LENGTH=1
 export DO_FCST=${DO_FCST:-true}
 export DO_POST=${DO_POST:-true}
 export DO_IC_LBC=${DO_IC_LBC:-true}
--- a/workflow/exp/exp.conus12km
+++ b/workflow/exp/exp.conus12km
@@ -23,7 +23,6 @@ export HYB_ENS_PATH=""     # if empty, the workflow will try to find ensembles a
 
 export DO_CYC=true
 export CYC_INTERVAL=1
-export FCST_LENGTH=3
 for i in {0..23};    do arr[$i]="03"; done # 3hr fcst
 for i in {0..23..6}; do arr[$i]="12"; done # 12hr fcst every 6hrs
 export FCST_LEN_HRS_CYCLES="${arr[*]}"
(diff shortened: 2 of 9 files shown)


MatthewPyle-NOAA (approved):

ShunLiu-NOAA (approved):

HuiLiu-NOAA:
Thanks @guoqing-noaa. This is great for disk space management!!

workflow/config_resources/config.base
```diff
@@ -18,6 +18,15 @@ export DO_FCST=${DO_FCST:-true}
...
 
+#
+if ${DO_CLEAN:-false} && ! ${REALTIME:-false}; then
+  # when DO_CLEAN is set to true in retros, if STMP_CLEAN_HRS, COM_CLEAN_HRS and LOG_CLEAN_HRS
+  # are not specified, use the following values
+  export STMP_CLEAN_HRS=${STMP_CLEAN_HRS:-24}
+  export COM_CLEAN_HRS=${COM_CLEAN_HRS:-2400}  # 100 days
+  export LOG_CLEAN_HRS=${LOG_CLEAN_HRS:-2400}
+fi
+
```
delippi, line 29: @guoqing-noaa, what is this time relative to? For example, 

1. Is it checking `mtime` in which case it would only keep data that is 24 hours old? 
2. Or is it checking relative to the cycle time that is currently running so that when in retro mode I'll have to cycle through a 24 hour period before it starts deleting? 
3. Or is it comparing against the actual cycle date so that when in retro mode all the cleanup jobs automatically run because we are already 1 year past the retro period?

guoqing-noaa: @delippi Good questions.
The clean task runs each cycle, and the `*_CLEAR_HRS` is w.r.t the cycle time at the running cycle.
If you want to purge stmp more aggressively, you can set `STMP_CLEAN_HRS` to for example, 3 or even 1

PR #760, opened 2025-04-24
Bug Fixes and Updates for Functionality of Recenter Task

spanNOAA:
This PR includes bug fixes and necessary updates to ensure two-way coupling functions correctly after introducing the recentering task. It addresses issues related to incomplete data flow, improper resource setup, and lack of synchronization between components during coupled runs.


workflow/rocoto_funcs/recenter.py
```diff
@@ -21,10 +21,11 @@ def recenter(xmlFile, expdir):
...
   <dependency>
-    <and>{timedep}
-      <metataskdep metatask="prep_ic"/>
-    </and>
-    </dependency>'''
+  <and>{timedep}
+    <datadep age="00:05:00"><cyclestr>&COMROOT;/&NET;/&rrfs_ver;/&RUN;.@Y@m@d/@H/jedivar/det/mpasout.@Y-@m-@d_@H.@M.@S.nc</cyclestr></datadep>
+    <metataskdep metatask="prep_ic"/>
+  </and>
+  </dependency>'''
```
SamuelDegelia-NOAA, line 28: If running ensemble DA, do we also need a dependency on the GETKF solver task? That way recentering doesn't run before the ensemble members are updated.

spanNOAA: "That's a great point. I'm currently running a coupling experiment without GETKF, but I'll address this soon.

scripts/exrrfs_jedivar.sh
```diff
@@ -159,3 +159,4 @@ fi
 cp "${DATA}"/jdiag* "${COMOUT}/jedivar/${WGF}"
 cp "${DATA}"/jedivar*.yaml "${COMOUT}/jedivar/${WGF}"
 cp "${DATA}"/log.out "${COMOUT}/jedivar/${WGF}"
+cp "${DATA}"/mpasin.nc "${COMOUT}/jedivar/${WGF}/mpasout.${timestr}.nc"
```
guoqing-noaa, line 162: Will we need the mpasout file for one cycle or different cycles?
If we only need it for one cycle, we only need to copy it to the UMBRELLA directory.

workflow/rocoto_funcs/base.py
```diff
@@ -257,21 +257,22 @@ def get_required_env(env_name):
 def get_cascade_env(env_name):
     seperator = "_"  # underscore
     revStr = env_name[::-1]  # reverse the string
-    env_value = os.getenv(env_name)
+    env_value = os.getenv(env_name) or None
```
guoqing-noaa, line 260: I don't think we need to modify this function. Could you show me what error triggers your changes here?

workflow/rocoto_funcs/recenter.py
```diff
@@ -19,12 +19,22 @@ def recenter(xmlFile, expdir):
         starttime = get_cascade_env(f"STARTTIME_{task_id}".upper())
         timedep = f'\n    <timedep><cyclestr offset="{starttime}">@Y@m@d@H@M00</cyclestr></timedep>'
     # ~~
-    dependencies = f'''
+    if os.getenv("DO_ENSEMBLE", "FALSE").upper() == "TRUE":
```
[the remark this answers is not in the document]
spanNOAA: There are no errors, but when Python attempts to retrieve RESERVATION_CLEAN, it returns an empty string. As a result, the subsequent logics are skipped.

workflow/rocoto_funcs/base.py
spanNOAA: There are no errors, but when Python attempts to retrieve RESERVATION_CLEAN, it returns an empty string. As a result, the subsequent logics are skipped.

scripts/exrrfs_jedivar.sh
spanNOAA: The recentering task always depends on the JEDI output file, meaning this file is required every few hours based on the ensemble tasks interval.

workflow/exp/rt_jet/exp.rt_jet_ens_conus12km
```diff
@@ -52,6 +52,7 @@ export NODES_FCST="<nodes>1:ppn=40</nodes>"
 export NODES_MPASSIT="<nodes>1:ppn=40</nodes>"
 export NODES_UPP="<nodes>1:ppn=40</nodes>"
 export NODES_GETKF="<nodes>4:ppn=20</nodes>"
+export NODES_RECENTER="<nodes>2:ppn=16</nodes>"
```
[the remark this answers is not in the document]
spanNOAA: Yes, I’ll add this to CONUS3km. However, I’m not certain if this resource allocation is optimal for the 3km configuration.

workflow/rocoto_funcs/recenter.py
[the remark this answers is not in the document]
spanNOAA: I’m thinking we might consider running the recenter task only during ensemble analysis in the future. @hu5970 , What do your think about this?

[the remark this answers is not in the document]
spanNOAA: I made a mistake. I should use DO_JEDI, but I used DO_ENSEMBLE instead.

workflow/rocoto_funcs/recenter.py
hu5970: We actually run recenter before EnKF in RRFSv1.

hu5970 (approved):

MatthewPyle-NOAA (approved):

guoqing-noaa (approved):
LGTM. Thanks for addressing the comments, @spanNOAA

PR #762, opened 2025-04-25
add PREP_LBC_LOOK_BACK_HRS to allow configuring exact LBCs for retros

guoqing-noaa:
Let me use an example to explain the change here:

I generate LBC 4 times a day (00, 06,12, 18), cold start at 00 and 12.
For retros, I would expect 12z to use the LBCs from the 12z LBC task.
But in the running time, depending on how fast the 12z LBC task can complete, the PREP_LBC task may use the LBCS either from 00z, 06z, or 12z. This is fine for real-time run, but we would like to use the same prescribed LBCs among different retros to make fair comparisons.

This PR is to change the hardcoded `for hr in range(0, 12):` to `for hr in range(0, PREP_LBC_LOOK_BACK_HRS):` so that users can configure to use exact LBCs for retros.

For retros, `PREP_LBC_LOOK_BACK_HRS` defaults to `CYCLEDEF_LBC_STEP_HRS`.


SamuelDegelia-NOAA:
I'm a little confused on how the dependencies for the `prep_lbc` task work. Is it because the task dependencies are defined as `or` dependencies through `<or>{taskdep}` that it will take whichever LBC is available first? For example: 

```
  <dependency>
  <and>
   <or>
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-0:00:00" />
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-1:00:00" />
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-2:00:00" />
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-3:00:00" />
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-4:00:00" />
     <metataskdep metatask="lbc_m#ens_index#" cycle_offset="-5:00:00" />
   </or>
  </and>
  </dependency>
</task>
</metatask>
```

Does that mean that the `prep_ic` task will launch as soon as any of those lbc files are generated from 0 to 5 h before the current cycle? 

If that is the case and we were looking to force it to use the same LBC for each retro, wouldn't we want to change these to `<and>${taskdep}` at some defined time?

guoqing-noaa:
@SamuelDegelia-NOAA Good question.
We don't generate LBCs hourly. Instead, we generate LBCs every 6 or 12 hours. So, for example, the LBC task at 00z will serve the 00, 01, 02, 03, 04, 05, cycles,  and they will need to depend on the 00z LBC task using a different `cycle_offset`. So that is an `<or>` instead of an `<and>`.

guoqing-noaa:
For real-time, we may look back 12, 24 hours, or even longer. So to be consistent for both retro and real-time, we use <or> instead of depending on an lbc task at a given time.

Feel free to chime in if you have a better idea to handle this situation. Thanks!

SamuelDegelia-NOAA:
Ahhh I see, okay that makes sense now. Thanks for the change.

Junjun-NOAA (approved):

HuiLiu-NOAA (approved):
Thanks for the improvement!

PR #763, opened 2025-04-27
Add three utilities to use the `Herbie` Python package to download gfs/gefs/rap data from AWS

guoqing-noaa:
Previously, we used `wget` to download gfs/gefs/rap data from AWS. It turned out that we got quite a few sporadic, bad, incomplete files, and they crashed the model forecasts. It took non-trivial effort to sort out all the bad files.

[Herbie](https://herbie.readthedocs.io/en/stable/) is a popular Python package for downloading NWP model data from AWS. It is more reliable in terms of downloading a complete copy of grib2 files.

This PR is to add three utilities to use the `Herbie` Python package to download gfs/gefs/rap data from AWS.

Example:
```
source workflow/ush/load_bokeh.sh
cd workflow/ush
herbie_get_GFS_from_aws  20240501-20240501  0-3
```

### DIFF
--- a/modulefiles/BOKEH/gaeaC6.lua
+++ b/modulefiles/BOKEH/gaeaC6.lua
@@ -1,0 +1,19 @@
+help([[
+Load environment for running BOKEH.
+]])
+
+local pkgName    = myModuleName()
+local pkgVersion = myModuleVersion()
+local pkgNameVer = myModuleFullName()
+
+conflict(pkgName)
+
+prepend_path("MODULEPATH", '/gpfs/f6/bil-fire10-oar/world-shared/gge/Miniforge3/modulefiles')
+
+load("Miniforge3/24.11.3-2")
+load("bokeh/3.7.0")
+
+whatis("Name: ".. pkgName)
+whatis("Version: ".. pkgVersion)
+whatis("Category: BOKEH")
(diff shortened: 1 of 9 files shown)


MatthewPyle-NOAA:
I'm guessing that Herbie won't be available on WCOSS2, at least in the short term.  Will that create any problems?  It seems like this PR just adds the new utility, but doesn't impact existing functionality.

guoqing-noaa:
> I'm guessing that Herbie won't be available on WCOSS2, at least in the short term. Will that create any problems? It seems like this PR just adds the new utility, but doesn't impact existing functionality.

@MatthewPyle-NOAA Thanks for the information about WCOSS2. 
No, this will NOT create any problem for WCOSS2. As you mentioned, this PR adds additional utilities to download gfs/gefs/rap etc model data from AWS for users to stage retro IC/LBC data, and does not affect existing functionality. 

One can still use the traditional methods to stage IC/LBC data (but it may need more robust exception handling and examination, especially for a long period of data)
And if needed, we can stage data on RDHPCS and then transfer the staged data to WCOSS2.

Junjun-NOAA (approved):

MatthewPyle-NOAA (approved):

PR #764, opened 2025-04-28
Implement `HYB_ENS_PATH`, allow specifying an external path for BEC ensembles

guoqing-noaa:
This PR implements the long-awaited `HYB_ENS_PATH`, which allows specifying an external path for BEC ensembles and hence different retros can use the same ensemble BEC for fair comparisons.


Junjun-NOAA (approved):

SamuelDegelia-NOAA (approved):
LGTM. Simple but nice change.

SamuelDegelia-NOAA:
Maybe not a decision for this PR, but are we planning to stage a set of BEC files to use with the retros?

guoqing-noaa:
> Maybe not a decision for this PR, but are we planning to stage a set of BEC files to use with the retros?

Good idea!  

I can transfer my current copy of GEFS static BEC to orion/hera.
But this copy of static BEC crashed my retro at 20240508_07z. I have not figured out why yet.

PR #776, opened 2025-05-03
implement `HYB_ENS_PATH` to rrfs ensembles so as to be able to use external ensembles

guoqing-noaa:
PR #764 implemented `HYB_ENS_PATH` for interpolated GEFS/GDAS ensembles.
This PR extends that to rrfs ensembles so that we can use one copy of external rrfs ensembles for multiple retros.


HuiLiu-NOAA (approved):
Thanks for the update

chunhuazhou (approved):

PR #784, opened 2025-05-09
[rrfs-mpas-jedi] Update modulefiles and submodules hashes to build rrfs-workflow on WCOSS2

### LINKED ISSUE #773: [rrfs-mpas-jedi] Update submodules or module files to allow rrfs-workflow to be built and run on WCOSS2
We are now testing building `rrfs-mpas-jedi` on WCOSS2. RDASApp and UPP can be successfully built as is, but the other submodules will need some updates. This issue is opened to track the progress of updating these submodules to allow rrfs-workflow to be built and run on WCOSS2. 

Building progress
- [x] `RDASApp`: 
- [x] `MPAS`: 
- [x] `MPASSIT`: 
- [x] `UPP`: 
- [x] `RRFS_UTILS`: 
- [x] `WPS`: 

Running progress: 
- [x] `RDASApp`: runs successfully
- [x] `MPAS`: runs successfully after recompiling with `ftn` target
- [x] `MPASSIT`: runs successfully after updating build script
- [x] `UPP`: runs successfully after updating modulefile
- [x] `RRFS_UTILS`: runs successfully
- [x] `WPS`: runs successfully

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR introduces support for building `rrfs-workflow` version 2 on WCOSS2.

Changes include:
- Updated `versions/build.ver` and `modulefiles/rrfs/wcoss2.intel.lua` to use `spack-stack-v1.6.0` module versions, consistent with those used on other RDHPCS platforms.
- Updated hashes for `MPASSIT` and `RRFS_UTILS` to reflect recent updates to their modulefiles and/or build scripts.

With these changes, all components of `rrfs-workflow` now compile successfully on WCOSS2.

Note: this PR **only enables building** rrfs-workflow on WCOSS2; runtime functionality has not yet been tested. Further updates will be needed as workflow testing progresses.

## TESTS CONDUCTED: 
Tested building on WCOSS2. Also tested building on Hera to confirm that these updates do not break things elsewhere. 

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [x] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

## ISSUE: 
Partially resolves #773


versions/build.ver
```diff
@@ -1,44 +1,47 @@
...
+export zlib_ver=1.2.13
 
-export modulepath_compiler="/apps/test/hpc-stack/i-19.1.3.304__m-8.1.12__h-1.14.0__n-4.9.2__p-2.5.10__e-8.4.2/modulefiles/compiler/intel/19.1.3.304"
-export modulepath_mpi="/apps/test/hpc-stack/i-19.1.3.304__m-8.1.12__h-1.14.0__n-4.9.2__p-2.5.10__e-8.4.2/modulefiles/mpi/intel/19.1.3.304/cray-mpich/8.1.12"
-export modulepath_scotch="/apps/prod/lmodules/INTEL_cray_mpich/19.1.3.304/cray-mpich/8.1.9"
+export modulepath_compiler="/apps/ops/test/spack-stack-1.6.0-nco/envs/nco-intel-19.1.3.304/install/modulefiles/intel/19.1.3.304"
+export modulepath_mpi="/apps/ops/test/spack-stack-1.6.0-nco/envs/nco-intel-19.1.3.304/install/modulefiles/cray-mpich/8.1.9/intel/19.1.3.304"
+export modulepath_jasper="/apps/ops/test/spack-stack-1.6.0-nco/envs/nco-intel-19.1.3.304/install/intel/19.1.3.304/jasper-2.0.32-c3gaymz"
+export modulepath_netcdf="/apps/ops/test/spack-stack-1.6.0-nco/envs/nco-intel-19.1.3.304/install/intel/19.1.3.304/netcdf-c-4.9.2-bcln2jr"
+export modulepath_pnetcdf="/apps/ops/test/spack-stack-1.6.0-nco/envs/nco-intel-19.1.3.304/install/intel/19.1.3.304/parallel-netcdf-1.12.2-77j7vnx"
```
guoqing-noaa, line 47: Just curious,  when we load `modulefiles/rrfs/wcoss2.intel.lua`, do we have any env variables which already point to the above hardcoded paths? I think we may have. If so, we can set all these modulepath variables to the corresponding env variables. This may mitigate the hardcoded paths here. Thanks!

SamuelDegelia-NOAA: We have to source `build.ver` first to tell `modulefiles/rrfs/wcoss2.intel.lua` which version of the modules to load. It would be easier to just set the versions in `wcoss2.intel.lua` similar to how we do it on other machines. But I think using `build.ver` to define the version numbers is a EE2 requirement, so I kept this same method as was already present in the repo.

guoqing-noaa: Ah, got it. I had thought the reverse way. Thanks for the information.

guoqing-noaa (approved):

versions/build.ver
MatthewPyle-NOAA, line 1: Was curious about all of the version changes, particularly the ones like cray_mpich_ver that seem to be shifting toward an older version.  Is this all a consequence of using spack-stack 1.6.0?

MatthewPyle-NOAA (commented):
Had one small question, but otherwise looks good to me.

versions/build.ver
SamuelDegelia-NOAA: Yes, the module version changes are representative of those included with spack-stack 1.6.0. This spack-stack version only includes version 8.1.9 of `cray-mpich` (now loaded through `stack-cray-mpich` instead of `cray-mpich`. It looks like the only other module that was downgraded was `g2tmpl` from 1.12.0 to 1.10.2. 

If this is a problem then I can try rebuilding with the later version of `cray-mpich`, but it might be better to keep the versions from spack-stack since those are (hopefully) guaranteed to all work together.

modulefiles/rrfs/wcoss2.intel.lua
```diff
@@ -5,27 +5,24 @@ NOAA operational machine WCOSS2 (Catcus/Dogwood)
...
 load(pathJoin("wgrib2", os.getenv("wgrib2_ver")))
-
+setenv("Jasper_ROOT",os.getenv("modulepath_jasper"))
+setenv("NETCDF",os.getenv("modulepath_netcdf"))
 setenv("PNETCDF",os.getenv("modulepath_pnetcdf"))
-
-setenv("CMAKE_C_COMPILER","mpiicc")
-setenv("CMAKE_CXX_COMPILER", "mpiicpc")
-setenv("CMAKE_Fortran_COMPILER", "mpiifort")
+setenv("CMAKE_C_COMPILER","cc")
```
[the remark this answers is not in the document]
TingLei-NOAA: @SamuelDegelia-NOAA  thanks for this explanation.  We could revisit this general issue later.  Then, a specific question is : why , for example, "CC" is used to replace "mpicc" ?   The latter seems the normal ones we use, if we need to specify ones explicitly.

MatthewPyle-NOAA (approved):

PR #792, opened 2025-05-13
allow creating and copying ZTD and satellite ioda files in all RDHPCS platforms

guoqing-noaa:
The Python-based ioda converters for ZTD and satellite observations use functions from the [`wxflow`](https://github.com/NOAA-EMC/wxflow) package.
This PR includes the wxflow package as a submodule so that we can generate ZTD and satellite ioda files on all supported RDHPCS platforms.

This PR also adds ZTD, abi_g16, and abi_g18 into `ush/copy_obs.sh`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,3 +27,7 @@
 	path = sorc/RRFS_UTILS
 	url = https://github.com/hu5970/RRFS_UTILS.git
         ignore = all
+[submodule "ush/wxflow"]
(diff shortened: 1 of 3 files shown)


HuiLiu-NOAA (approved):
Thanks!

xyzemc (approved):

Junjun-NOAA (approved):

PR #46, opened 2023-08-23
Move build-related scripts to sorc directory

chan-hoo:
## DESCRIPTION OF CHANGES: 

- Move the build-related scripts such as CMakeLists.txt, devbuild.sh, and Externals.cfg to the `sorc` directory.
- Rename `devbuild.sh` to `app_build.sh`.
- Update README and Read-the-Docs files.
- Build steps:
```
git clone -b feature/build https://github.com/chan-hoo/rrfs-workflow
cd rrfs-workflow/sorc
./manage_externals/checkout_externals
./app_build.sh -p=hera
```

## TESTS CONDUCTED: 

- On machines/platforms:

- [ ] WCOSS2
- [x] Hera
- [ ] Orion
- [ ] Jet

- Test cases: 

- [ ] Non-DA engineering test
- [ ] DA engineering test
- [ ] Other sample scripts:

## ISSUE: 

- Fixes the issue(s) mentioned in #45

PR #794, opened 2025-05-14
Adds GSI-style ramping in pressure check function

spanNOAA:
Temporarily adds support for GSI-style ramping in surface observation data assimilation when vertical interpolation is enabled. This implementation extends the established methodology previously used in GSI.

### DIFF
--- a/parm/jedivar.yaml
+++ b/parm/jedivar.yaml
@@ -858,7 +858,7 @@ cost function:
                  # The new feature "surface observation error ramp"
                  # needs to be added to UFO to align with the ramp
                  # options for temperature and humidity in GSI.
-                 #surface observation error ramp: true
+                 surface observation error ramp: true
 
 #         # Error inflation based on errormod (qcmod.f90)
 #         - filter: Perform Action
--- a/sorc/_workaround_/RDASApp/ObsErrorFactorPressureCheck.cc
+++ b/sorc/_workaround_/RDASApp/ObsErrorFactorPressureCheck.cc
@@ -1,0 +1,424 @@
+/*
+ * (C) Copyright 2023 NASA
+ * (C) Copyright 2025 NOAA/NCEP/EMC
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+#include "ufo/filters/obsfunctions/ObsErrorFactorPressureCheck.h"
+
+#include <float.h>
+
+#include <algorithm>
+#include <cmath>
+
+#include "eckit/exception/Exceptions.h"
+#include "ioda/ObsDataVector.h"
+#include "oops/util/Logger.h"
+#include "oops/util/missingValues.h"
+#include "ufo/filters/ObsFilterData.h"
+#include "ufo/utils/Constants.h"
+
+#include "ufo/GeoVaLs.h"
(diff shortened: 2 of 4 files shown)


SamuelDegelia-NOAA:
@spanNOAA Thanks for adding this. Do you have any before/after plots of the observation errors to see if this better matches those from GSI? I assume it would since we have `l_sfcobserror_ramp_t` and `l_sfcobserror_ramp_q` set to true [here](https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/gsi_fix/gsiparm.anl). But it would be nice to see some results just to confirm that everything is working. 

Also, are you planning to open a PR in UFO for this change? I would like to start moving away from these `_workaround_` codes if possible, even though I know the JCSDA repos move fairly slowly. It might also be useful to open an issue in RDASApp to bring attention to this issue.

spanNOAA:
@SamuelDegelia-NOAA , while I don't have direct comparison plots available, the graphics in RDAS issue [#46](https://github.com/NOAA-EMC/RDASApp/issues/46#issuecomment-2449124692) show results with ramping enabled. I can share statistics from our realtime implementation if you'd like to see them. The changes are obvious - median effective error reduced by 50%, and at the 75th percentile, effective error decreased by approximately 80%.

Before ramping:
mean     9.691992e+08
std      2.959094e+09
min      2.258770e+00
25%      2.460933e+00
50%      5.834560e+00
75%      1.336419e+01
max      1.000000e+10
Name: Effective_Errors, dtype: float64

After ramping:
mean     8.780488e+08
std      2.830685e+09
min      2.258589e+00
25%      2.311147e+00
50%      2.383275e+00
75%      2.677892e+00
max      1.000000e+10
Name: Effective_Errors, dtype: float64

guoqing-noaa:
@spanNOAA As @SamuelDegelia-NOAA mentioned, could you create a PR (or draft PR, or even an issue) to the UFO repo?

I would like to clarify the use of workarounds. They are NOT intended to avoid or replace a PR.  
Instead, they are designed to help us avoid any local disk changes. 
Local disk changes are not maintainable, prone to errors and limited to one or two persons who make the local disk changes.  

We should avoid any local disk changes and always start from a frozen version so that we can repeat experiments and collaborate on the same code base instead of telling each developer to add any local changes manually.

It is encouraged to create a PR (even a draft PR or an issue) for a workaround, and add its link along with the workaround so that we can have better tracking.

Here is an example of why we will need workarounds:

I added one line `<var name="dzs"/>` into the DA stream of Registry.xml in https://github.com/ufs-community/MPAS-Model/pull/115. This is a simple, straightforward change. The PR was created on Mar. 29 and merged on April 16, 2025. So it took about 20 days to get merged.  I don't have any complaints about this. I understand each repo has its PR queue and has its priorities.
But our rrfs-workflow does not need to wait for this PR, and we can move forward with this change through our workaround mechanism.

guoqing-noaa:
> I totally understand the need for these workarounds, and I'm fine with them as a temporary solution. That said, I want to make sure we’re actively pushing the corresponding PRs in the JCSDA (or other relevant) repos so we can get those fixes merged as soon as possible. Once the submodules are updated, we should be able to remove the workaround and rely on the upstream changes instead. We've done a good job of cleaning up these temporary fixes in the past—let's keep that up.

Great! Glad that we are on the same page.

SamuelDegelia-NOAA (approved):

PR #798, opened 2025-05-20
[main, bugfix] More efficient netcdf data loading in pre-blending python script

delippi:
## DESCRIPTION OF CHANGES: 
This PR only changes the method for reading netcdf fields, initializing arrays, or removed unused fields and was shown to produce identical results and will free up memory usage and improve run times.
- Six atmospheric tracer fields were removed from the script as they were found to not be necessary to be read in (liq_wat, o3mr, ice-wat, rainwat, snowwat, graupel). Only sphum was needed from those fields.
- changes things like
  - `ud = np.float64(0.0*u_s)` --> `ud = np.zeros_like(u_s, dtype=np.float64)`. 
    - While the first method works, it performs a full element-wise multiplication, creating a temporary array in memory just to zero it out. This results in two full memory traversals: one for the multiplication and one for the conversion to float64. In contrast, np.zeros_like(...) allocates the array directly in a single pass, without any computation, making it both more efficient and clearer in intent.
  - `Atm_delp = 1.0 * delp_cold[:, :, 1:]` --> `Atm_delp = np.zeros_like(delp_cold[:, :, 1:], order="F")`.  
    - The original line creates a C-contiguous copy of the sliced array by multiplying all elements by 1.0, again resulting in an unnecessary full traversal and temporary allocation. This also copies real data, even though the Atm_ fields are only being used as intent(inout) working buffers in the Fortran routine. The updated version allocates a Fortran-contiguous, zero-initialized array in one pass, matching only the needed shape. This ensures compatibility with Fortran, improves runtime performance, and makes the purpose (zero-init a work array) more explicit and maintainable.

## TESTS CONDUCTED: 
Tests were run on Cactus in one-of experiments (i.e., just member 001) using the same job card for make_ics except I explicitly requested 500gb of memory (each node has 512gb but caused my jobs to sit in queue for too long to test). My thinking is that if we explicitly request the full amount of memory in real-time we could potentially save edge cases that would normally fail. I was thinking they could be failing due to insufficient memory and hence would run out of wall time.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood

## CONTRIBUTORS (optional): 
Ming and Shun made comments/suggestions to improve code and helped test outside of my one-of experiments.

### DIFF
--- a/ush/chgres_cold2fv3.py
+++ b/ush/chgres_cold2fv3.py
@@ -93,8 +93,8 @@ if ColdStartWinds:
     v_w = np.asfortranarray(v_w.transpose())
 
     # Initialize some computed fields to zero
-    ud = np.float64(0.0*u_s)  # (3950, 2701, 66)
-    vd = np.float64(0.0*u_w)  # (3951, 2700, 66)
+    ud = np.zeros_like(u_s, dtype=np.float64)  # (3950, 2701, 66)
+    vd = np.zeros_like(u_w, dtype=np.float64)  # (3951, 2700, 66)
 
     # rotate winds to model d-grid
     chgres_winds.main(gridx, gridy, u_s, v_s, u_w, v_w, ud, vd)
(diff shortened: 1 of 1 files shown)


BenjaminBlake-NOAA (approved):
@delippi The changes look good to me, thanks for your work on this.

ShunLiu-NOAA (approved):
Looks good. Thank you for improving the efficiency of pre-blending script.

PR #803, opened 2025-05-23
[rrfs-mpas-jedi] Updates for running rrfs-workflow on WCOSS2 

### LINKED ISSUE #773: [rrfs-mpas-jedi] Update submodules or module files to allow rrfs-workflow to be built and run on WCOSS2

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR adds config files and various other updates to allow running rrfs-workflow (version 2) on WCOSS2. Results from the 2024052700 retro are compared against results from Hera in #773. The workflow appears to be working as expected on WCOSS2.  

A few notes:
- I had to update MPAS-Model and MPASSIT to build with the cray-specific wrapper compilers (`cc`, `CC`, and `ftn`). These compilers are needed to correctly handle MPI on WCOSS2. IFor MPASSIT, this is handled by a hash update. For MPAS-Model, since we only plan to update the model at certain times, I instead replace `Makefile` using the `_workaround_` method. 
- There were some issues with the sourcing of `versions/build.ver` which overwrote the module versions loaded for UPP. To solve this, I added `versions/unset.ver` which clears these vars before loading `sorc/UPP/modulefiles/wcoss2.lua`.  
- The MPI_RUN_CMD for WCOSS2 needs job configuration info (e.g., `NTASKS`, `PPN`) and thus is defined in `workflow/sideload/launch.sh` instead of in the `exp.setup` file. This means we can probably remove `MPI_RUN_CMD` from the `exp.setup` files in a future update.  
- The spack-stack 1.6.0 installation on WCOSS2 does not include some of the python modules we need for the offline domain check in `ioda_bufr`. So instead we source a python virtual environment used for RRFSv1. 

## TESTS CONDUCTED: 
Ran 24-h of cycling with the default `exp/exp.conus12km` configuration file. Results are shown in [issue #773](https://github.com/NOAA-EMC/rrfs-workflow/issues/773#issuecomment-2904806949). 

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

## ISSUE: 

Resolves #773


sorc/build.mpas
```diff
@@ -28,6 +29,7 @@ module list
 
 cd "${HOMErrfs}/sorc/MPAS-Model" || exit 1
 cp ../_workaround_/MPAS-Model/Registry.xml src/core_atmosphere
+cp ../_workaround_/MPAS-Model/Makefile ./Makefile
```
guoqing-noaa, line 32: Line 32 is only needed for wcoss2, right? If so, suggest moving it to after line 20.

SamuelDegelia-NOAA: Sure, just moved it.

workflow/rocoto_funcs/setup_xml.py
```diff
@@ -42,8 +42,8 @@ def setup_xml(HOMErrfs, expdir):
     source(f"{expdir}/config/config.{machine}")
     source(f"{expdir}/config/config.base")
     #
-    source(f"{HOMErrfs}/workflow/config_resources/config.{machine}")
     source(f"{HOMErrfs}/workflow/config_resources/config.base")
+    source(f"{HOMErrfs}/workflow/config_resources/config.{machine}")
```
guoqing-noaa, line 46: By design, `config.{machine}` will run before `config.base`. 
Does this order cause any issues for wcoss2?

workflow/sideload/launch.sh
```diff
@@ -57,9 +75,15 @@ case ${task_id} in
     module load "build.${MACHINE}.intel"
     ;;
   upp)
+    source "${HOMErrfs}/versions/unset.ver"
```
guoqing-noaa, line 78: Line 78 is only needed by wcoss2 and I think it has been sourced in line 17.

workflow/rocoto_funcs/base.py
```diff
@@ -214,11 +219,13 @@ def wflow_task_part1(self, xmlFile):  # write out part1 which excludes dependenc
...
             text = text + f'  <queue>&QUEUE_DEFAULT;</queue>\n'
-            text = text + f'  <partition>&PARTITION;</partition>\n'
+            if self.dcTaskRes["partition"] != "":
+                text = text + f'  <partition>&PARTITION;</partition>\n'
         else:
             text = text + f'  <account>{self.dcTaskRes["account"]}</account>\n'
             text = text + f'  <queue>{self.dcTaskRes["queue"]}</queue>\n'
-            text = text + f'  <partition>{self.dcTaskRes["partition"]}</partition>\n'
+            if self.dcTaskRes["partition"] != "":
+                text = text + f'  <partition>{self.dcTaskRes["partition"]}</partition>\n'
```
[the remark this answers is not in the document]
guoqing-noaa: Agree. Making it optional will be a cleaner solution.

workflow/rocoto_funcs/setup_xml.py
SamuelDegelia-NOAA: My main motivation for this change was to get rid of the `--exclusive` tag on WCOSS2. If we include that tag in `rrfs.xml` (set via `NATIVE_UPP` in `config.base`), then we get the error: 

```
05/23/25 20:27:49 UTC :: rrfs.xml :: qsub: invalid option -- '-'
qsub: directive error: --exclusive 
Job submit error: 1.
```

Switching the order allowed me to set `export NATIVE_UPP=""` in `config.wcoss2` which removed the `--exclusive` tag. I tested the order change with my cycling on Hera and it worked. But if this is an issue, then we can just set `NATIVE_UPP` individually in each `config.{machine}` file.

workflow/sideload/launch.sh
SamuelDegelia-NOAA: Line 17 sources `run.ver` which loads in the module versions for all tasks. But then we need to unset those versions for the UPP task specifically to prevent loading different versions than what UPP was built with. Line 78 is the only time `unset.ver` is called. 

This is only needed by WCOSS2 but running it won't cause issues on any other machines (unset does nothing if the environment variable is not defined). But I can add an if-block to only run it for wcoss2 just to be clearer.

sorc/build.mpas
```diff
@@ -17,6 +17,8 @@ if [[ "${MACHINE}" == "wcoss2" ]]; then
   if [[ -f "${BUILD_VERSION_FILE}" ]]; then
     source "${BUILD_VERSION_FILE}"
   fi
+  compiler_str="ftn"
+  cp "${HOMErrfs}/sorc/_workaround_/MPAS-Model/Makefile" "${HOMErrfs}/sorc/MPAS-Model/Makefile"
```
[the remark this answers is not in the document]
SamuelDegelia-NOAA: Sure, good idea!

workflow/rocoto_funcs/setup_xml.py
guoqing-noaa: [The config file cascade structure](https://github.com/NOAA-EMC/rrfs-workflow/wiki/The-cascade-config-structure) is expected to go this way:
`   config.${MACHINE} -> config.base -> [config.realtime if it is a realtime run]`
Any variable exported before `config.base` is expected to overwrite the corresponding default values in config.base.
I just checked config.base Line 76:
`export NATIVE_UPP=${NATIVE_UPP:-"--exclusive"}`

I would expect it will work if you define `export NATIVE_UPP=""` in config.wcoss2.  Not sure why it did not.

guoqing-noaa: Line 81 has this:
` 81 export NATIVE_GRAPHICS="--exclusive"`
But the graphics task is only available on Jet at the moment. I think your workflow will not generate the graphics task.

guoqing-noaa:
README.md can be updated to include WCOSS2 now, :)

workflow/rocoto_funcs/setup_xml.py
guoqing-noaa: @SamuelDegelia-NOAA I just did a test and now I understand the situation better.

Could you modify as follows:
1. Change line 76 of config.base to:
`export NATIVE_UPP=${NATIVE_UPP:-""}`
2. add `export NATIVE_UPP="--exclusive"` to other machines under `config_resources`
3. You may remove `NATIVE_UPP` and `NATIVE_GRAPHICS` from config.wcoss2.

Thanks!

SamuelDegelia-NOAA: Just updated it, thanks.

SamuelDegelia-NOAA:
> README.md can be updated to include WCOSS2 now, :)

Done!

modulefiles/prod_util/wcoss2.lua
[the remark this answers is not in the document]
MatthewPyle-NOAA: Thanks for the explanations, @SamuelDegelia-NOAA and @guoqing-noaa

PR #804, opened 2025-05-24
Make the ic task dependency for `prep_ic` be optional based on the `DO_IC_LBC` setting

guoqing-noaa:
PR https://github.com/NOAA-EMC/rrfs-workflow/pull/746 added the granularity to run a customized set of workflow tasks for different experiments.

One can run one copy of ioda/ic/lbc tasks and use it for multiple DA/NWP experiments.
Under this situation, the `prep_ic` will NOT depend on the `ic` task, as IODA/IC/LBC files have already been pre-generated.

We have dropped that ic dependency for pure forecasts if `DO_IC_LBC=false`
This PR applies this logic to the cycling DA experiments.


HaidaoLin-NOAA (approved):

HaidaoLin-NOAA:
Thanks @guoqing-noaa for making this change quickly!

PR #806, opened 2025-05-27
[hotfix] Turn off hofx_scaling for uv287

spanNOAA:
This PR disables hofx scaling for uv287, which was causing unreasonable fit ratios for METAR observations. The change specifically affects only METAR wind observations. Further investigation is still required for other types of surface wind observations.


guoqing-noaa (approved):
Thanks for the hotfix PR, @spanNOAA 

You may add "hotfix" in the title for awareness.

Issue #809, opened 2025-05-28
[rrfs-mpas-jedi] Problems with running `ncks` to copy state variables during cold start cycles


SamuelDegelia-NOAA:
If performing DA during cold start cycles, we copy the updated state variables from `ana.nc` into a linked in `mpasin.nc` that contains additional variables needed to start the `fcst` task. The `mpasin.nc` link points to `stmp/20240527/rrfs_prep_ic_00_v2.1.1/det/init.nc` that is used to initialize the forecast. 

The copying is done using the `ncks` utility through nco. However, if the structure (like dimension names, chunking, or compression) differs, ncks might leave the original structure intact, leading to mismatches. This means that `ncks` can fail to copy the updated state variables back into `init.nc` which can lead to the subsequent forecast being initialized from the background instead of the analysis. The same problem occurs for both the deterministic jedivar task and the ensemble GETKF task. 

A solution to this is to remove the state variables from `mpasin.nc` before doing the replacement from `ana.nc`. So instead of 

```
ncks -A -H -v "${var_list}" ana.nc mpasin.nc
```

We would do: 

```
ncks -O -C -x -v ${var_list} mpasin.nc tmp.nc    # remove variables and output to tmp.nc
ncks -A -v ${var_list} ana.nc tmp.nc             # copy updated variables from ana.nc to tmp.nc
mv tmp.nc "$(readlink -f mpasin.nc)"             # move tmp.nc back into mpasin.nc, preserving the linking
```

I performed tests with this on Hera and it seems to work correctly for both jedivar and GETKF.

PR #810, opened 2025-05-28
[rrfs-mpas-jedi] Fix `ncks` utility to copy state variables when performing DA during cold start cycles

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

As discussed in #809, the `ncks` utility is used to copy updated state variables from `ana.nc` into the linked `mpasin.nc` during DA cold start cycles. However, differences in file structure (e.g., dimension names, chunking, or compression) can prevent `ncks` from properly overwriting existing variables, leading to inconsistencies. As a result, the forecast may be incorrectly initialized from the background state instead of the analysis.

This PR resolves the issue by explicitly removing the relevant state variables from `mpasin.nc` before appending them in from `ana.nc`. Corresponding updates have also been made to the GETKF scripts.

EDIT: This PR also adds an if-block to `exrrfs_getkf.sh` to disable DA for cold start cycles if `${COLDSTART_CYCS_DO_DA}` is false. This is consistent with the same option in `exrrfs_jedivar.sh`. 

## TESTS CONDUCTED: 
Tested running cold start DA cycles for jedivar and GETKF. 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [x] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

## ISSUE: 
#809


scripts/exrrfs_getkf.sh
```diff
@@ -129,7 +129,7 @@ else # move post mean to umbrella if solver
       ncks -A -v ${var_list} "data/ana/mem${mem}.nc" "data/ens/tmp${mem}.nc"
       export err=$?
       err_chk
-      mv "data/ens/tmp${mem}.nc" "$(readlink -f data/ens/mem${mem}.nc)"
+      mv "data/ens/tmp${mem}.nc" "$(readlink -f "data/ens/mem${mem}.nc")"
```
guoqing-noaa, line 132: change the inside `"`  to `'`

SamuelDegelia-NOAA:
> @SamuelDegelia-NOAA This is still a mystery to me. I tested the original form on Gaea and it worked for me. I did not test it on Hera. I hope I can double check this on Hera before a merge. Thanks!

Feel free to try my simple test directory I set up on Hera at `/scratch2/BMC/zrtrr/Samuel.Degelia/workflow_det/stmp/20240527/rrfs_jedivar_00_v2.1.1/det/jedivar_00/TESTING`. 

Running `calc_diff.sh` with the old method shows differences for the theta field between mpasin.nc and ana.nc in that directory. But if we change to the new method, then theta is correctly written into mpasin.nc. I am okay with waiting to merge if you want to double check.

scripts/exrrfs_getkf.sh
SamuelDegelia-NOAA: Okay, then I will just change it back to not using quotes inside the readlink function.

guoqing-noaa:
@SamuelDegelia-NOAA I tested your case and I got the same behavior as your described in this PR.  Thanks a lot for your identifying this issue and providing a good solution!

scripts/exrrfs_getkf.sh
guoqing-noaa: Thanks! @SamuelDegelia-NOAA 
I suggest we resolve the linter comment so that future changes on this file will not bother with the linter comment repeatedly.
We may have two ways to address the linter comment:
1. add a new variable, such as "dest", to store the final destination of the link:
```
dest=$(readlink -f "data/ens/mem${mem}.nc")
mv "data/ens/tmp${mem}.nc"  "${dest}"
```

2. add a single-use shellcheck disable before line 132 as follows:
```
      # shellcheck disable=SC2086                                                                                                                                      
      mv "data/ens/tmp${mem}.nc" "$(readlink -f data/ens/mem${mem}.nc)"
```

scripts/exrrfs_getkf.sh
```diff
@@ -125,9 +125,11 @@ else # move post mean to umbrella if solver
   if [[ "${start_type}" == "cold" ]]; then
     var_list="pressure_p,rho,qv,qc,qr,qi,qs,qg,ni,nr,ng,nc,nifa,nwfa,volg,surface_pressure,theta,u,uReconstructZonal,uReconstructMeridional"
     for mem in $(seq -w 1 030); do
-      ncks -A -H -v "${var_list}" "data/ana/mem${mem}.nc" "data/ens/mem${mem}.nc"
+      ncks -O -C -x -v ${var_list} "data/ens/mem${mem}.nc" "data/ens/tmp${mem}.nc"
```
guoqing-noaa, line 128: We use the `COLDSTART_CYCS_DO_DA` variable to control whether to do cold start DA in jedivar:
```
if [[ ${start_type} == "warm" ]] || [[ ${start_type} == "cold" && ${COLDSTART_CYCS_DO_DA} == "true" ]]; then
```

We can either add that through this PR or update it in a separate PR. Thanks!

SamuelDegelia-NOAA: Thanks for catching this. I think this PR is a good place to add it since we are modifying DA for coldstart cycles.

scripts/exrrfs_getkf.sh
SamuelDegelia-NOAA: Just fixed this using the new `dest` variable.

PR #811, opened 2025-05-28
[rrfs-mpas-jedi] Split GETKF posterior observer into separate task 

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR splits the GETKF posterior observer into a separate task (`getkf_post`) that can run separately from the GETKF solver. This allows the `fcst` task to start as soon as the GETKF solver is done and thus we no longer need to wait the extra 5+ mins for the posterior observer to finish. We can now also use the RoundRobin observation distribution for the GETKF posterior observer to speed things up. 

The new task is run by setting `run as observer only: true` similar to the GETKF prior observer. But now we use the ensemble analyses as the input data. We also use the jdiag files output from the prior observer as the input obs for this new `getkf_post` task. That ensures that the same ob counts are output in the prior and posterior jdiag files. 

## TESTS CONDUCTED: 
Ran 6 h of cycling with the default `exp/exp.ens_conus12km` config file. Confirmed that the `getkf_post` task can run in parallel with the `fcst` task and that the analysis jdiag files match the size of the prior jdiag files. 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [x] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

## ISSUE: 
None


scripts/exrrfs_getkf.sh
```diff
@@ -116,8 +116,19 @@ err_chk
...
+
+# rename ombg to oman for posterior observer jdiag files
+if [[ "${TYPE}" == "post" ]]; then
+  for jdiag in "${DATA}"/jdiag*; do
+    jdiag_an="${jdiag%.nc}_ana.nc"
+    nccopy -k 3 "${jdiag}" "${jdiag_an}"
+    ncrename -g ombg,oman "${jdiag_an}"
+    rm "${jdiag}"
+  done
+fi
```
guoqing-noaa, line 128: Line 119-128: Since we run the post observer separately now, do we have to rename ombg to oman?  I think it is acceptable to keep it as is without confusion.

SamuelDegelia-NOAA: We don't have to rename it, but I think using `oman` instead of `ombg` makes more sense for the posterior jdiag files. I would like to keep it that way unless there is a concern about the extra runtime (though since this is now a separate task, the runtime addition doesn't matter much).

guoqing-noaa: Thanks! That works for me.
I have some follow-up questions: 1. why we cannot run "ncrename" directly but need to convert the NETCDF format first?  2. Those jdiag files will be under a separate `getkf_post` directory, right? If yes, it looks like we don't need to add the `_ana` suffix?  3. We will save those post jdiag files to the com/ directory, right?

guoqing-noaa: @SamuelDegelia-NOAA Thanks for the detailed information. It is interesting that the ioda issue 364 was closed while the problem remains unresolved.

guoqing-noaa (approved):
LGTM. Thanks for addressing my comments, @SamuelDegelia-NOAA

MatthewPyle-NOAA (approved):

PR #817, opened 2025-06-11
update the checkDA/checkrrfsxml utilities under qrocoto

guoqing-noaa:
Update the checkDA/checkrrfsxml utilities under qrocoto so that one may use them to check other users' retro experiments by specifying expdir at the command line


MatthewPyle-NOAA (approved):
Looks okay to me....how much testing was done of the new functionality?

PR #775, opened 2025-05-02
Add function to JEDI to conduct direct radar DBZ analysis.

hu5970:
This PR is to add the changes for  conducting the direct DBZ analysis. 
There are two steps to turn on the direct DBZ analysis:
1) set DO_ENVAR_RADAR_REF=true in configure file
2) set usage the flag of the reflectivity to 1 in convinfo

This is tested with CONUS 12km retrospective experiment.

PR #824, opened 2025-06-17
make the dbz-DA-related analysis variables optional based on the `DO_ENVAR_RADAR_REF` setting

guoqing-noaa:
`DO_ENVAR_RADAR_REF=true` adds six more analysis variables into JEDIVAR (see PR #775).
This significantly increate the memory use and run time of the JEDIVAR task (see issue https://github.com/NOAA-EMC/RDASApp/issues/407)

This PR makes the dbz-DA-related analysis variables optional based on the `DO_ENVAR_RADAR_REF` setting.
So if we do not do dbz-DA, we don't need the 6 extra analysis variables. This way, JEDIVAR needs less memory and can run faster.

### DIFF
--- a/ush/yaml_finalize
+++ b/ush/yaml_finalize
@@ -122,6 +122,7 @@ if os.path.exists(yfile2):
         yfile2 = f'{basename}_old{knt:03}.yaml'
 os.replace(yfile, yfile2)
 #
+do_envar_radar_ref = ( os.getenv("DO_ENVAR_RADAR_REF", "FASLE").upper() == "TRUE" )
 with open(yfile2, 'r') as infile, open(yfile, 'w') as outfile:
     buffer_zone = []
     in_buffer_zone = False
@@ -213,6 +214,10 @@ with open(yfile2, 'r') as infile, open(yfile, 'w') as outfile:
             buffer_zone.append(line)
 
         if not in_buffer_zone:
+            if not do_envar_radar_ref and ", cloud_liquid_water, cloud_liquid_ice, rain_water, snow_water, graupel, equivalent_reflectivity_factor" in line:
+                line = line.replace(", cloud_liquid_water, cloud_liquid_ice, rain_water, snow_water, graupel, equivalent_reflectivity_factor", "")
+            elif not do_envar_radar_ref and "    - equivalent_reflectivity_factor" in line:
+                line = line.replace("    - equivalent_reflectivity_factor\n", "")
             outfile.write(line)
     # ~~~~
     if buffer_zone:


ush/yaml_finalize
```diff
@@ -213,6 +214,10 @@ with open(yfile2, 'r') as infile, open(yfile, 'w') as outfile:
             buffer_zone.append(line)
 
         if not in_buffer_zone:
+            if not do_envar_radar_ref and ", cloud_liquid_water, cloud_liquid_ice, rain_water, snow_water, graupel, equivalent_reflectivity_factor" in line:
```
TingLei-NOAA, line 217: @guoqing-noaa Thanks for this quick fix!   It would give the option to set up the analysis variables as we need.  However, from my perspective, it is less user/developer-friendly in terms of  readability.  I think it would be easier to understand/use if : 1) The analysis variables are listed explicitly when corresponding option is selected rather than  being derived by removing some entries of a pre-defined lengthy analysis variable list  (users have to check  out the current yaml to find out what the analysis variable would be). 2) This option of different group of analysis variables could be visible/done in either the experiment setup file or the job script rather than being buried in the yaml_finalize. In addition to the readability consideration, changes here would not work for other yaml generation procedures.

TingLei-NOAA: @guoqing-noaa  1)Better readability (and maintainability )  is "potentially" beneficial to everyone (to be honest, I don't know how long and what kind of users don't want to know what analysis variables are  and  how analysis variables could be changed) and the above suggested changes won't cause any extra burden to the users/developers who don't want to know this at that time.  2) I suggested "visable/done"  in either the setup configure file or the job script. The actual action on the yaml could be done in the yaml_final or other under hood procedure , but the choice/options (visiability) could be implemented on the "interface level" with a better readability as suggested.  
For more advanced uses (though I don't think it is so advanced), in fact, if the analysis variables had been put ,say, in the set up configure , I won't have to dig a little to find out how , "suddenly", 11 analysis variables were used which I need to match with with the mgbf setup. That will also be easier for you to find out why the workflow became significantly slower since a hash. 
I think there should be a balance between readability and "encapsulation".  Good designs could help on both. But if an comprise has to be made, I prefer to the readability.   Of course, that is my own review opinions.   Thanks for the discussion.

guoqing-noaa:
@MatthewPyle-NOAA This PR is also some kind of "hotfix" as we need it to run the real-time conus3km DA experiment (otherwise, the jeidvar will crash with the current limited nodes/cores).

@TingLei-NOAA, we can always revisit this issue to make it more general when that kind of need comes up. Thanks!

guoqing-noaa:
@MatthewPyle-NOAA Thanks a lot for the quick responses!!

PR #827, opened 2025-06-18
Allow setting extra environmental variables to override some default MPAS-Model build flags

guoqing-noaa:
This PR allows setting extra environmental variables to override some default MPAS-Model build flags.
This enables use to remove the `Makefile` workaround for WCOSS2 and streamline the compiling of the MPAS-Model using spack-stack on Ursa.

The related PR: https://github.com/ufs-community/MPAS-Model/pull/141


guoqing-noaa:
> I believe I remember hearing that the plan was to move from `RRFSx/MPAS-Model` back to `ufs-community/MPAS-Model` eventually. Does this mean that we will have to remove these lines that export the build flags in `build.rdas` at that point? Either using the `_workaround_` Makefile or doing it this way works for me. But just noting this.

Yes, we will use `ufs-community/MPAS-Model` and we will make the switch once all our "workaround" change PRs have been merged, which may take a few weeks.

guoqing-noaa:
It is expected that once we switch to `ufs-community/MPAS-Model`, we won't need any of our previous MPAS-Model workarounds. Thanks!

SamuelDegelia-NOAA (approved):

PR #831, opened 2025-06-20
Add surface cycling

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Add surface cycling, plus 2 minor fixes


workflow/rocoto_funcs/prep_ic.py
```diff
@@ -22,10 +22,12 @@ def prep_ic(xmlFile, expdir, do_ensemble=False, spinup_mode=0):
         cycledefs = 'prod'
     coldhrs = os.getenv('COLDSTART_CYCS', '03 15')
     cyc_interval = os.getenv('CYC_INTERVAL')
+    sfc_cycling_hrs = os.getenv('SFC_CYCLING_HRS', '99')
 
     # Task-specific EnVars beyond the task_common_vars
     dcTaskEnv = {
-        'COLDSTART_CYCS': f'{coldhrs}'
+        'COLDSTART_CYCS': f'{coldhrs}',
+        'SFC_CYCLING_HRS': f'{sfc_cycling_hrs}',
```
guoqing-noaa, line 30: It may be more consistent with the existing `COLDSTART_CYCS` if we use something like `SFC_UPDATE_CYCS`?

With that said, I am fine with `SFC_CYCLING_HRS`.

jobs/JRRFS_PREP_IC
MatthewPyle-NOAA, line 1: Can the "module load nco" be removed from this job?  Modules should be loaded at a higher level within with the workflow.

MatthewPyle-NOAA: @chunhuazhou This PR has been sitting for a while.  Can you make the change recommended to move where the nco module is loaded?  Thanks.

MatthewPyle-NOAA (approved):
Thanks for making the requested change...

PR #838, opened 2025-06-25
Implement solution to clean up DATA, Umbrella DATA at the end of a cycle

lgannoaa:
Implement solution to clean up DATA, Umbrella DATA at the end of a cycle as agreed by NCO in RRFS kick off meeting.
Implement task dependency improvement in Ecflow workflow.
Improve exception handling on jobs.

## TESTS CONDUCTED: 
A cycled retro test was conducted on development system.

### Machines/Platforms:

- WCOSS2
  - [X] Cactus/Dogwood

### Test cases: 

- [X] Engineering tests
    - [X] Retro


BenjaminBlake-NOAA (commented):
@lgannoaa Thank you for working on this and testing this capability.  I am curious why the KEEPDATA=NO if statements are removed from the J-jobs.  Is the data cleanup now handled exclusively by the clean tasks?

BenjaminBlake-NOAA (approved):
@lgannoaa Got it, that makes sense.  Thanks!  The changes look good to me.

ShunLiu-NOAA:
The changes looks good to me. Could you also document the changes of dependencies at 00, 06, 12, 18 cycle? 
@MatthewPyle-NOAA and @lgannoaa   Please don't merge this PR till I completed the validation of https://github.com/NOAA-EMC/rrfs-workflow/pull/787 with the additional fix directories and parm/config changes so that I don't need to revalidate the files and directories.

lgannoaa:
> > > @lgannoaa Thank you for working on this and testing this capability. I am curious why the KEEPDATA=NO if statements are removed from the J-jobs. Is the data cleanup now handled exclusively by the clean tasks?
> > 
> > 
> > You are correct. The J-jobs will no longer doing individual DATA directory delete. It will be handled by clean job running at the end of each cycle. This job will use KEEPDATA to identify if DATA should be renamed or removed. We presented this solution to the NCO on the second around of the RRFS kick off meeting. They have no disagreement.
> 
> @lgannoaa This is a big change. I think it will work great in operation. Thinking of RRFS.v2, the file size is huge and one may run a number of cycles at the same time in retro experiments. If we don't clean up temporary run directories upon the completion of tasks (but waiting for the completion of a cycle), we may get "out of disk space" in the development RDHPCS.

I need to review the RRFS v2 design to come up with solution for operation. Ideally, the developer need to identify the DATA and Umbrella DATA VS share /not share condition in clear documentation to design cleanup job effectively.

lgannoaa:
> The changes looks good to me. Could you also document the changes of dependencies at 00, 06, 12, 18 cycle? @MatthewPyle-NOAA and @lgannoaa Please don't merge this PR till I completed the validation of #787 with the additional fix directories and parm/config changes so that I don't need to revalidate the files and directories.

1. The definition change in 00, 06, 12, and 18 is for startup job to ensure prep cycle run after ICS jobs are completed. This issue was discovered doing testing on development machine.
2. This PR has no FIX directory changing components. No need to wait for the FIX PR to proceed. @MatthewPyle-NOAA I recommend to proceed with merge now.
3. NCO just announced a production switch is scheduled on July 1st. We need to focus on production switch event after get this change merged.

MatthewPyle-NOAA (approved):

PR #851, opened 2025-07-12
add a trap in build.all, allowing Ctrl+C to stop the whole build process

guoqing-noaa:
Currently `build.all` uses background processes to build multiple components at the same time. 

Users find it is difficult to kill all related background processes once `build.all` starts. This is inconvenient when someone needs to stop currently-running `build.all`.

This PR adds a trap in `build.all` so that CTRL+C can stop the whole build process


Junjun-NOAA (approved):

SamuelDegelia-NOAA (approved):
LGTM, good idea!

PR #877, opened 2025-07-28
Port rrfs-workflow to Derecho Step 1

guoqing-noaa:
Port rrfs-workflow to NCAR supercomputer Derecho.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,11 @@ Team Charter (draft): https://docs.google.com/document/d/1uLbPx-pOWp7eECz_7VHRt_
 
 ### Currently supported platforms
 -   NOAA RDHPCS Hera
+-   NOAA RDHPCS Ursa
 -   NOAA RDHPCS Jet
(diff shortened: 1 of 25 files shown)


Junjun-NOAA (approved):

MatthewPyle-NOAA:
@guoqing-noaa  One small question - should the MPI_RUN_CMD definition be removed from the exp files not under rt_jet?

guoqing-noaa:
> @guoqing-noaa One small question - should the MPI_RUN_CMD definition be removed from the exp files not under rt_jet?

Great observation! Thanks!
However, removing `MPI_RUN_CMD` from the exp files is NOT required. Users can still define that explicitly in their exp files and it will NOT break anything.

For rrfs-workflow, we hope to maintain a minimum set of core exp files (i.e. 4 under exp/ and 4 under exp/rt_jet). I had thought users better keep their own exp files in separate repos instead of crowding the rrfsflow/exp directory.

PR #819, opened 2025-06-12
update the MPAS-Model to the latest version with DAcycling changes

guoqing-noaa:
This PR updates the MPAS-Model to the latest version, which has made significant changes for the `config_do_DAcycling` option.

It includes the changes from the following PRs and issues, and removes all relevant temporary workarounds in the workflow.
https://github.com/ufs-community/MPAS-Model/pull/136
https://github.com/ufs-community/MPAS-Model/pull/137
https://github.com/ufs-community/MPAS-Model/issues/93

Note: the above PRs have not been merged. But a few GSL developers have been waiting for this function and would like to use this latest version ASAP to do a few important retro experiments. 

So after the discussion, we decided to merge all the above PRs into a fork and use that fork temporarily for our retro tests.
It is estimated that the authoritative `ufs-community/MPAS-Model` will be available in a month or so. At that time, we will switch to use the `ufs-community` repo (instead of the `NOAA-GSL` repo)

Note: relevant RDASApp PR: https://github.com/NOAA-EMC/RDASApp/pull/402

PR #888, opened 2025-08-03
supplemental changes to match recent updates on the use of MPAS model streams 

guoqing-noaa:
Previously, we used the same `input` stream to read both `init.nc` and `mpasout.nc` in the cycling process. At that time, we introduced an intermediate `mpasin.nc` file to simplify the logic (i.e the `prep_ic` task will link either `init.nc` or `mpasout.nc` to a unified file `mpasin.nc`.)

With recent MPAS model updates, after the merge of PR #819, we are clear that we will initialize forecasts from three different files under different scenarios:

- cold start forecasts use `init.nc` (input stream)
- restart forecasts use `restart.nc` (restart stream)
- mpasout warm start forecasts use `mpasout.nc` (da_state stream)

So at this moment, `mpasin.nc` becomes an obsolete thing. And some users have reported confusion on this.
This PR addresses this and phases out the no-longer-needed `mpasin.nc`.

`workflow/ush/mpasout_to_restart.py` has been updated accordingly.


guoqing-noaa:
Tested 6 cycles, worked as expected.

Issue #899, opened 2025-08-11
[rrfs-mpas-jedi] `UMBRELLA_PREP_IC_DATA` can be removed before `getkf_post` task finishes


SamuelDegelia-NOAA:
When `KEEPDATA` is false, the `UMBRELLA_PREP_IC_DATA` directory is deleted at the end of the forecast task (`JRRFS_FCST`). However, for ensemble DA, `getkf_post` still needs this directory which contains the ensemble analyses. If the forecasts finish quickly, or if `getkf_post` spends a long time in the queue, then the forecast task may delete `UMBRELLA_PREP_IC_DATA` before `getkf_post` starts, causing the latter to fail.

An easy solution would just be to only remove `UMBRELLA_PREP_IC_DATA` in `JRRFS_FCST` for deterministic runs. For the ensemble runs, we can instead remove `UMBRELLA_PREP_IC_DATA` in `JRRFS_GETKF`.

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for catching this. 
Yes, at this time, your proposed solution looks good.

I have been working on expanding the `clean` task for retros (currently it mostly serves the real-time experiments).
The idea is to clean UMBRELLA_DATA at the end of each cycle. Since rrfs-workflow is highly configurable to run different combination of tasks, we need to figure out what the last task is in a cycle. I hope to submit this PR this week. Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa Glad to hear that! 

Along the lines of what you describe, would it be possible to have some kind of option to set the number of cycles to wait before  cleaning? By default, we clean the current cycle at the end of that cycle. But we had a discussion today at EMC that it might be helpful to always keep the last cycle around for debugging. In that scenario, it we could wait to clean up a cycle until the subsequent cycle is complete. Just a thought. Thanks!

guoqing-noaa:
> [@guoqing-noaa](https://github.com/guoqing-noaa) Glad to hear that!
> 
> Along the lines of what you describe, would it be possible to have some kind of option to set the number of cycles to wait before cleaning? By default, we clean the current cycle at the end of that cycle. But we had a discussion today at EMC that it might be helpful to always keep the last cycle around for debugging. In that scenario, it we could wait to clean up a cycle until the subsequent cycle is complete. Just a thought. Thanks!

@SamuelDegelia-NOAA Thanks for the discussion. The current `do_clean` task already has this capability.
[...]

PR #900, opened 2025-08-11
Hotfix: remove cascading cleaning up of UMBRELLA_DATA

guoqing-noaa:
PR #875 introduced cascading cleaning up of `UMBRELLA_DATA`.
This idea does not work in some situations, for example, the GETKF situation reported in issue #899 or using multiple groups of lbc tasks.
This PR is to revert changes made in PR #875 

For now, to meet data cleaning needs, one can set the following variables in exp files to only keep the latest once cycle of stmp data:
```
export DO_CLEAN=true
export STMP_CLEAN_HRS=1
export COM_CLEAN_HRS=840
export LOG_CLEAN_HRS=840
```

I have been working on expanding the `clean` task to purge UMBRELLA_DATA at the end of each cycle and will commit that in a separate PR later this week.

### DIFF
--- a/jobs/JRRFS_FCST
+++ b/jobs/JRRFS_FCST
@@ -65,13 +65,8 @@ fi
 #----------------------------------------
 # Remove the Temporary working directory
 #----------------------------------------
-if [[ "${KEEPDATA}" == "NO" ]]; then
-  # rm -rf "${DATA}"
-# we cannot remove fcst DATA here since save_fcst lags behind and
-# still needs to access them. Instead, let save_fcst remove fcst DATA
-  rm -rf "${UMBRELLA_PREP_LBC_DATA}"
-  rm -rf "${UMBRELLA_PREP_IC_DATA}"
-fi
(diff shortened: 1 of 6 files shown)


guoqing-noaa:
> @guoqing-noaa Thanks for making this change - this works for me! I also have a working version of the solution I described in #899 where `UMBRELLA_PREP_IC_DATA` is removed by the `getkf_post` task for ensemble DA. Will your solution to expand the `clean` task resolve the need for that fix? If so, I will refrain from making a PR.

Thanks for the discussion. Yes, expanding the `clean` task will resolve issue #899.  
Actually, UMBRELLA_DATA is by design to be removed upon the completion of each cycle.
(cascading removal was only a temporary solution I made in a hurry. I happened to do a few conus3km experiments at the same time and found they took so many disk spaces. I should have used the `clean` task).

SamuelDegelia-NOAA:
Sounds good to me, thanks @guoqing-noaa.

SamuelDegelia-NOAA (approved):

PR #921, opened 2025-09-02
revamp `yaml_finalize` based on `hifiyaml` to make it more powerful and general

guoqing-noaa:
With the futher developments of rrfs-workflow (v2), we have more and more advanced needs to edit the super YAML files in terms of blocks.

For example,  for @Junjun-NOAA and Masanori's upcoming GSIBEC PR, users can choose what BEC models to use and rrfs-workflow is expected to do the switch easily.

We may use [`PyYAML`](https://github.com/yaml/pyyaml) to handle this, but unfortunately it changes the original formatting. Comments are lost, anchors/aliases are rendered differently, other YAML elements (such as fold, pipe, override, list, tag, etc) also changes structurely. This makes it is almost impossible to compare or visually inspect the changes it makes and hence prevents it from being used in the YAML editing stage.

This PR utilizes the [`hifiyaml`](https://pypi.org/project/hifiyaml/) package to revamp `yaml_finalize` so that it is more powerful and general to handle more complicated situations. [`hifiyaml`](https://pypi.org/project/hifiyaml/) is a `High-fidelity YAML parser that preserves formatting`. It keeps comments, anchors, aliases, etc and hence is great tool for editing rrfs super YAML files.

Beside the revamp of `yaml_finalize`,  a few related necessary changes were also made:

1. `getkf_solver.yaml` and `getkf_observer.yaml` are combined into one file `getkf.yaml` to reduce duplicate work and streamline the workflow.   
`yaml_finalize` will convert an observer-flavored getkf.yaml to a solver-flavored getkf.yaml automatically when it is needed.

3. For GETKF, change the ambiguous `TYPE` to more descriptive `GETKF_TYPE`

Tests were done for all the current use cases and the generated final YAML file were found to be identical to those before this PR. I have been running a retro test and will report results when it is done.

### DIFF
--- a/ush/yaml_finalize
+++ b/ush/yaml_finalize
@@ -51,14 +51,14 @@ if not do_radar_ref and ytype == "jedivar":
     hy.modify(data, "_member/state variables", varlist)
     pos, _ = hy.get_start_pos(data, "cost function/background/state variables")
     next_one = hy.next_pos(data, pos)
-    del data[next_one-3:next_one]  # delete the extra 3 lines: equivalent_reflectivity_factor, w, upward_air_velocity
+    del data[next_one - 3:next_one]  # delete the extra 3 lines: equivalent_reflectivity_factor, w, upward_air_velocity
 
 elif do_radar_ref and ytype == "getkf":
     varlist = "increment variables: [air_temperature, water_vapor_mixing_ratio_wrt_moist_air, eastward_wind, northward_wind, air_pressure_at_surface]"
     hy.modify(data, "increment variables", varlist)
     pos, _ = hy.get_start_pos(data, "_member/state variables")
     next_one = hy.next_pos(data, pos)
-    del data[next_one-3:next_one]  # delete the extra 3 lines: equivalent_reflectivity_factor, w, upward_air_velocity
+    del data[next_one - 3:next_one]  # delete the extra 3 lines: equivalent_reflectivity_factor, w, upward_air_velocity
 
 # tweaks for cold start DA
 if ytype == "jedivar" and start_type == "cold":
--- a/ush/yamltools4rrfs.py
+++ b/ush/yamltools4rrfs.py
@@ -145,8 +145,8 @@ def getkf_observer_tweak(data, getkf_type):
     if getkf_type == "post":
         i = 0
         while i < len(data) - 1:
-            if (data[i].strip().startswith("action:") and data[i+1].strip().startswith("name: reduce obs space")):
-                del data[i:i+2]  # delete and move to the next line
+            if (data[i].strip().startswith("action:") and data[i + 1].strip().startswith("name: reduce obs space")):
+                del data[i:i + 2]  # delete and move to the next line
             else:
                 i += 1  # if no deletion, move to the next line
 


ush/yaml_finalize
```diff
@@ -1,233 +1,147 @@
...
+  read HX from disk: true
+  save posterior ensemble: true
+  save prior mean: true
+  save posterior mean: true
+  do posterior observer: false
+'''
+if ytype == "getkf" and (GETKF_TYPE == "solver" or GETKF_TYPE == "post"):
+    need_getkf_solver = True
+    block = hy.text_to_yblock(solver_driver_str)
+    hy.modify(data, "driver", block)
```
SamuelDegelia-NOAA, line 86: I don't think we want these driver options for `GETKF_TYPE == "post"`. The posterior observer basically runs like the prior observer task but with replacing the background ensemble with the analysis ensemble. So it should use the default driver options in the now renamed `getkf.yaml`.

parm/getkf.yaml
```diff
@@ -31,12 +31,9 @@ _member: &memberConfig
   - rain_water
   - snow_water
   - cldfrac
-  - equivalent_reflectivity_factor
-  - w
-  - upward_air_velocity
   stream name: background
 
-increment variables: [air_temperature, water_vapor_mixing_ratio_wrt_moist_air, eastward_wind, northward_wind, air_pressure_at_surface, cloud_liquid_water, cloud_liquid_ice, rain_water, snow_water, graupel, equivalent_reflectivity_factor, upward_air_velocity]
```
SamuelDegelia-NOAA, line 39: I feel like it is better to have all state variables and increment variables in the super yamls, and then use the new yaml tools to remove those variables if they are not needed. That way users can easily see which variables are needed just by inspecting file in parm instead of digging through the detailed yaml tool scripts.

SamuelDegelia-NOAA (commented):
Had a couple of comments, but thanks for organizing this effort @guoqing-noaa!

ush/yaml_finalize
guoqing-noaa: @SamuelDegelia-NOAA Thanks for your comments on this part.
I was trying to repeat the current workflow behavior and did not check what is really needed by the `post` GETKF.
In this case, what other changes will be made for the `post` GETKF?  And we can combine your `yaml_getkf_post` into `yaml_finalize`? Thanks!

guoqing-noaa: @SamuelDegelia-NOAA Thanks for the detailed information.
For `post`, do we use `RoundRobin` or `Halo`?

SamuelDegelia-NOAA: Good catch. We should use `RoundRobin` for the `post` task. It could technically use `Halo` but would probably be quite a bit slower.

guoqing-noaa: Great. I will keep `RoundRobin` for post. Thanks!

guoqing-noaa:
@SamuelDegelia-NOAA Thanks a lot for your review and discussions! I have updated this PR accordingly.
You may take a peek of what `yaml_finalize` does for getkf as follows:
```
export USE_CONV_SAT_INFO=false
export GETKF_TYPE=post
module load python

cd rrfs-workflow/ush
cp ../parm/getkf.yaml .
./yaml_finalize getkf.yaml debug > tmp.yaml
vimdiff getkf.yaml tmp.yaml
```

guoqing-noaa:
@SamuelDegelia-NOAA Since all temporal thinning filters are currently commented out in `getkf.yaml`, you may uncomment one or more temporal thinning filters to confirm they are removed by `yaml_finalize`. Thanks!

HaidaoLin-NOAA (approved):

MatthewPyle-NOAA (approved):

SamuelDegelia-NOAA:
Thanks @guoqing-noaa for making the updates - looks good to me!

PR #963, opened 2025-09-25
update RDASApp hash, solve memory leak issue in gsibec workaround

Junjun-NOAA:
This PR is to align with the latest RDASApp submodule updates, synchronizing the RDASApp hash in rrfs-workflow.

Updates to the GSIBEC workaround files are also included, mainly motivated to resolve the memory leak issue. Among the workaround files, we figured out that three files can be removed.


guoqing-noaa (approved):
LGTM. Thanks for updating these, @Junjun-NOAA!

ShunLiu-NOAA (approved):

Issue #75, opened 2023-09-26
[main] Add functionality to post processing to add AWIPS headers to GRIB output


MatthewPyle-NOAA:
A functionality that is needed for operations is adding what are known as "headers" onto the GRIB files send to AWIPS/SBN.  Should be able to follow an example from an existing system, such as HRRR or NAM, that uses the $TOCGRIB2 utility for the purpose (such as /lfs/h1/ops/prod/packages/nam.v4.2.4/util/ush/mkawpgrb.sh)

AndrewBenjamin-NOAA:
I already have parm files, and a basic script modeled on the HRRR that uses wgrib2 to parse out the proper wmo products and tocgrib2 to place headers onto the deterministic grib2 files. When the time comes I can provide all of this to incorporate it into the deterministic RRFS workflow.

MatthewPyle-NOAA:
From Andrew:

/lfs/h2/emc/vpppg/noscrub/andrew.benjamin/rrfs_vpppg/parm
faa headers are in the faa dir, ifi icing in the ifi icing dir
also included the template and make files that generate the headers for all fhr

@MatthewPyle-NOAA to get them all into a parm/wmo space in workflow.  Keep template/script in a util/script/ space like is done for HREF.  Item to ask NCO about??

MatthewPyle-NOAA:
Also need to wait on a GRIB2 library update that can handle > 2 GB files (breaks toc2grib for the full NA files)

MatthewPyle-NOAA:
Have added the parm files into a feature branch.  May want to create a PR for that even though they won't be used for anything yet (still need to add a mkawp.sh kind of functionality).

PR #991, opened 2025-10-07
Use grib_util/1.2.4 to generate the AWIPS products for RRFS

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- In this PR, the AWIPS products with WMO headers on the native RRFS rotated lat-lon grid over North America are added to the RRFS workflow.  There is a bewildering issue with tocgrib2 in grib_util/1.5.0 where tocgrib2 crashes on the wind speed records for various forecast hours, and it cannot currently be used with the RRFS grid.  The current working solution is to use grib_util/1.2.4 for generating the AWIPS products and to run tocgrib2 twice - once for all the records (not including wind speed), and a second time for the wind speed records - then cat the two files together at the end.  Running tocgrib2 twice is needed in order to get around the file size limitation of 2 GB with grib_util/1.2.4.

## TESTS CONDUCTED: 

These changes were successfully tested on Cactus for the 10/7 00Z cycle.  GRIB2 output files are available here out to 84 hours:
`/lfs/h3/emc/lam/noscrub/ecflow/ptmp/emc.lam/ecflow_rrfs/para/com/rrfs/v1.0/rrfs.20251007/00/wmo`

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [x] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Fixes issue #75 

## CONTRIBUTORS (optional): 
Thanks to @AndrewBenjamin-NOAA for helping figure out that tocgrib2 with grib_util/1.5.0 is not working correctly.
Thanks to @lgannoaa for assisting with testing the prdgen changes on Cactus.


BenjaminBlake-NOAA:
@MatthewPyle-NOAA I reverted back to grib_util/1.2.4 for everything so there is no need to load multiple versions.  Would it be preferable to use two different versions of grib_util, 1.2.4 for the wind speed records and 1.5.0 for everything else?  I know NCO wants us to move towards newer versions of libraries but 1.5.0 is not working properly (that's correct, only the wind speed records are problematic).

PR #1042, opened 2025-10-31
use NOAA-GSL/WPS which allows smoke/dust in Vtable

guoqing-noaa:
This PR sets the WPS repository to `NOAA-GSL/WPS` which allows smoke/dust in Vtable.
This is the first step to include chemistry related changes into rrfs-workflow.

Tested and confirmed that this PR and the current rrfs-workflow generated identical ungrib output (i.e. WPS intermediate binary files).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "sorc/WPS"]
 	path = sorc/WPS
-	url = https://github.com/wrf-model/WPS
-        branch = master
+	url = https://github.com/NOAA-GSL/WPS
+        branch = rrfs
         ignore = all
 [submodule "sorc/MPAS-Model"]
 	path = sorc/MPAS-Model


haiqinli (approved):

MatthewPyle-NOAA:
@guoqing-noaa Is there a reason why these changes can't be made with the current WPS repository?  Is the goal just to  have more control/flexibility?

guoqing-noaa:
> @guoqing-noaa Is there a reason why these changes can't be made with the current WPS repository? Is the goal just to have more control/flexibility?

Good question!  I will pass this to the chemistry developers to urge them seek ways to commit back.
But my estimation is that it will take at least 6 months~1 year or even longer to get this resolved. As you mentioned, using NOAA-GSL/WPS will give us more control/flexibility to move forward quickly. Thanks!

PR #945, opened 2025-09-15
Modification to ensmean per comments by Matthew Pyle

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Modification to ensmean per comments by Matthew Pyle, using a script to combine all the ncea jobs

## TESTS CONDUCTED: 
Tested on Jet.

PR #926, opened 2025-09-03
Add ensmean calculation and post-processing

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Add ensemble mean forecast calculation based on ncea, and mpassit and upp for the ensemble mean forecast

## TESTS CONDUCTED: 
Tested for conus-12km on Jet in retro mode

PR #1045, opened 2025-11-01
[rrfs-mpas-jedi] add an MPI tool `rank_run` for running serial jobs in parallel

guoqing-noaa:
Sometimes, we need to run serial jobs in parallel to simplify task configuration and speed up turnaround time.
PR https://github.com/NOAA-EMC/rrfs-workflow/pull/926 discussed a NCO tool `cfp`, but this tool is not available for public use.
PR #945 @chunhuazhou introduced `srun --multi-prog` to run multiple serials jobs in parallel.
https://github.com/NOAA-EMC/rrfs-workflow/blob/03eca2ab2c1ef9db3d8578ef47a3380ba5d7f528/scripts/exrrfs_ensmean.sh#L90

But `--multi-prog` only applies for SLURM and does not work for other job scheduler (such as PBS, etc).

This PR adds a universal light-weight MPI tool `rank_run`, which applies to all job schedulers, for running serial jobs in parallel. This way, we can use the same ex-scripts across different platforms ( such as WCOSS2, Derecho (using PBSPro), or RDHPCS (using SLURM). It also automatically handle the mismatch between the number of commands and the number of MPI ranks, making the workflow scripts simpler and more robust.

`rank_run` supports two ways to run serial jobs:     
(1) get a file name from the command line, read all lines from that given file, and then run each line in a different MPI rank;
(2) get a script name pattern from the command line, replace the wildcard `*` with the rank numbers and then each rank run its corresponding script if exists.
eg:
```
${MPI_CMD} rank_run.x cmdfile
${MPI_CMD} rank_run.x 'postgrib/wgrib_*.sh'  # have to bound the name pattern with single quotes
```

For more information, check this wiki: https://github.com/RRFSx/rank_run/wiki

This PR only adds the `rank_run` source code and associated build script. We will apply it to rrfs-workflow ex-scripts in separate PRs.

### DIFF
--- a/doc/build_and_run.md
+++ b/doc/build_and_run.md
@@ -9,7 +9,7 @@ Note:
 1. The first command is to make sure `git-lfs` is loaded as it is required for cloning RDASApp
 2. `GIT_LFS_SKIP_SMUDGE=1` is to skip downloading git-lfs binary data used by JEDI ctests, which is NOT needed by rrfs-workflow.  
    This will avoid intermittent RDASApp checkout failures when JCSDA repositories exceed their LFS budget.
-3. If you run cold start forecasts only and don't need data assimilation, you can `vi build.all` and comment out this line `./build.rdas &> ./log.build.rdas 2>&1 &` before running `./build.all`
+3. If you run cold start forecasts only and don't need data assimilation, run `./build.all noda`.
 
 # 2. Setup and run experiments:
 ### 2.1. copy and modify exp.setup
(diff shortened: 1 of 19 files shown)


guoqing-noaa:
I may not make it clear in my last post. I would provide a little further information.

`rank_run` can be used in any tasks (such as ioda converter, UPP, forecast, chem preprocessing, GDAS to WPS, etc). 

Most times, those tasks use different modulefiles from `rrfs_utils`. If we put `rank_run` in rrfs_utils, we will need to unload the default modules in a task and load rrfs_utils modules instead and then restore the default modules in that task. This will look strange and is also very inconvenient.

So the natural choice is to let spack-stack provide compiled `rank_run`. Every component in rrfs-workflow is based on spack-stack although different components may use different spack-stack versions due to various technical/practical reasons.

For now, each rrfs-workflow task loads the `rrfs/<machine>` module by default, so it is a reasonable workaround for now to build `rank_run` using the default simple rrfs module.

@MatthewPyle-NOAA @ShunLiu-NOAA Let me know if you have any further questions. Thanks!

PR #1039, opened 2025-10-30
[main] Updates for the real-time RRFS system

MatthewPyle-NOAA:
## DESCRIPTION OF CHANGES: 

- Changes bad symbolic linking syntax to a cpreq to use the proper workflow.conf files.
- Corrects some problematic resource request in the spinup save_restart ecf script.
- Switches parallel copies to only do 10 at a time
- Increases memory requests in some ecf scripts
- Fixes .idx file creation issue for new NBMFLD output.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [x] RRFS_A:  Based on 20251030 (starting 12Z) testing of new real-time parallel hash.
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## CONTRIBUTORS (optional): 

@lgannoaa provided the cpreq suggestion for the link script, the ecf memory changes, the parallel copy change.  @BenjaminBlake-NOAA provided the UPP script fix.

PR #1051, opened 2025-11-03
[rrfs-nco] fix link target args in scripts/setup_ecf_links.sh

JamesPolly-NCO:
## DESCRIPTION OF CHANGES: 

- fixed scripts/setup_ecf_links.sh to no longer create broken links to workflow.conf_{dev,prod} files.
  - relative paths used in link commands (`ln TARGET LINK_NAME`) are counterintuitive and the target argument is relative to the location where the link name will be created. 
  - existing script has targets defined as relative to working directory, but targets need to be changed to be in the same directory as the link name (i.e., no subdirectory).
- @lgannoaa Should confirm.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [x] Others:
  - Ops environment; jobs won't run without being able to locate workflow.conf

### DIFF
--- a/ecf/setup_ecf_links.sh
+++ b/ecf/setup_ecf_links.sh
@@ -46,15 +46,15 @@ cd ${ECF_DIR}/../fix/workflow/
 echo "point at proper workflow.conf version..."
   rm -f ./det/workflow.conf ./enkf/workflow.conf ./ensf/workflow.conf ./firewx/workflow.conf
 if [ ${resource_config} == "NCO" ]; then
-  ln -s ./det/workflow.conf_prod ./det/workflow.conf
-  ln -s ./enkf/workflow.conf_prod ./enkf/workflow.conf
-  ln -s ./ensf/workflow.conf_prod ./ensf/workflow.conf
-  ln -s ./firewx/workflow.conf_prod ./firewx/workflow.conf
+  ln -s ./workflow.conf_prod ./det/workflow.conf
+  ln -s ./workflow.conf_prod ./enkf/workflow.conf
+  ln -s ./workflow.conf_prod ./ensf/workflow.conf
+  ln -s ./workflow.conf_prod ./firewx/workflow.conf
 else
-  ln -s ./det/workflow.conf_dev ./det/workflow.conf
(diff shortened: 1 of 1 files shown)


MatthewPyle-NOAA:
@JamesPolly-NCO Yes, we've changed that logic to do a copy rather than the bad linking logic.  Is included in an open PR for the "main" branch.  https://github.com/NOAA-EMC/rrfs-workflow/pull/1039/files#diff-4a815add20497ca4a1963efab908db00e081ad8991e88d8f1c2797bf3f16b3a2

MatthewPyle-NOAA:
Also, I think what you have here is using the same configuration file for all of the WGFs, which wouldn't be correct I believe.

JamesPolly-NCO:
> Also, I think what you have here is using the same configuration file for all of the WGFs, which wouldn't be correct I believe.

As I have it here the link will point at the workflow.conf file that is in the respective (det, enkf, ensf, etc.) directory. 

When the link is accessed, it will point to the workflow.conf file sitting in the same directory as the link.

This is why I say it's kinda counter intuitive. Looking at the script and the commands, the TARGET_NAME in the four commands looks like it would be the same "file" (which doesn't exist in active directory) but the TARGET_NAME arg simply tells the link where to look, which is relative to the location where the link resides.

Using cpreq will require loading prod_utils, which means that when this package is being installed, we'll need to load prod_utils.

Is this script being used in production runs or is it just to be used on package installation/setup? If not being used in production maybe no need for cpreq?

JamesPolly-NCO:
Thanks for the input! Closing.

PR #1058, opened 2025-11-05
Hotfix - turn off radiosonde DA for now

guoqing-noaa:
We have not extensively tested Radiosonde data assimilation in RRFS.v2.
The UV220 observations crashed the JEDIVAR task and we are working to debug this.

This PR is to turn off radiosonde DA by default so other users can run the workflow out-of-the-box successfully.


Junjun-NOAA (approved):

MatthewPyle-NOAA:
Strange that such a basic ob type would cause trouble, but will get it merged.

PR #1054, opened 2025-11-04
add/modify fix file and parameters to use GSIBEC with MPASv8.3.1 and L65

Junjun-NOAA:
This PR is a follow up for [PR1012](https://github.com/NOAA-EMC/rrfs-workflow/pull/1012). 
After MPAS model is updated to v8.3.1 and the number of vertical levels are increased to L65, a fix file and some parameters are updated to run 3DVAR with GSIBEC. 
Also, GSIBEC parameters for conus3km are updated to those generated using MPAS grid directly.
3DVAR GSIBEC has been tested for conus12km and conus3km and both runs as expected. 
Updating bumploc files for 3DEnVar application will be included in a separated PR.

PR #1012, opened 2025-10-17
[rrfs-mpas-jedi] MPASv8.3.1 update

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Updating MPAS-Model to version 8.3.1, including the new fix files with L65 vertical levels

PR #1059, opened 2025-11-05
add bumploc file for conus12km and conus3km with L65

Junjun-NOAA:
Follow up [PR1054](https://github.com/NOAA-EMC/rrfs-workflow/pull/1054) and [PR1012](https://github.com/NOAA-EMC/rrfs-workflow/pull/1012) , this PR add bumploc file for conus12km and conus3km with L65.

### DIFF
--- a/fix/bumploc/conus12km_L60_120_401km11levels
+++ b/fix/bumploc/conus12km_L60_120_401km11levels
@@ -1,1 +1,0 @@
-../.agent/bumploc/conus12km.20250326/conus12km_L60_120_401km11levels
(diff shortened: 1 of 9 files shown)


guoqing-noaa:
> Is a 65 level, south 3.5 km BUMPLOC file coming? The 60 level versions for that domain are removed by this PR with nothing to replace them.

@MatthewPyle-NOAA Great question! At this moment and in the foreseen future, no one will run south3.5km 3DEnVar. So we  don't update it to avoid crowding the FIX_RRFS2 disk space. It will be added only when needed. Thanks!

PR #1060, opened 2025-11-06
[dev-sci] Add 6km , 9k and 12km North American domain parameters for grid creation 

TingLei-NOAA:
When "dual resolution" hybrid , in which , the inner loop grids (and in-memory ensembles grids) are coarser than the background, are considered in the fv3jedi variational runs,  the fv3 grids on coarser resolution of the North American domains are needed. 
This PR is to add 6km, 9km and 12 km parameters for those coarser resolution North American domains. 
Other changes to use those coarser NA domain in the cycling of the workflow like time steps and parameterization schemes for the forecast model are pending.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 #*       @defunkt
-*       @MatthewPyle-NOAA @ShunLiu-NOAA @JacobCarley-NOAA @chan-hoo @BenjaminBlake-NOAA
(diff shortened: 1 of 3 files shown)


ShunLiu-NOAA:
@TingLei-daprediction Do you mean you want to get the PR merged first and then further test the performance of dual-resolution function? Is there any fix files updated for this PR?

TingLei-NOAA:
@ShunLiu-NOAA  Yes. This PR would allow other developers to use the dual resolution functions, otherwise, they need get those changes from my side.

PR #1067, opened 2025-11-10
remove the outdated 'FCST_LENGTH' variable

guoqing-noaa:
In the very early stage, we had used one `FCST_LENGTH` value to specify the forecast length for every cycle.
In practice, RRFS will have different forecast lengths for different cycles, so we have moved away from using `FCST_LENGTH`. Instead, we have used a new variable `FCST_LEN_HRS_CYCLES` to explicitly define forecast lengths all 24 cycles.
But we did not remove the `FCST_LENGTH` variable in time. This caused confusions to lots of users and some developers wrongly used this outdated variable for developments.

This PR is to remove the outdated `FCST_LENGTH` variable and clean up the codes.

### DIFF
--- a/scripts/exrrfs_prep_chem.sh
+++ b/scripts/exrrfs_prep_chem.sh
@@ -25,9 +25,8 @@ cd "${DATA}" || exit 1
 #
 # find forecst length for this cycle
 #
-fcst_length=${FCST_LENGTH:-1}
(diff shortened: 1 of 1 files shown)


guoqing-noaa:
Tested retros on Ursa and it worked as expected.

PR #1086, opened 2025-11-13
[rrfs-mpas-jedi] Update the offline domain checks for better efficiency 

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR follows https://github.com/NOAA-EMC/RDASApp/pull/467 by introducing new versions of the offline domain check (`offline_domain_check.py` and `offline_domain_check_satrad.py`) to rrfs-workflow. When testing these codes on larger domains, it was found that the original versions became extremely slow and could not finish within a reasonable amount of time (e.g., upwards of an hour for a 3-km NA domain). This is not feasible when running on multiple observation types in the workflow. 

This update replaces the old alpha-shape triangulation (which scaled poorly with large MPAS or FV3 grids) with a fast method that generates the true model boundary directly. By avoiding Delaunay triangulation over millions of points, the runtime drops from minutes or hour to a few seconds while producing an equally accurate domain mask.

This new version of the domain check is significantly faster. Running it on most IODA files takes 2-3 seconds. For the larger files such as `ioda_msonet.nc` or radiance observations, it takes ~10 s. The new version also yields very similar results to the original version. 

Here are some comparisons for `ioda_msonet.nc` when running the domain check on the conus3km MPAS mesh: 

```
Full obs count: 521,505
Domain check, original version: 495,238
Domain check, new version: 495,239

Runtime, original version: 142 s
Runtime, new version: 12 s 
```

## TESTS CONDUCTED: 
Tested by running a retro for three cycles to make sure the `ioda_bufr` task completes successfully and with expected results. 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [x] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
None


SamuelDegelia-NOAA:
I addressed the linter comments. Also, I looked at the entire runtime of the `ioda_bufr` task at 00 UTC on the conus3km mesh. Prior to the update this task took ~32 min while after these changes it only takes ~4 min.

PR #1044, opened 2025-11-01
[rrfs-mpas-jedi] Fix download time out in offline domain check in ioda_bufr

huiying-noaa:
This PR is in response to a previous issue (https://github.com/NOAA-EMC/rrfs-workflow/issues/987) following comments from @SamuelDegelia-NOAA. 

## TESTS CONDUCTED: 
Machines/Platforms: Gaea C6
Test cases: 
A. 2024050700 - 2024050800 hourly for 12km GSIBEC 3dvar
Results: ioda_bufr run time reduced from ~40 min (2024050700-2024050705) to ~3 min (2024050706-2024050800). 

B. 2024050700 - 2024050706 hourly for 3km GSIBEC 3dvar (intended for the same 25 cycles, but fcst crashed at 2024050704)
Results: fixed time out crashes of ioda_bufr with run time ~23 min (2024050700-2024050706).

## ISSUE: 
- Fixes the issue(s) mentioned in #987 . 

## Participants
@SamuelDegelia-NOAA @guoqing-noaa

PR #1091, opened 2025-11-14
[dev-sci] Add `err_chk` for python codes in `ioda_bufr` task

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

It was found that the `ioda_bufr` task can complete successfully even if the offline python codes (domain check, vad thinning) fail. Here, we add error checking using `err_chk` to exit the task if these python codes fail.

Two bugs were also fixed that were causing these python codes to crash (but weren't triggering the entire job to fail due to the previous lack of error checking): 
* The "-f" flag is removed from the offline domain check. The same change was made for `[rrfs-mpas-jedi]` in [#1044](https://github.com/NOAA-EMC/rrfs-workflow/pull/1044), but here we make the same change for `[dev-sci]`. Thanks to @xyzemc for catching this problem.   
* The `satwnd` obs now run with the regular offline domain check instead of the satrad version. 

## TESTS CONDUCTED: 

This was tested by running three cycles of `ioda_bufr` for the simplified workflow on WCOSS2. Everything completes as expected. 

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
None

## CONTRIBUTORS (optional): 
@xyzemc

### DIFF
--- a/ush/set_predef_grid_params.sh
+++ b/ush/set_predef_grid_params.sh
@@ -896,6 +896,152 @@ case ${PREDEF_GRID_NAME} in
   fi
   ;;
 #
+# The RRFS North America domain with ~6km cells.
+#
+#-----------------------------------------------------------------------
+#
+"RRFS_NA_6km")
+
+  GRID_GEN_METHOD="ESGgrid"
+  ESGgrid_LON_CTR="-112.5"
+  ESGgrid_LAT_CTR="55.0"
+  ESGgrid_DELX="6000.0"
+  ESGgrid_DELY="6000.0"
+  ESGgrid_NX=1975
+  ESGgrid_NY=1350
+  ESGgrid_PAZI="0.0"
+  ESGgrid_WIDE_HALO_WIDTH="6"
+  DT_ATMOS="${DT_ATMOS:-60}"
+  LAYOUT_X="${LAYOUT_X:-40}"
+  LAYOUT_Y="${LAYOUT_Y:-45}"
+  BLOCKSIZE="${BLOCKSIZE:-28}"
+
+  if [ "$QUILTING" = "TRUE" ]; then
+    WRTCMP_write_groups="${WRTCMP_write_groups:-1}"
+    WRTCMP_write_tasks_per_group="${WRTCMP_write_tasks_per_group:-50}"
+    WRTCMP_output_grid="${WRTCMP_output_grid:-rotated_latlon}"
+    if [ "${WRTCMP_output_grid}" = "rotated_latlon" ]; then
+      WRTCMP_cen_lon="-113.0"
+      WRTCMP_cen_lat="55.0"
+      WRTCMP_lon_lwr_left="-61.0"
+      WRTCMP_lat_lwr_left="-37.0"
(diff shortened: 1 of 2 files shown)


scripts/exrrfs_ioda_bufr.sh
```diff
@@ -259,25 +259,33 @@ cp "${RDASAPP_DIR}"/rrfs-test/IODA/offline_vad_thinning.py .
 # offline domain check & patch
 for ioda_file in ioda*nc; do
   grid_file="${FIX_GSI}/${PREDEF_GRID_NAME}/fv3_grid_spec"
-  if [[ "${ioda_file}" == *abi* ]]; then
+  if [[ "${ioda_file}" == *abi* && "${ioda_file}" != *satwnd* ]]; then
```
delippi: Thanks for this! I actually found this today as well and was going to update after my current PR (currently a draft) gets in!

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -259,25 +259,33 @@ cp "${RDASAPP_DIR}"/rrfs-test/IODA/offline_vad_thinning.py .
...
-    ./offline_domain_check_satrad.py -o "${ioda_file}" -g "${grid_file}" -f -s 0.005 >> $pgmout
+    export pgm="offline_domain_check_satrad.py"
+    ./offline_domain_check_satrad.py -o "${ioda_file}" -g "${grid_file}" -s 0.005 >> $pgmout
+    export err=$?; err_chk
     base_name=$(basename "$ioda_file" .nc)
     mv  "${base_name}_dc.nc" "${base_name}.nc"
   elif [[ "${ioda_file}" == *atms* || "${ioda_file}" == *cris* ]]; then
     echo " ${ioda_file} ioda file detected: temporarily skipping offline domain check"
   else
+    export pgm="offline_domain_check.py"
```
delippi: Do we want all the pgm exports? Do they get used by the error checking?

SamuelDegelia-NOAA: The `err_chk`  utility outputs the error status based on the `pgm` variable. For example: 

```
...
Time to find obs within domain: (0.15s), 00:00:00
Time to create new obs file: (0.08s), 00:00:00
+ 63 + export err=0
+ 63 + err=0
+ 63 + err_chk
offline_domain_check.py completed cleanly
```

Leaving out the exports would not break anything, but it would cause all of these new `err_chk` messages to output `bufr2ioda.x completedly cleanly` since that is what `pgm` is currently set to in the script.

PR #1093, opened 2025-11-16
remove the hardcoded use of the `intel` compilers

guoqing-noaa:
I got feedback that rrfs-workflow hardcoded the use of `intel` compilers and does not have the flexibility to use GNU compilers. Actually rrfs-workflow is capable of handling different type of compilers, `intel`, `gnu` or others, but we just have not paid attention to this yet.

This PR removes the hardcoded use of the `intel` compilers and enables the use of other types of compilers (such as `gnu`).


guoqing-noaa:
Tested in a few retros and worked as expected.

PR #1107, opened 2025-11-19
[rrfs-mpas-jedi] fix/workaround for incorrect logic of the vertical level index searching

spanNOAA:
## Description

This PR introduces a temporary workaround (ahead of the full UFO fix being merged) for an index-searching bug in the vertical interpolation routine. The issue affects how the interval index `ix` is determined from a monotonic array `gh` (either increasing or decreasing), and can lead to incorrect indexing or out-of-bounds access during interpolation.

A corresponding permanent fix has been submitted to UFO: https://github.com/JCSDA-internal/ufo/pull/3899.

## Issue(s) addressed

In the original implementation, the default value of `ix` was set to `nlevs - 1`. However, the subsequent interpolation expression unconditionally accesses: `gh[ix]` and `gh[ix+1]`. Thus `ix` must never exceed `nlevs - 2`.
When the input `d` exceeds the bounds of `gh`, the search loop may fail to assign a new `ix`, leaving `ix = nlevs - 1`, which leads to out-of-bounds memory access during `gh[ix+1]`.

Additionally, the original search loop started from `k = 0` even though the edge case `d <= gh[0]` or `d >= gh[0]` was already handled by the outer conditional, making the logic slightly inconsistent.

This PR updates the index-search logic as follows:
The default value of `ix` is changed from nlevs - 1 to nlevs - 2, ensuring safe access to gh[ix+1] under all conditions.
The search loop now iterates from `k = 1` to `k < nlevs`, producing correct interval selection for all interior points.
Both increasing (iflag == 1) and decreasing (iflag == -1) branches are updated symmetrically.
The resulting `ix` is always guaranteed to satisfy:
```c
0 ≤ ix ≤ nlevs - 2
```

### DIFF
--- a/sorc/build.rdas
+++ b/sorc/build.rdas
@@ -43,6 +43,7 @@ EXEC4="${HOMErrfs}/sorc/RDASApp/build/bin/bufr2netcdf.x"
 cd "${HOMErrfs}/sorc/RDASApp" || exit 1
 
 # fix for the incorrect logic of the vertical level index interpolation in pressurecheck function
+# PR: https://github.com/JCSDA-internal/ufo/pull/3899
 cp ../_workaround_/RDASApp/obsfunctions/ObsErrorFactorPressureCheck.cc sorc/ufo/src/ufo/filters/obsfunctions/
 
 # workaround for using the GSL surface operator


guoqing-noaa:
@SamuelDegelia-NOAA rrfs-workflow has more workarounds, not limited to RDASApp and even has more JEDI workarounds than RDASApp. I would avoid crowding RDASApp with rrfs-workflow workarounds.

Also, there are community users who use our workflow but using jedi-bundle, mpasjedi-bundle directly. 
So I would prefer keep rrfs-workflow workarounds in one place. Thanks!

PR #1134, opened 2025-11-28
remove `modulefiles/prod_util` and use the one provided in the current environment

guoqing-noaa:
In the very early stage, we copied `prod_util` from a given spack-stack version to accommodate the situation that some subcomponents do not provide a slurm-ready version. Now that all subcomponents have adopted spack-stack-1.9.x and we can remove `modulefiles/prod_util` from rrfs-workflow and load the default one in each subcomponents' environment.

### DIFF
--- a/fix/bumploc/na12km_L65_400_401km11levels
+++ b/fix/bumploc/na12km_L65_400_401km11levels
@@ -1,0 +1,1 @@
+../.agent/bumploc/na12km.20251125/na12km_L65_400_401km11levels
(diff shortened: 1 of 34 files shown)


guoqing-noaa:
@MatthewPyle-NOAA FYI, the merged commit for this PR does not have a version number.
I think it should be `[v2.1.2-8.1]`.  Thanks!

daviddowellNOAA:
Thanks @guoqing-noaa!  The rrfs-workflow scripts are now working on Jet following this change to the spack-stack installation.

PR #1145, opened 2025-12-03
[dev-sci] Fixes for running EnKF retros with our simplified configuration (no spinup, no DA at cold start)

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

This PR updates the EnKF workflow to support simplified RRFSv1.2 retros that run without spinup and without DA during cold start cycles. A follow-up PR will introduce JEDI-based EnKF support.

The changes here include: 

**1. Prevent empty Rocoto metatasks**

Conditional logic is added to `parm/FV3LAM_wflow.xml` to avoid generating empty `run_ensemble` metatasks.
Previously, setting `DO_SPINUP="FALSE"` and/or `DO_DACYCLE="FALSE"` caused empty tasks to be written to the XML, which led `run_rocoto.sh` to fail.

**2. Skip ensemble mean calculation on cold start cycles**

An option is added to `scripts/exrrfs_calc_ensmean.sh` to bypass ensemble-mean computation on cold start cycles. This is similar to what is done for GSI and jedivar scripts and prevents failures when restart files do not yet exist. 

**3. Prevent unrealistic EnKF increments**

The ppn was reduced for the `run_enkfupdt` task to produce more realistic increments. See https://github.com/NOAA-EMC/GSI/pull/961 for more details.

## TESTS CONDUCTED: 
Tested by running two cycles of the simplified EnKF retro on WCOSS2. 

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
None


SamuelDegelia-NOAA:
Actually, converting this to draft temporarily. I think there is one more fix to include here but I am testing it now.

MatthewPyle-NOAA:
@SamuelDegelia-NOAA Still testing an additional fix to include in this one?

SamuelDegelia-NOAA:
Okay I think this PR is ready for re-review now. The unphysical increments in the EnKF were resolved by reducing ppn per https://github.com/NOAA-EMC/GSI/pull/961. 

I also found that the `calc_ensmean` and `run_enkfupdt` tasks were sometimes failing with a `Cannot place all ranks on node list` error. I found an older issue that referenced solving this by switching from `place=excl` to `place=vscatter`. That seemed to have solved the problem here since I was seeing crashes every couple of cycles before the change, and I have now run for ~50 cycles after the change without seeing a crash. But if anyone knows a better way to solve that, please let me know.

PR #1151, opened 2025-12-04
[dev-sci] [bugfix] Copy ioda_empty_satwnd.nc prior to bufr2ioda step in case of no satwnd data

delippi:
## DESCRIPTION OF CHANGES: 
This PR fixes a failure that occurs when a cycle contains zero SATWND observations. In these cases, bufr2ioda produces no files (`ioda_satwnd.abi_goes-16.nc` or `ioda_satwnd.abi_goes-18.nc`). The JEDI YAML is configured with:
```
missing file action: "warn"
```
which should allow a missing obs file to be handled quietly. Instead, the workflow still crashes.

### Fix
Before running bufr2ioda, the workflow now places an empty IODA file (with the correct groups, dimensions, and variable schema) in the target location. If bufr2ioda detects real observations, it overwrites that file. If it finds nothing, the preplaced empty file remains, and JEDI reads a valid but zero-obs dataset.

### Why this approach
- We guarantee the existence of a properly structured obs file for every SATWND source.
- JEDI receives a valid obs space even when no observations exist.
- The workflow no longer depends on inconsistent behavior of “missing file action: warn.”
- Diagnostics downstream can produce explicit zero-obs jdiags instead of failing or silently skipping.

This brings SATWND handling in line with the strategy used for other observation types and produces deterministic behavior whether the cycle contains data or not.

### Misc
- Changed default walltime for archive retro task.
- Updated the RDASApp hash.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

### DIFF
--- a/sorc/Externals.cfg
+++ b/sorc/Externals.cfg
@@ -68,7 +68,7 @@ protocol = git
 repo_url = https://github.com/NOAA-EMC/RDASApp
 # Specify either a branch name or a hash but not both.
 # branch = develop
-hash = 6e37ee1
+hash = f3b7ea9
 local_path = RDASApp
 required = True
 


MatthewPyle-NOAA (approved):
Looks good to me, but wanted to confirm the reduction from a 5:30 to 0:10 limit for the archive job.  Is pretty dramatic, but maybe by default you aren't archiving much?

PR #1178, opened 2025-12-08
descope `save_fcst` to `save_for_next` whose job is to copy the cycling mpasout for the next cycle(s)

guoqing-noaa:
Currently `save_fcst` tries to move all forecast netcdf files from the fcst unmbrella directory to the save_fcst umbrella directory and copy cycling mpasout files to com/. This scope is too big and not needed.  

In fact, we only need to copy one (or two or three) mpasout file (f1h for hourly cycling, f3h for 3-hourly cycling, etc) to com/ so that the next cycle(s) can start without waiting the long forecast to complete.  

The current `save_fcst` way also causes confusions to some users on what forecasts to be saved. It also makes `rocotoreboot save_fcst` impossible since the first few hours files have already been moved and the rebooted `save_fcst` just hangs there until reaching the wall time limit.

This PR descopes the `save_fcst` task and replaces it with the `save_for_next` task to streamline the overall workflow.

Tested on multiple HPCs for 25 cycles and worked as expected.


MatthewPyle-NOAA (approved):
I've been thinking if there might be a better name than mpasout_cycling to describe what it is doing.  `mpas_for_cycling`?  `prep_nextcycle`?  Something else?

guoqing-noaa:
> I've been thinking if there might be a better name than mpasout_cycling to describe what it is doing. `mpas_for_cycling`? `prep_nextcycle`? Something else?

Thanks a lot for the discussion and good ideas!

I have been debating on an appropriate name for a while, :). So far, IMHO, `mpasout_cycling` looks like the best option available so far. I have tried `mpasout_for_cycling` but I think it is too long.  Then, I found `for` is not a must and we can still convey the same meaning without it.

`mpas_for_cycling` looks like too general and a little ambiguous? Folks may think the major point of rrfs-workflow is to do mpas cycling, why we still need such a task? IMHO, it is not as accurate as `mpasout_for_cycling`.
Also MPAS generates `mpasout.nc`, `history.nc`, `diag.nc` (and a potential `chem_diag.nc`) and we only need to save `mpasout`, not other two (or three).

`prep_nexcycle` is not accurate as well since we don't prep the next cycle here. We only save them to the com/ directory of current cycle so that next cycle may fetch those files and run `prep_ic`. In cold start only runs, we don't need the `mpasout_cycling` task at all, but `prep_nextcycle` may give folks an impression that they need this step.  

I have tried `save_mpasout`, but we only save one mpasout file (not all of them);  or `save_mpasoutf1h`, but 3/6-hourly cycling, we save f3h or f6h fcst instead of f1h

We can continue thinking about this. Once we get a better one, we can change to that at that time. Thanks!

guoqing-noaa:
> Thanks for indulging me. How about `save_for_nextcyc`? Seems hard to be sufficiently descriptive without being very long!

No worries. Thanks for the good discussions!

Between `save_for_cycling` and `save_for_nextcyc`, I may slightly prefer the former.  I think although we only save one mpasout at the moment, we will need a fallback capability sometime where we save f1h and f2h (or even f3h?) so that the next cycle after the next may do DA based on the 2h fcst. Also, this workflow will be used by quite a few people to run coldstart only experiments, where they don't have "cycling" but they do have next cycle. So for them, "cycling" may be a little bit more accurate than `nextcyc` (although the differences might be tiny).

guoqing-noaa:
@MatthewPyle-NOAA Finished 6 retro cycles and everything worked as expected after the latest task name change. Thanks!

PR #1192, opened 2025-12-10
[rrfs-mpas-jedi] A few fixes

chunhuazhou:
## DESCRIPTION OF CHANGES: 
   A few fixes:
     - Saving a copy of log.atmosphere.0000.out to com/ so we can check back after data scrubbed under stmp/
     - use "ln" instead of "cp" for prep_lbc to save run time
     - fix GEFS interpolation after the change from "NAME_PATTERN" to "FILENAME_PATTERN"
     - remove ensemble data dependency when not doing DA at cold start cycles

Tests done on Ursa.


scripts/exrrfs_prep_lbc.sh
```diff
@@ -42,7 +42,7 @@ if [ -r "${checkfile}" ]; then
   for fhr in  ${fhr_all}; do
     CDATElbc=$(${NDATE} "${fhr}" "${CDATE}")
     string_time=$(date -d "${CDATElbc:0:8} ${CDATElbc:8:2}" +%Y-%m-%d_%H.%M.%S)
-    ${cpreq} "${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}/lbc/${WGF}${MEMDIR}/lbc.${string_time}.nc"  "${UMBRELLA_PREP_LBC_DATA}/."
+    ln -sf "${COMINrrfs}/${RUN}.${YYYYMMDDlbc}/${HHlbc}/lbc/${WGF}${MEMDIR}/lbc.${string_time}.nc"  "${UMBRELLA_PREP_LBC_DATA}/."
```
guoqing-noaa: @chunhuazhou Do not change line 45.

Two solutions:
1.  add an extra line before line 5:
`cpreq="ln -snf"`

The original idea is that in the development stage, we use links; but in operation, we use `cpreq` by defining it in the job cards.

2. More elegant solution is to add `<envar><name>cpreq</name><value>ln -snf</value></envar>` in the prep_lbc task in rrfs.xml. I am fine with either one at the moment.

Thanks!

workflow/rocoto_funcs/jedivar.py
```diff
@@ -112,11 +114,37 @@ def jedivar(xmlFile, expdir, do_spinup=False):
...
+    {iodadep}{ens_dep}
+    </and>
+    {streqs}
+    </or>
+        '''
+    else:
+        da_dep = f'''
+        {iodadep}{ens_dep}
+        '''
+    #
```
guoqing-noaa: @chunhuazhou Could you elaborate the issue you met for the changes here? Thanks!

chunhuazhou: When we have COLDSTART_CYCS_DO_DA=false, the dependency on the ensemble forecast files at cold start cycles are not needed and would slow down the whole cycling. The changes here address this in different situations:
- if COLDSTART_CYCS_DO_DA=false, disable ensemble data dependency for the jedivar task at cold start cycles; keep the ensemble data dependency for the other cycles
-  if COLDSTART_CYCS_DO_DA=true, keep the ensemble data dependency for all the cycles

guoqing-noaa: Looks great. Thank you!

guoqing-noaa (approved):

guoqing-noaa (approved):
LGTM. Thanks, @chunhuazhou

MatthewPyle-NOAA (approved):

PR #1195, opened 2025-12-10
each mesh now has its own dedicated subdirectory under FIXrrfs

guoqing-noaa:
We will have more and more meshes/domains for different projects and different testing purposes. 

So far, to my knowledge, we have `conus3km, conus12km, na12km, south3.5km, ea5km, ar3p5km, global-15-3km, eu12km, na3km`.
Chemistry applications will have a few more small domains coming as well.

To facilitate adding new domains with a minimum effort, this PR reorganizes mesh related fix files so that each mesh now has its own dedicated subdirectory under `${FIXrrfs}`. This way, a new domain can be easily added by linking to under `${FIXrrfs}` to provide needed fix files/directories.

In rrfs-workflow (v2), by default, we only provide `conus3km,conus12km,na12km` (a future `na3km` may be included).  Other meshes fix files are expected to be maintained by their stakeholders outside rrfs-workflow.

### DIFF
--- a/fix/chemistry/stream_list/stream_list.atmosphere.output
+++ b/fix/chemistry/stream_list/stream_list.atmosphere.output
@@ -1,5 +1,4 @@
 aod550
-hwp
 vis
 PM2_5
 PM10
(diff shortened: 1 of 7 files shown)


guoqing-noaa:
12 retro cycle completed on Ursa and worked as expected.

Junjun-NOAA (approved):

PR #1207, opened 2025-12-12
[rrfs-mpas-jedi] Add nonvar cloud analysis functionality to ensemble experiments

ShawnMurdzek-NOAA:
## DESCRIPTION OF CHANGES: 
This PR enables users to run the nonvar cloud analysis on each ensemble member separately when rrfs-workflow is run in ensemble configuration. This is done by setting `DO_NONVAR_CLOUD_ANA=true` in the `exp` file used to generate the workflow. By default, the nonvar cloud analysis is turned off for ensemble runs.

## TESTS CONDUCTED: 
Successfully generated workflow XML files for the following experiment configurations on Gaea C6 and successfully ran each experiment for multiple cycles:
- 3-member conus12km ensemble with nonvar cloud analysis
- 3-member conus12km ensemble without nonvar cloud analysis
- Deterministic conus12km run with nonvar cloud analysis

For the experiments with the nonvar cloud analysis, I verified that the nonvar cloud analysis ran successfully for at least one cycle. Output from the two ensemble experiments diverged after the first cycle, which indicates that the nonvar cloud analysis was changing the model output (as expected).

As a sidenote, I noticed that the `upp_ensmean_g01` task failed for the majority of the cycles in my ensemble runs. This is true for both the runs with and without the nonvar cloud analysis, so I do not think this is a result of my changes. Is this a known issue?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ workflow/rrfsexps/
 workflow/config.chemistry
 sorc/build_timing.txt
 superyamls
+fix/ar3.5km
+fix/ea5km
+fix/global-15-3km
+fix/eu12km
--- a/sorc/_workaround_/RDASApp/mpas-jedi/mpas_fields_mod.F90
+++ b/sorc/_workaround_/RDASApp/mpas-jedi/mpas_fields_mod.F90
@@ -1256,7 +1256,6 @@ subroutine push_back_other_fields(self, key, other)
   call self%push_back(key, other%subFields, key)
 end subroutine push_back_other_fields
 
-
 subroutine to_fieldset(self, geom, vars, afieldset, include_halo, flip_vert_lev)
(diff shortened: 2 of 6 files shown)


guoqing-noaa:
@ShawnMurdzek-NOAA For the upp crash, could you paste your log file path here?

@EricJames-NOAA and I can take a loot at it. Thanks!

guoqing-noaa:
@ShawnMurdzek-NOAA Thanks for reporting this.
Yes, it was because the clean task, whose dependency is the completion of the `upp` metatasks.
In your situation, 'upp_ensmean_g01' the last task of that cycle, so we need to change the dependency to it.  This is a bug we will fix soon. Thanks!

guoqing-noaa (approved):

PR #1234, opened 2025-12-30
[rrfs-mpas-jedi] Implement the JCB YAML generation - step 1

guoqing-noaa:
This PR is the first step to implement the JCB YAML generation for rrfs-mpas-jedi.

We utilize the JCB concise form so that both the current YAML_GEN method 1 and the JCB method share the same header  and the observer YAML files as much as possible.

The JCB method and the current YAML_GEN method 1 also share the same logic to use `convinfo/satinfo` to control what observations to be assimilated at the rrfs-workflow runtime (this step can be skipped if `export USE_CONV_SAT_INFO=false` in the exp files).

Tested on Ursa and worked as expected.

The future steps will be running one week of retros to make sure the current YAML_GEN method 1 and the JCB method generate the same verification statistics (although this is expected through current visual YAML file examination, a retro is still needed to confirm this). Also, effort will be made to minimize the differences between two methods while adhering to the current rrfs-workflow code norm.


delippi (commented):
This is not how JCB should be implemented. The majority of things should be in RDASApp which is 90% there already, you're just missing mpas related things. The only JCB related items in rrfs-workflow should be the JCB configuration file and actually running JCB in the ex scripts.

- All of the obs space yamls to be used are in RDASApp here: https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/observations/atmosphere
- Non obs space yaml parts are here: https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/model/atmosphere
- jcb-algorithms and jcb-rdas are here: https://github.com/NOAA-EMC/RDASApp/tree/develop/parm

You can make use of the obs space yamls without any changes. The model parts will require some changes specific to mpas.

You can follow how we run JCB in the exrrfs_analysis_jedi.sh of [dev-sci] branch starting here: https://github.com/NOAA-EMC/rrfs-workflow/blob/716f85b52a854137bc6a02422c07ae6a38a977e6/scripts/exrrfs_analysis_jedi.sh#L329

The jcb configuration file that we use in [dev-sci] is here: https://github.com/NOAA-EMC/rrfs-workflow/blob/dev-sci/parm/rdas-atmosphere-templates-fv3_c13.yaml

Please let me know if you have any questions.

SamuelDegelia-NOAA:
Thanks for putting this together. I agree that this approach works for enabling JCB in rrfs-mpas-jedi, but I do have some concerns about consistency with how we are already using JCB in `[dev-sci]` and RDASApp.

From a configuration standpoint, I have found it much easier to work with JCB when the YAML is split into component templates, similar to what we do in RDASApp [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/model/atmosphere). That structure makes it straightforward to swap components in and out at the top-level JCB configuration depending on the experiment. For example, we currently maintain separate background-error templates for BUMP and for MGBF, and switching between them is just a matter of referencing a different template in the main JCB YAML, as done [here](https://github.com/NOAA-EMC/RDASApp/blob/543b306184f5c638e7a683f9610ff243a618b90b/parm/jcb-rdas/test/ci/jcb-rrfs_fv3jedi_2024052700_3denvar.yaml#L19).

My bigger concern, though, is the addition of new copies of all observation-space YAMLs under `parm/jcb-rrfs/obs`. We already maintain these observation YAMLs in RDASApp [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/observations/atmosphere). Those files have been updated many times over the last few months, and keeping separate copies in `[rrfs-mpas-jedi]` risks the two branches drifting apart over time. 

Since we are actively developing and tuning these obs-space configurations, it would be really beneficial to have a unified set of YAMLs shared between `[dev-sci]` and `[rrfs-mpas-jedi]`, rather than duplicating them in each branch. I think aligning the JCB structure and reusing the existing RDASApp obs-space YAMLs would help reduce divergence and make future maintenance much easier across all rrfs-workflow branches.

delippi:
@hu5970 I think that will be beneficial. I would be happy to walk through how we use JCB.

guoqing-noaa:
> Thanks for putting this together. I agree that this approach works for enabling JCB in rrfs-mpas-jedi, but I do have some concerns about consistency with how we are already using JCB in `[dev-sci]` and RDASApp.
> 
> From a configuration standpoint, I have found it much easier to work with JCB when the YAML is split into component templates, similar to what we do in RDASApp [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/model/atmosphere). That structure makes it straightforward to swap components in and out at the top-level JCB configuration depending on the experiment. For example, we currently maintain separate background-error templates for BUMP and for MGBF, and switching between them is just a matter of referencing a different template in the main JCB YAML, as done [here](https://github.com/NOAA-EMC/RDASApp/blob/543b306184f5c638e7a683f9610ff243a618b90b/parm/jcb-rdas/test/ci/jcb-rrfs_fv3jedi_2024052700_3denvar.yaml#L19).
> 
> My bigger concern, though, is the addition of new copies of all observation-space YAMLs under `parm/jcb-rrfs/obs`. We already maintain these observation YAMLs in RDASApp [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/observations/atmosphere). Those files have been updated many times over the last few months, and keeping separate copies in `[rrfs-mpas-jedi]` risks the two branches drifting apart over time.
> 
> Since we are actively developing and tuning these obs-space configurations, it would be really beneficial to have a unified set of YAMLs shared between `[dev-sci]` and `[rrfs-mpas-jedi]`, rather than duplicating them in each branch. I think aligning the JCB structure and reusing the existing RDASApp obs-space YAMLs would help reduce divergence and make future maintenance much easier across all rrfs-workflow branches.

@SamuelDegelia-NOAA Thanks for the discussions! I agree with your point to get things started (as "step 1" means in the title).

I think you should be very familiar with JCB now. All JCB does is to combine different pieces into a (nested) Python  dictionary and then use PyYAML to write out a format-not-perserving super YAML file. For the header part, we have lots of methods to make it flexible, using a jinjia2 template is one way; modifying the JCB-rendered dictionary directly is another way; using a different header is a third way. It is not a difficult task to use different Headers.

The observer part is at least 95% percent the same between jcb-rrfs/obs and jcb-rdas/observations. There are some differences at the moment and I listed them here:
https://github.com/RRFSx/superyamls/wiki/obsSpace-YAML-differences-between-rrfs%E2%80%90workflow-(v2)-and-jcb%E2%80%90rdas

We will work to minimize the differences. But this will be an evolving process, not to be resolved completely in one PR. And some small non-scientific differences will remain for a while. Thanks!

HaidaoLin-NOAA (approved):

PR #1242, opened 2026-01-05
RRFS Prep Cycle Hourly Recovery Capability

lgannoaa:
Modify the RRFS package and ecflow workflow to implement an hourly recovery capability for the prep cycle job during the initial three-hour window.
Proposed Technical Changes
  Increased RESTART file set:
    Update the DET forecast to save RESTART file sets at FHR 001, 002, and 003 for all cycles.
    Update the ENKF forecast to save RESTART file sets at FHR 001, 002, and 003 for all even-hour cycles.
  ECFLOW Workflow:
    Modify the ecflow workflow to support additional jobs to save new RESTART set.
    Update job cards to accommodate increased resource demands, specifically targeting higher memory allocation and extended wallclock time.
  Job Dependencies:
    Adjust the prep cycle job to ensure it generates and provides a sufficient volume of INPUT files required to forecast jobs.
  Implement necessary updates to the fix files to maintain consistency with the new output cadence.

### DIFF
--- a/ecf/defs/rrfs_prod.def
+++ b/ecf/defs/rrfs_prod.def
@@ -57,7 +57,99 @@ suite rrfs_dev
                 event 35 release_enkf_prep_cyc_mem030
                 event 36 release_det_analysis_gsi
                 event 37 release_save_restart_long_1
+                event 500 release_save_restart_long_2
+                event 501 release_save_restart_long_3
                 event 38 release_save_restart_long_6
+                event 600 release_enkf_save_restart_001_f1
+                event 601 release_enkf_save_restart_001_f2
+                event 602 release_enkf_save_restart_001_f3
+                event 603 release_enkf_save_restart_002_f1
+                event 604 release_enkf_save_restart_002_f2
+                event 605 release_enkf_save_restart_002_f3
+                event 606 release_enkf_save_restart_003_f1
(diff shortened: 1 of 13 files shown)


WeiWei-NCO:
It looks like `exrrfs_blend_ics.sh` also checks for restart files. Do we need to update this job too?

lgannoaa:
> It looks like `exrrfs_blend_ics.sh` also checks for restart files. Do we need to update this job too?
Good question!
It is my understanding that ENKF blending job on 07/19 is designed to run in pair with the ensinit job. If the RESTART files from previous cycle is available the blending will run. Otherwise, ensinit process will need to be run. This part of the system design is not related to the scope of two/three hourly restart. Therefore, it should not be changed.

MatthewPyle-NOAA (approved):

ShunLiu-NOAA (approved):

JamesPolly-NCO (approved):

WeiWei-NCO:
@MatthewPyle-NOAA do we need this PR into rrfs-nco as well?

MatthewPyle-NOAA:
@WeiWei-NCO Good catch - hadn't realized this wasn't made for the rrfs-nco branch.  @lgannoaa can you make it for rrfs-nco as well?

PR #1279, opened 2026-01-21
[rrfs-mpas-jedi] Prep ic para

chunhuazhou:
## DESCRIPTION OF CHANGES: 
The ensemble prep_ic is taking too much time for the real-time runs on Ursa (could be as long as ~10 min per member in some cases), and even causing dead prep_ic jobs after finishing some members. This PR adds the capability to run prep_ic parallelly for ensemble members and significantly reduce the run time to less than 5 min for all the members. This has been tested in retro mode and is now in the real-time run.

### DIFF
--- a/scripts/exrrfs_prep_ic.sh
+++ b/scripts/exrrfs_prep_ic.sh
@@ -177,7 +177,7 @@ done
 #
 echo "===== CMDFILE ====="
 cat  "$CMDFILE"
-cat  ${CMDFILE} | xargs -I {} -P "${SLURM_NTASKS}" sh -c '{}'
+xargs -I {} -P "${SLURM_NTASKS}" sh -c '{}' < "${CMDFILE}"
 
 # Check for errors
 export err=$?
--- a/scripts/exrrfs_prep_ic.sh
+++ b/scripts/exrrfs_prep_ic.sh
@@ -31,6 +31,13 @@ if (( spinup_mode == -1 )); then
 fi
 echo "this cycle is ${start_type} start"
 
+timestr=$(date -d "${CDATE:0:8} ${CDATE:8:2}" +%Y-%m-%d_%H.%M.%S)
+
+export CMDFILE="${DATA}/poescript_prep_ic"
+export CMDFILE2="${DATA}/poescript_ncks"
+mkdir -p "$(dirname "$CMDFILE")"
+: > "$CMDFILE"
+: > "$CMDFILE2"
 
 # Populate the list for the ensemble members, or deterministic member
 if [[ "${ENS_SIZE:-0}" -gt 2 ]]; then
@@ -39,27 +46,20 @@ else
   mem_list=("/") # if determinitic
 fi
 
+#
+#  find the right background file
+#
 for memdir in "${mem_list[@]}"; do
   # Determine path
   if [[ ${#memdir} -gt 1 ]]; then
     umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
     mkdir -p "${COMOUT}/prep_ic/${WGF}${memdir}"
-    pid=$((10#${memdir: -2}-1))
-    export CMDFILE="${DATA}/script_prep_ic_${pid}.sh"
   else
     umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}"
-    export CMDFILE="${DATA}/script_prep_ic_0.sh"
   fi
-
-  mkdir -p "$(dirname "$CMDFILE")"
-  : > "$CMDFILE"
 
   # Create directory safely
   mkdir -p "${umbrella_prep_ic_data}"
-#
-#  find the right background file
-#
-timestr=$(date -d "${CDATE:0:8} ${CDATE:8:2}" +%Y-%m-%d_%H.%M.%S)
 
 if [[ "${start_type}" == "cold" ]]; then
   thisfile=${COMINrrfs}/${RUN}.${PDY}/${cyc}/ic/${WGF}${memdir}/init.nc
@@ -109,6 +109,7 @@ fi
 #
 # do sfc cycling
 #
+
 for hr in ${SFC_UPDATE_CYCS:-"99"}; do
   shr=$(printf '%02d' $((10#$hr)) )
   var_list="smois,snow,snowh,snowc,sst,canwat,tslb,skintemp,landmask,isltyp,ivgtyp,soilt1"
@@ -125,21 +126,50 @@ for hr in ${SFC_UPDATE_CYCS:-"99"}; do
     done
     if [[ -r ${thisfile} ]]; then
       targetfile=${umbrella_prep_ic_data}/mpas_sfc.nc
-      echo "${cpreq}" "${thisfile}" "${targetfile}"  >> "$CMDFILE"
+      echo "${cpreq}" "${thisfile}" "${targetfile}" >> "$CMDFILE"
       if [[ -r "${umbrella_prep_ic_data}/init.nc" ]]; then
         to_file="${umbrella_prep_ic_data}/init.nc"
       elif [[ -r "${umbrella_prep_ic_data}/mpasout.nc" ]]; then
         to_file="${umbrella_prep_ic_data}/mpasout.nc"
       fi
       echo "surface update from ${thisfile} to ${to_file}"
-      echo ncks -O -C -x -v ${var_list} "${to_file}"  tmp.nc  >> "$CMDFILE"
-      echo ncks -A -v ${var_list} "${targetfile}" tmp.nc  >> "$CMDFILE"
-      echo mv tmp.nc "${to_file}"  >> "$CMDFILE"
+      echo " ncks -O -C -x -v ${var_list} \"${to_file}\"  tmp.nc ; \
+             ncks -A -v ${var_list} \"${targetfile}\" tmp.nc ; \
+             mv tmp.nc \"${to_file}\" " >>  "$CMDFILE2"
     else
       echo "SFC_UPDATE failed, cannot find warm start file: ${thisfile}"
     fi
   fi
 done
+
+done
+#
+#
+# use xargs to run the command using  one core at a time (this takes a long time!)
+#
+echo "===== CMDFILE ====="
+cat  "$CMDFILE"
+xargs -I {} -P "${NTASKS}" sh -c '{}' < "$CMDFILE"
+export err=$?
+
+# Check for errors
+export err=$?
+if (( err != 0 )) ; then
+    echo "prep_ic failed with error code ${err} "
+    err_exit
+fi
+
+# ncks commands
+echo "===== CMDFILE2 ====="
+cat  "$CMDFILE2"
+xargs -I {} -P "${NTASKS}" sh -c '{}' < "${CMDFILE2}"
+
+# Check for errors
+export err2=$?
+if (( err2 != 0 )) ; then
+    echo "sfc_update ncks failed with error code ${err2}"
+    err_exit
+fi
 
 #
 #  find the right satbias file
@@ -166,24 +196,10 @@ if [[ "${PREP_IC_TYPE}" == "jedivar" ]] || [[ "${PREP_IC_TYPE}" == "getkf"  ]]; 
     satbias_path=${COMINrrfs}/${RUN}.${PDYii}/${cycii}/${PREP_IC_TYPE}${spinup_str}/${WGF}
     nSatbias=$(find "${satbias_path}"/*satbias*.nc | wc -l)
     if (( nSatbias > 0 )); then
-      echo cp "${satbias_path}"/*satbias*.nc  "${umbrella_prep_ic_data}"  >> "$CMDFILE"
+      cp "${satbias_path}"/*satbias*.nc  "${umbrella_prep_ic_data}"
       echo "found satbias from ${satbias_path}"
       break
     fi
   done
-fi
-
-done
-
-${cpreq} "${EXECrrfs}"/rank_run.x .
-${MPI_RUN_CMD} ./rank_run.x "${DATA}/script_prep_ic_*.sh"
-
-# Check for errors
-export err=$?
-if (( err != 0 )); then
-    echo "prep_ic failed with error code ${err}"
-    err_exit
-else
-    echo "prep_ic completed successfully"
 fi
 
--- a/scripts/exrrfs_prep_ic.sh
+++ b/scripts/exrrfs_prep_ic.sh
@@ -30,15 +30,39 @@ if (( spinup_mode == -1 )); then
   done
 fi
 echo "this cycle is ${start_type} start"
+
+export CMDFILE="${DATA}/poescript_prep_ic"
+mkdir -p "$(dirname "$CMDFILE")"
+: > "$CMDFILE"
+
+# Populate the list for the ensemble members, or deterministic member
+if [[ "${ENS_SIZE:-0}" -gt 2 ]]; then
+  mapfile -t mem_list < <(printf "/mem%03d\n" $(seq 1 "$ENS_SIZE"))
+else
+  mem_list=("/") # if determinitic
+fi
+
+for memdir in "${mem_list[@]}"; do
+  # Determine path
+  if [[ ${#memdir} -gt 1 ]]; then
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
+    mkdir -p "${COMOUT}/prep_ic/${WGF}${memdir}"
+  else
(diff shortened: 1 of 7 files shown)


workflow/rocoto_funcs/prep_ic.py
```diff
@@ -81,7 +72,14 @@ def prep_ic(xmlFile, expdir, do_ensemble=False, spinup_mode=0):
...
+    if do_ensemble:
+        datadep_prod = ""
+        for i in range(1, int(ens_size) + 1):
+            memdirstr = f'/mem{i:03d}'
+            datadep_prod = datadep_prod + f'''\n        <datadep age="00:05:00"><cyclestr offset="-{cyc_interval}:00:00">&COMROOT;/&NET;/&rrfs_ver;/&RUN;.@Y@m@d/@H/fcst/&WGF;{memdirstr}/</cyclestr><cyclestr>mpasout.@Y-@m-@d_@H.00.00.nc</cyclestr></datadep>'''
+    else:
+        datadep_prod = f'''\n        <datadep age="00:05:00"><cyclestr offset="-{cyc_interval}:00:00">&COMROOT;/&NET;/&rrfs_ver;/&RUN;.@Y@m@d/@H/fcst/&WGF;/</cyclestr><cyclestr>mpasout.@Y-@m-@d_@H.00.00.nc</cyclestr></datadep>'''
+
     datadep_spinup = f'''\n        <taskdep task="fcst_spinup" cycle_offset="-1:00:00"/>'''
     if spinup_mode == 0:  # no parallel spinup cycles
```
hu5970, line 84: Do you know what is this for?

scripts/exrrfs_prep_ic.sh
```diff
@@ -137,11 +163,28 @@ if [[ "${PREP_IC_TYPE}" == "jedivar" ]] || [[ "${PREP_IC_TYPE}" == "getkf"  ]];
...
 
-exit 0
+done
+#
+#
+# use xargs to run the command using  one core at a time (this takes a long time!)
+#
+echo "===== CMDFILE ====="
+cat  "$CMDFILE"
+xargs -I {} -P "${SLURM_NTASKS}" sh -c '{}' < "${CMDFILE}"
```
hu5970: SLURM_NTASKS is a SLURM related number. Do we have better way to set parallel core number?

scripts/exrrfs_prep_ic.sh
```diff
@@ -137,11 +163,28 @@ if [[ "${PREP_IC_TYPE}" == "jedivar" ]] || [[ "${PREP_IC_TYPE}" == "getkf"  ]];
...
+      cp "${satbias_path}"/*satbias*.nc  "${umbrella_prep_ic_data}"
       echo "found satbias from ${satbias_path}"
       break
     fi
   done
 fi
 
-exit 0
+done
+#
```
hu5970: This runs after surface cycle, right? If so, surface cycle will fail.

guoqing-noaa:
@chunhuazhou Currently, the ensemble workflow will launch 30 prep_ic tasks, right? If so, I think it should be as fast as the changes proposed in this PR. 

I guess the current issue (include the dead task) is because the copy process needs sufficient memory to complete successfully but prep_ic does not request sufficient memory or to be exclusive on Ursa (Ursa is very aggressive in sharing a node as much as possible among different tasks).

workflow/rocoto_funcs/prep_ic.py
guoqing-noaa: @hu5970 We use one `prep_ic` task for the following spin up situations:
```
cold, spinup_mode = 0   # regular prod cycles, i.e no spin up cycles in an experiment

cold, spinup_mode = 1   #  spin up, cold start
warm, spinup_mode = -1   #  spin up, continue cycling

warm, spinup_mode = -1, prod_switch   #  prod switching from spinup
warm, spinup_mode = -1, regular          #  prod parallel to spinup, continue cycling
```
Check this slide for more details:
https://docs.google.com/presentation/d/1HPx2LzX8Hf9Imztl4OpyXdyTTWoN4hBOId4AP7bhyYM/edit?slide=id.g332010f5a43_4_374#slide=id.g332010f5a43_4_374

scripts/exrrfs_prep_ic.sh
guoqing-noaa: @chunhuazhou `NTASKS` is expected to be a defined env variable from the job card.
In the rocoto workflow, it is already defined in `launch.sh`

https://github.com/NOAA-EMC/rrfs-workflow/blob/3ce4c701c99a93d0ae8e27789f69780da7f7e2c4/workflow/sideload/launch.sh#L15

guoqing-noaa: Also, if we want to move forward with running serial copying in parallel, we would want to be generic, not binding all logics to SLURM only.
We already have a generic `rank_run` tool in rrfs-workflow to be used. I can help with this.

But at this moment, I think 30 copy may work well if we use `--exclusive` or request sufficient memory for each task

hu5970:
Using one node to copy the files is a waste of resources. We need to avoid such setup.

scripts/exrrfs_prep_ic.sh
hu5970: Should use "NTASKS" here.

guoqing-noaa:
> Using one node to copy the files is a waste of resources. We need to avoid such setup.

@hu5970 we don't have to use one node to do copy.
We can just request enough memory just as the changes in this PR and then let SLURM/PBS to manage how to allocate resources efficiently. 

Also, I think initially we plan to let `prep_ic` to do more tasks such as surface updating/soilSurgery etc. So using one node is not that bad in practice, especially if it runs fast.

guoqing-noaa:
There is one drawback to "manually" do 30 parallel copies in 1 or 2 nodes.  Different HPCs have different core/memory per node. 30 parallel "manual" copying may work on Ursa (192 cores) but may NOT work on Hera (40 cores).

hu5970:
> There is one drawback to "manually" do 30 parallel copies in 1 or 2 nodes. Different HPCs have different core/memory per node. 30 parallel "manual" copying may work on Ursa (192 cores) but may NOT work on Hera (40 cores).

We do not need to worry about this. Xarge will run with the core number after -P to run command parallel with the core number and then run rest of the command after the first set finished. So, just the core number ad -P as the core number pernode.

hu5970:
> > Using one node to copy the files is a waste of resources. We need to avoid such setup.
> 
> @hu5970 we don't have to use one node to do copy. We can just request enough memory just as the changes in this PR and then let SLURM/PBS to manage how to allocate resources efficiently.
> 
> Also, I think initially we plan to let `prep_ic` to do more tasks such as surface updating/soilSurgery etc. So using one node is not that bad in practice, especially if it runs fast.

Those surface update and soil surgery are all small single core program. And most of them is actually running for the deterministic cycle only.

scripts/exrrfs_prep_ic.sh
hu5970: rank_run needs extra lib. Using xargs should be good enough for now.

workflow/rocoto_funcs/prep_ic.py
hu5970: spinup_mode needs more discussion. We learn from RRFSv1 that the cycle mode could be very complex and we need to set a good parameter to control all the possible cycle tasks: det spin-up, det prod, enkf prod, enkf spin-up, blending etc. Let though if we can define cycle_mode (0-99) to help?

guoqing-noaa: This workflow is different from RRFSv1, as it will always support coldstart-only forecast experiments and non-spinup experiments. I think we are good for now. We can definitely talk more when we run into any issues.

scripts/exrrfs_prep_ic.sh
guoqing-noaa: @hu5970 What do you mean `rank_run nees extra lib`?

We should not develop a script working only for SLURM.

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
...
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
+    mkdir -p "${COMOUT}/prep_ic/${WGF}${memdir}"
+    pid=$((10#${memdir: -2}-1))
+    export CMDFILE="${DATA}/script_prep_ic_${pid}.sh"
+  else
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}"
+    export CMDFILE="${DATA}/script_prep_ic_0.sh"
+  fi
+
+  mkdir -p "$(dirname "$CMDFILE")"
```
guoqing-noaa: `${DATA}` should be always available in the ex-script, no need to `mkdir -p` here

scripts/exrrfs_prep_ic.sh
```diff
@@ -137,11 +165,24 @@ if [[ "${PREP_IC_TYPE}" == "jedivar" ]] || [[ "${PREP_IC_TYPE}" == "getkf"  ]];
...
-      cp "${satbias_path}"/*satbias*.nc  "${UMBRELLA_PREP_IC_DATA}"
+      echo cp "${satbias_path}"/*satbias*.nc  "${umbrella_prep_ic_data}" >> "$CMDFILE"
       echo "found satbias from ${satbias_path}"
       break
     fi
   done
 fi
 
-exit 0
+done
```
guoqing-noaa: We need to add two spaces to indent lines 42-171 correctly.

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
...
+if [[ "${ENS_SIZE:-0}" -gt 2 ]]; then
+  mapfile -t mem_list < <(printf "/mem%03d\n" $(seq 1 "$ENS_SIZE"))
+else
+  mem_list=("/") # if determinitic
+fi
+
+for memdir in "${mem_list[@]}"; do
+  # Determine path
+  if [[ ${#memdir} -gt 1 ]]; then
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
```
guoqing-noaa, line 55: Suggest changing `umbrella_prep_ic_data` to `umbrella_prep_ic_mem` to better distinguish it from `UMBRELLA_PREP_IC_DATA`

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
   done
 fi
 echo "this cycle is ${start_type} start"
+
+# Populate the list for the ensemble members, or deterministic member
+if [[ "${ENS_SIZE:-0}" -gt 2 ]]; then
+  mapfile -t mem_list < <(printf "/mem%03d\n" $(seq 1 "$ENS_SIZE"))
+else
+  mem_list=("/") # if determinitic
```
guoqing-noaa, line 46: This will create a double `/` situation in line 64:
`thisfile=${COMINrrfs}/${RUN}.${PDY}/${cyc}/ic/${WGF}${memdir}/init.nc`
It will generate something like `..../ic/det//init.nc`. This is expected to be avoided per the NCO standard.

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
...
+fi
+
+for memdir in "${mem_list[@]}"; do
+  # Determine path
+  if [[ ${#memdir} -gt 1 ]]; then
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
+    mkdir -p "${COMOUT}/prep_ic/${WGF}${memdir}"
+    pid=$((10#${memdir: -2}-1))
```
guoqing-noaa: I would think it is more straightforward to generate a list of member numbers fist and add "/" when it is needed.
Let me try a case and post my example here.

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
...
+  mapfile -t mem_list < <(printf "/mem%03d\n" $(seq 1 "$ENS_SIZE"))
+else
+  mem_list=("/") # if determinitic
+fi
+
+for memdir in "${mem_list[@]}"; do
+  # Determine path
+  if [[ ${#memdir} -gt 1 ]]; then
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}${memdir}"
+    mkdir -p "${COMOUT}/prep_ic/${WGF}${memdir}"
```
guoqing-noaa, line 56: Why we create COMOUT directories for each member's prep_ic? Will we save data there?

scripts/exrrfs_prep_ic.sh
```diff
@@ -30,15 +30,41 @@ if (( spinup_mode == -1 )); then
...
+  else
+    umbrella_prep_ic_data="${UMBRELLA_PREP_IC_DATA}"
+    export CMDFILE="${DATA}/script_prep_ic_0.sh"
+  fi
+
+  mkdir -p "$(dirname "$CMDFILE")"
+  : > "$CMDFILE"
+
+  # Create directory safely
+  mkdir -p "${umbrella_prep_ic_data}"
```
guoqing-noaa, line 62: @chunhuazhou FYI, here is an alternate example for lines 34-57:
```
if (( "${ENS_SIZE:-0}" > 1 )); then
  mapfile -t mem_list < <(printf "%03d\n" $(seq 1 "$ENS_SIZE"))
else
  mem_list=("000") # if determinitic
fi

for index in "${mem_list[@]}"; do
  # Determine path
  if (( 10#${index} == 0 )); then
    memdir=""
    umbrella_prep_ic_mem="${UMBRELLA_PREP_IC_DATA}"
    export CMDFILE="${DATA}/script_prep_ic_0.sh"
  else
    memdir="/mem${index}"
    umbrella_prep_ic_mem="${UMBRELLA_PREP_IC_DATA}${memdir}"
    mkdir -p "${umbrella_prep_ic_mem}"
    pid=$((10#${index}-1))
    export CMDFILE="${DATA}/script_prep_ic_${pid}.sh"
  fi
  echo $CMDFILE, $memdir
done
```

scripts/exrrfs_prep_ic.sh
hu5970: xargs is linux command, not for SLURM. It works on wcoss2. Also, WCOSS2 has its own way to run parallel command in within one node.

guoqing-noaa: Ah, thanks for the reminder. I did not pay much attention to this part. I had thought it is similar to this:
https://github.com/NOAA-EMC/rrfs-workflow/blob/3ce4c701c99a93d0ae8e27789f69780da7f7e2c4/scripts/exrrfs_ensmean.sh#L89

With that said, `xargs` cannot distribute tasks across multiple nodes.
For example, when we run NA3km or global-15-3km or more ensembles with few cores, we may need to use 2 or more nodes to copy files in parallel because the file size is much larger and we need more memories for each copying.

`rank_run` is a simple replacement tool to NCO's `CFP` in non-NCO machines. It does real parallelism and can use multiple nodes. It has no other library dependencies.

scripts/exrrfs_prep_ic.sh
```diff
@@ -90,22 +117,23 @@
...
+      if [[ -r "${umbrella_prep_ic_data}/init.nc" ]]; then
+        to_file="${umbrella_prep_ic_data}/init.nc"
+      elif [[ -r "${umbrella_prep_ic_data}/mpasout.nc" ]]; then
+        to_file="${umbrella_prep_ic_data}/mpasout.nc"
       fi
       echo "surface update from ${thisfile} to ${to_file}"
-      ncks -O -C -x -v ${var_list} "${to_file}"  tmp.nc
-      ncks -A -v ${var_list} "${UMBRELLA_PREP_IC_DATA}/mpas_sfc.nc" tmp.nc
-      mv tmp.nc "${to_file}"
+      echo ncks -O -C -x -v ${var_list} "${to_file}"  tmp.nc >> "$CMDFILE"
```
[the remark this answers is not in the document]
guoqing-noaa: @chunhuazhou Could you try this to see whether it works and addresses this ShellCheck complaints:
```
echo "ncks -O -C -x -v ${var_list}  ${to_file}"  tmp.nc >> "$CMDFILE"
```
Similarly for lines 135, 136, 168

scripts/exrrfs_prep_ic.sh
```diff
@@ -81,6 +109,7 @@ fi
 #
```
hu5970, line 109: Surface cycle have no relation with background copy. Better to separate to two sections?

Issue #1260, opened 2026-01-12
[rrfs-nco] Load Python in the build script


WeiWei-NCO:
In a recent build done by NCO, we noticed that the python module must be loaded explicitly before running `app_build.sh`. See #1258 

Please load python in the build script with the version specified in `/versions/build.ver`.

JamesPolly-NCO:
Fixed in #1297; closing.

Issue #1202, opened 2025-12-11
[rrfs-nco] Remove Loading of prod_util Module


JamesPolly-NCO:
**Problem:**
prod_util is currently being loaded in the .ecf job cards with a prod_util_ver defined in run.ver

prod_util:
- should be loaded by the ecflow header files, in a centralized location for the whole production suite.
- should not be loaded by individual packages (e.g. when prod_util is upgraded we don't have to upgrade all the packages).

**Solution:**
Remove all loads of the prod_util module and remove prod_util_ver from run.ver.

**Discussion:**
How does this change impact EMC runs, even after NCO takes over the parallel?

JamesPolly-NCO:
Fixed in #1297; closing.

PR #1297, opened 2026-02-03
[rrfs-nco] Remove loading of prod_util; remove lmod-setup.csh; fix python loading in app_build.sh

JamesPolly-NCO:
## DESCRIPTION OF CHANGES: 

- fixes issue where nco ops accounts have to manually load python before running `app_build.sh --extrn`
- fixes issue where prod_util is loaded twice in ops
- removes instances of prod_util_ver
- removes emc paths (obsproc) in run.ver
- removes .csh file which is no longer used and not permitted in ops

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [x] Others:
  - build works as expected without manual loading of python module
  - build and runs work as expected with removal of lmod-setup.csh and prod_util loads

## ISSUE: 

- Fixes the issue(s) mentioned in #1202 
- Fixes the issue(s) mentioned in #1260


MatthewPyle-NOAA (approved):
Thanks for these cleanups!

sorc/app_build.sh
```diff
@@ -296,7 +296,9 @@ if [ "${EXTRN}" = true ]; then
 
   # run check-out
   python --version 1>/dev/null 2>/dev/null
-  if [[ $? -ne 0 ]]; then
+  if [[ $(which python) == "/usr/bin/python" ]]; then
+       module unload python
+       module load PrgEnv-intel
        module load python
```
WeiWei-NCO, line 302: @JamesPolly-NCO Can you update them to use the particular versions defined in build.ver? for both prgenv and python

ecf/include/head.h
```diff
@@ -55,7 +55,6 @@ if [ -d /apps/ops/prod ]; then # On WCOSS2
   set -x
   . ${ECF_ROOT}/versions/run.ver
   set +x
-  module load prod_util/${prod_util_ver}
```
WeiWei-NCO, line 58: It's the header used by EMC runs. We probably want to keep it here.

sorc/app_build.sh
JamesPolly-NCO: Good call, I'll add it.

ecf/include/head.h
JamesPolly-NCO: This is the header used by EMC runs.

I took this out because we removed prod_util_ver from build.ver, and we did this because we don't want multiple versions of prod_util potentially being used. Keep it all in the header file.

We talked about this with Lin, and he said he'd use our header.

WeiWei-NCO: Sorry where do EMC load prod_util in their runs? 

There's only 1 head.h can be used at a time. If they use this header file, they will need this line. If they use NCO head.h, I'd replace this file with NCO version.

PR #1306, opened 2026-02-06
[rrfs-mpas-jedi] changes for real-time rrfsx on ursa

chunhuazhou:
## DESCRIPTION OF CHANGES: 
changes for real-time rrfsx on ursa: 
    - Add exp files for real-time rrfsx on ursa
    - Remove ensemble data dependency for real-time runs, and add the option to fall back to 3DVar if no enough ensemble files


scripts/exrrfs_jedivar.sh
```diff
@@ -84,6 +84,17 @@ source "${USHrrfs}/copy_obs.sh" "jedivar"
 #
 source "${USHrrfs}/find_ensembles.sh"
 #
+# check number of ensemble files, if not enough, default to pure 3DVar
+#
+ens_count=$(find ens -path "ens/mem*.nc" -type f | wc -l)
+if [ "${ens_count}" -lt 30 ]; then
```
guoqing-noaa: Line 90:  use the double parentheses `(( ))` instead of `[[ ]]` for arithmetic operations and comparisons. 
Suggest changing to 
```
if ((ens_count < 30)); then
```

guoqing-noaa: Do we have `ENS_SIZE` in the jedivar task? If so, we may want to use it instead of `30`.

workflow/rocoto_funcs/jedivar.py
```diff
@@ -102,6 +102,9 @@ def jedivar(xmlFile, expdir, do_spinup=False):
       <datadep age="00:05:00"><cyclestr offset="-6:00:00">{HYB_ENS_PATH}/{RUN}.@Y@m@d/@H/ic/enkf/mem030/init.nc</cyclestr></datadep>
     </or>'''
 
+    if realtime.upper() == "TRUE":
```
guoqing-noaa: Can we move lines 105-106 to before line 75 and change `if` to `elif` in line 75?

workflow/exp/rt_ursa/exp.rrfsv2x
guoqing-noaa, line 1: Are we okay to remove rt_jet now? Thanks!

scripts/exrrfs_jedivar.sh
guoqing-noaa: @chunhuazhou I think the current logic here only works for the realtime runs. In retro runs, we may choose to enforce a strict ensemble dependency. So, I think we need to utilize the `HYB_ENS_TYPE` variable:
```
export HYB_ENS_TYPE=1      
# 1.rrfs; 2. downscaled GEFS/GDAS; 0.TBD on the fly (rrfs->GDAS/GEFS->3DVAR)
```
The value `0` has been reserved for the situation to determine what ensembles to use on the fly. Thanks!

workflow/rocoto_funcs/jedivar.py
guoqing-noaa: Considering the reserved`HYB_ENS_TYPE=0`, Line 105 is better changed to:
```
elif HYB_WGT_ENS != "0" and HYB_WGT_ENS != "0.0" and HYB_ENS_TYPE == "0":  # TBD on the fly (rrfs->GDAS->3DVAR)
```
And then we need to `export HYB_ENS_TYPE=0` in the exp.setup files and also pass this variable to the jedivar task

workflow/exp/rt_ursa/exp.rrfsv2x
chunhuazhou: Let's keep it for now since we are still running real-time experiments on Jet. We can come back and remove them later.

workflow/exp/rt_ursa/exp.rrfsx
```diff
@@ -0,0 +1,120 @@
...
+# Aloha! This file is to setup some top level experiment options
+#
+# shellcheck disable=all
+export NET=rrfs                    # rrfs, rtma, 
+export MESH_NAME=conus3km         # conus12km, conus3km, atl3km, etc
+export WGF="det"   # working group function = "det", "enkf", "ens", "firewx"
+export EXP_NAME=rrfsx
+export VERSION=$(cat ./VERSION)
+export TAG=rrfsx
+export OPSROOT=/scratch4/BMC/rtrr/NCO_dirs/MPAS_${EXP_NAME}
```
guoqing-noaa: Should we change `MPAS_${EXP_NAME}` to `${EXP_NAME}`, i.e. using `rrfsx` instead of `MPAS_rrfsx`?

ShunLiu-NOAA (dismissed):

guoqing-noaa:
@chunhuazhou `RRFSv2X` looks a little bit strange. Did you have new offline discussions about changing `rrfsx` to `RRFSv2X`?

MatthewPyle-NOAA:
I think rrfsv2x would work well in terms of an experiment name.  We did have a meeting where this naming convention came up, but don't think we were intending for it to be capitalized for everything.

guoqing-noaa:
@MatthewPyle-NOAA Thanks for the background information.

Just curious, when we go to rrfsv3, we want to change to `rrfsv3x`, and then `rrfsv4x`, etc?

MatthewPyle-NOAA:
@guoqing-noaa I'd say yes, assuming there is something RRFS beyond version 2

guoqing-noaa:
> @guoqing-noaa I'd say yes, assuming there is something RRFS beyond version 2

Thanks for the reply. Hope so, :)

chunhuazhou:
@guoqing-noaa I got this name from Terra yesterday. Thanks!
@MatthewPyle-NOAA Thanks for chiming in!

guoqing-noaa (approved):

MatthewPyle-NOAA (approved):

PR #1308, opened 2026-02-09
[rrfs-mpas-jed] Feature/addamv da capability

keenaneure:
This PR makes necessary changes to ~/ush/copy_obs.sh and ~/scripts/exrrfs_ioda_bufr.sh in order to assimilate GOES AMVs. This includes allocating variables, pointing to the non-satellite radiance offline domain check, and copying the observation files.


scripts/exrrfs_ioda_bufr.sh
```diff
@@ -45,7 +45,8 @@ if (( ${YAML_GEN_METHOD:-1} == 2 )); then
   ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_msonet.nc
   ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_proflr.nc
   ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_rassda.nc
-  ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_satwnd.nc
+  ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_satwnd.abi_goes-16.nc
+  ${cpreq} "${FIXrrfs}"/jedi/ioda_empty.nc ioda_satwnd.abi_goes-18.nc
```
delippi: @keenaneure you likely don't use the `YAML_GEN_METHOD=2` do you? This will likely break because the structure of `ioda_empty.nc` file won't be what JEDI expects--I made a `ioda_empty_satwnd.nc` that I've used in [dev-sci] branch. JEDI also doesn't necessarily need an empty file either since you likely use `missing file action: "warn"`. If you do you yaml gen method 2 and you do want to copy an empty file in first, let me know and I will get you a file that should work. Otherwise, I would suggest removing this. It probably never came up before since I'm assuming nothing was ever trying to use `ioda_satwnd.nc`?

keenaneure: No, I did not realize there was a YAML_GEN_METHOD for this, but I did copy this change over from one of my forks that was successful. Since we split these observations by satID, I have not proceeded at all with using ioda_satwnd.nc...unless these should be appended together? What is the difference in structure between ioda_empty.nc and ioda_empty_satwnd.nc?

delippi: Well, they're both IODA obs files, but I think jedi still expects some particular metadata that the regular ioda_empty.nc doesn't have.

You don't want `ioda_satwnd.nc` either. That is supposed to be the satwnds from prepbufr which we do not use.

I think you can just remove these two lines from the YAML_GEN_METHOD==2 block.

keenaneure: Without these two lines, wouldn't the ioda_bufr process crash? Where would the two necessary files come from if not this line?

delippi: I don't think it will crash because I don't think that part of the code is being executed because I don't think you are using `YAML_GEN_METHOD=2`. If you were using method 2, and you had an instance of no satwnd obs using ioda_empty.nc as the file--that would crash.

The `cp ioda_empty.nc ioda_xyz.nc` is just an initial empty file in the case that bufr2ioda doesn't create one due to zero observations. You don't actually need to do that. I only had it in the workflow this way to more easily monitor obs. My guess is you use the default YAML_GEN_METHOD which doesn't do any of this.

You can double check by

1. Running a jedivar task with the empty ioda file (so make sure you copy the `ioda_empty.nc` file as `ioda_satwnd.abi_goes-16.nc` and also make sure to skip the bufr2ioda step to ensure using an empty file. I would expect this to fail because of missing meta data in the ioda_empty file.)
2. Removing these lines and see that it doesn't crash.

> Where would the two necessary files come from if not this line?

It simply comes from the bufr2ioda step. The empty file is the initial file and then if there are obs, it gets overwritten.

keenaneure: I removed those two lines just now.

scripts/exrrfs_ioda_bufr.sh
```diff
@@ -113,7 +115,8 @@
 ./bufr2ioda_ztd.py -c bufr2ioda_0.json
 
 # SATWND
-#./bufr2ioda_satwnd.py -c bufr2ioda_0.json
+./gen_bufr2ioda_json.py -t bufr2ioda_satwnd_amv_goes.json -o bufr2ioda_satwnd_amv_goes_0.json
+./bufr2ioda_satwnd_amv_goes.py -c bufr2ioda_satwnd_amv_goes_0.json >> $pgmout
```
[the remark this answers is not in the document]
guoqing-noaa: @keenaneure You can remove `>> $pgmout` in line 117 to address the Shell linting comment.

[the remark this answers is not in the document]
keenaneure: Done.

guoqing-noaa (approved):
LGTM. Thanks for adding AMV DA, @keenaneure

MatthewPyle-NOAA (approved):

ShunLiu-NOAA (approved):

PR #1330, opened 2026-02-19
update the clean task to allow retro and realtime use the same clean mode

guoqing-noaa:
Currently, we have two different clean modes for retro and realtime runs:
- Realtime runs: Each cycle cleans cycles older than `STMP_RETENTION_CYCS` (clean mode 1).
- Retro runs: Each cycle only cleans itself after all tasks in that cycle have successfully completed (clean mode 2).

Allowing retro runs to “clean cycles older than `STMP_RETENTION_CYCS`” was risky because, in retro experiments, old and latest cycles can be active simultaneously. If a latest cycle attempts to clean an older cycle that is still active, it can cause that cycle to fail and then break the retro experiment.

This PR adds the ability to query the experiment’s Rocoto database to verify that a cycle has completed before cleaning it. With this improvement, retro runs can safely use the same clean mode as realtime runs. 

This provides a consistent understanding of how the clean task works and allows users to retain the most recent cycles for inspection or debugging while purging other cycles to save disk space.

Introduce `STMP_KEPT_TASKS` to exclude some run directories (for example, jedivar) from being purged by the clean task. This capability is requested by users who want to save more cycles of jedivar run directories for investigation while other run directories can still be cleaned as normal.

### DIFF
--- a/workflow/config_resources/config.base
+++ b/workflow/config_resources/config.base
@@ -39,7 +39,7 @@ if ${DO_CLEAN:-false} && ! ${REALTIME:-false}; then
   export COM_LBC_RETENTION_CYCS=${COM_RETENTION_CYCS:-2400}
   export LOG_RETENTION_CYCS=${LOG_RETENTION_CYCS:-2400}
   export CHECK_IS_CYC_DONE=${CHECK_IS_CYC_DONE:-true}
-  export STMP_KEPT_TASKS=""  # eg. jedivar or jedivar,mpassit,upp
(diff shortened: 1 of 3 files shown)


guoqing-noaa:
tested 24 cycles and worked as expected

HaidaoLin-NOAA (approved):

chunhuazhou (approved):

MatthewPyle-NOAA (approved):

PR #1342, opened 2026-02-27
[rrfs-nco] Remove Development Scripts for Ops Builds

JamesPolly-NCO:
## DESCRIPTION OF CHANGES: 

- Add parm file containing a list of developer-only codes that should be removed before running in ops/production.
- Use above file in `sorc/app_build.sh` to remove these files at the end of the build.
- Update the local git index to ignore the deletion of these version-controlled files (`git status` will not show them as deleted).

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [x] Others:
  - Ran in NCO real-time parallel with "dev-only" files removed.
  - Tested build and removal of "dev-only" files.

## CONTRIBUTORS (optional): 

@MatthewPyle-NOAA @WeiWei-NCO Please review.


sorc/app_build.sh
```diff
@@ -350,8 +372,7 @@ if [ "${EXTRN}" = true ]; then
   fi
 
   # run check-out
-  python --version 1>/dev/null 2>/dev/null
```
JamesPolly-NCO, line 353: As the ops.para user, this line fails on Cactus but not on Dogwood. Removing python from the check.

WeiWei-NCO: Hold off on this change. We need to first understand why Cactus and Dogwood are different.

JamesPolly-NCO: Thanks for this input. I am happy to wait to get input from GDIT on this difference between cactus and dogwood.

That said, I think I like the reliance on boolean $NCO_BUILD better in general.

MatthewPyle-NOAA (approved):
Very slick.  Initially was going to ask if the NCO_BUILD switch should be moved higher in the app_build.sh script, but then saw that it is intended to be automatically reset for a production build.   But I did force it to be true, and confirmed that it worked as advertised.

PR #1355, opened 2026-03-05
[rrfs-mpas-jedi]Fix the ensemble data check for jedivar

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Fix the ensemble data check for jedivar. When HYB_ENS_TYPE=0, the jedivar task searches for ensemble files (RRFS, then GDAS). If an insufficient number of files are found, the system defaults to 3DVAR. This approach is ideal for real-time runs where maintaining cycle continuity is prioritized over strict ensemble dependency.
Minor fix for dependency in save_for_next.py

## TESTS CONDUCTED: 
Tested for conus12km.

### DIFF
--- a/scripts/exrrfs_jedivar.sh
+++ b/scripts/exrrfs_jedivar.sh
@@ -86,7 +86,7 @@ source "${USHrrfs}/find_ensembles.sh"
 #
 # For HYB_ENS_TYPE=0, check number of ensemble files, if not enough, default to pure 3DVar
 #
-if [[ ${HYB_ENS_TYPE} == 0 ]]; then
+if (( HYB_ENS_TYPE == 0 )) ; then
   ens_size=$(( 10#${ENS_SIZE} ))
   ens_count=$(find ens -name "mem*.nc" | wc -l)
   if (( ens_count < ens_size )); then
--- a/scripts/exrrfs_jedivar.sh
+++ b/scripts/exrrfs_jedivar.sh
@@ -84,16 +84,18 @@ source "${USHrrfs}/copy_obs.sh" "jedivar"
 #
 source "${USHrrfs}/find_ensembles.sh"
 #
-# check number of ensemble files, if not enough, default to pure 3DVar
+# For HYB_ENS_TYPE=0, check number of ensemble files, if not enough, default to pure 3DVar
 #
-ens_size=$(( 10#${ENS_SIZE} ))
-ens_count=$(find ens -name "mem*.nc" | wc -l)
-if (( ens_count < ens_size )); then
-   echo "Number of ensemble files is ${ens_count}, less than 30, default to 3DVar"
-   export HYB_WGT_ENS=0.0
-   export HYB_WGT_STATIC=1.0
-else
-   echo "Found ${ens_count} ensemble files"
+if [[ ${HYB_ENS_TYPE} == 0 ]]; then
+  ens_size=$(( 10#${ENS_SIZE} ))
+  ens_count=$(find ens -name "mem*.nc" | wc -l)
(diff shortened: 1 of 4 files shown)


scripts/exrrfs_jedivar.sh
```diff
@@ -84,16 +84,18 @@ source "${USHrrfs}/copy_obs.sh" "jedivar"
...
-ens_count=$(find ens -name "mem*.nc" | wc -l)
-if (( ens_count < ens_size )); then
-   echo "Number of ensemble files is ${ens_count}, less than 30, default to 3DVar"
-   export HYB_WGT_ENS=0.0
-   export HYB_WGT_STATIC=1.0
-else
-   echo "Found ${ens_count} ensemble files"
+# For HYB_ENS_TYPE=0, check number of ensemble files, if not enough, default to pure 3DVar
+#
+if [[ ${HYB_ENS_TYPE} == 0 ]]; then
```
guoqing-noaa: suggest changing to `if (( HYB_ENS_TYPE == 0 )); then`

workflow/rocoto_funcs/save_for_next.py
```diff
@@ -37,7 +37,9 @@ def save_for_next(xmlFile, expdir, do_ensemble=False, do_spinup=False):
         timedep = f'\n    <timedep><cyclestr offset="{starttime}">@Y@m@d@H@M00</cyclestr></timedep>'
     #
     dependencies = f'''
-  <dependency>{timedep}{datadep}
+  <dependency>
+    <and>{timedep}{datadep}
```
guoqing-noaa: @chunhuazhou I think by default, we have the `<and>`  relation.. Without the explicit `<and>...</and>`, it is still the and relation. Did you get any errors without it?

chunhuazhou: yes, without <and></and> it can't handle the 30 files data dependency.

guoqing-noaa: It looks like I misunderstood something. Thanks for testing! 

Could you remove two spaces before lines 41 and 42? It is okay to not indent <and>, but we want to indent other dependencies. Thanks!

chunhuazhou: Done. Thanks!

ush/find_ensembles.sh
```diff
@@ -2,47 +2,60 @@
 # find ensemble forecasts based on user settings
 #
 # shellcheck disable=SC2154,SC2153
+set -x
```
guoqing-noaa: We don't need `set -x` here as this file is sourced in jedivar.sh

ush/find_ensembles.sh
```diff
@@ -2,47 +2,60 @@
...
+         for (( iii=1; iii<=ens_size; iii=iii+1 )); do
+            memid=$(printf %03d "${iii}")
+            ln -s "${ensdir}/ic/enkf/mem${memid}/init.nc" "ens/mem${memid}.nc"
+         done
+         found_data=true
        fi
-    done
+     fi
+# Break if we actually found and linked files
+     if [[ "$found_data" == "true" ]]; then break; fi
```
guoqing-noaa: To be consistent,  change`$found_data` -> `${found_data}` in lines 53, 57

chunhuazhou: Done. Thanks!

ush/find_ensembles.sh
chunhuazhou: Removed. Thanks!

MatthewPyle-NOAA (approved):

guoqing-noaa (approved):

PR #1376, opened 2026-03-12
[rrfs-mpas-jedi]Remove coldda dep

chunhuazhou:
## DESCRIPTION OF CHANGES: 
This PR includes:
- Fixing a bug in prep_ic. This issue came up when the external data source of surface state was not successfully generated and the variable name "thisfile" ended up pointing to the same file for the source and target files for the surface update. The fix includes a new variable name for the surface source files and more safety checks.
- Applying configurable cyclethrottle, taskthrottle  and cyclelifespan to the real-time runs 
- Removing data assimilation dependency for coldhrs when COLDSTART_CYCS_DO_DA=false. This avoids waiting for the data assimilation to complete when the data assimilation tasks are not actually needed for the cold start cycles.
- Adding more graphinfo files for conus3km
- Updating exp files

## TESTS CONDUCTED: 
Tested 3 cycles in retro on Ursa, for both deterministic and getkf cycling runs, with or without data assimilation.


scripts/exrrfs_prep_ic.sh
```diff
@@ -103,43 +103,59 @@ for index in "${mem_list[@]}"; do # loop through all the members
...
-        PDYii=${CDATE:0:8}
-        cycii=${CDATE:8:2}
-        thisfile_init=${COMINsfc}/${RUN}.${PDYii}/${cycii}/ic/${WGF}${memdir}/init.nc
-        if [[ -r ${thisfile_init} ]]; then
-          var_list="smois,snow,snowh,snowc,sst,canwat,tslb,skintemp,landmask,isltyp,ivgtyp"
-          thisfile=${thisfile_init}
+      # if no mpasout files, use init.nc from another run
+      if [[ ! -s "${source_file}" ]]; then
+        if [[  "${COMINsfc}" == "${COMINrrfs}" ]]; then
+	  echo "SFC_UPDATE: ${COMINsfc} = ${COMINrrfs}, same init.nc,  skip sfc_update!"
```
guoqing-noaa: Line 127 does not have the correct indentation

guoqing-noaa: Line 126: should we check whether `if [[  "${COMINsfc}" == "${COMINrrfs}" ]] && [[  "${COMINsfc}" == "" ]]`?

For retros, are users required to define `COMINsfc`?

workflow/rocoto_funcs/base.py
```diff
@@ -158,15 +158,16 @@ def header_end(xmlFile):
 
 def wflow_begin(xmlFile):
     realtime = os.getenv("REALTIME", "false").upper()
-    cyclethrottle = os.getenv("RETRO_CYCLETHROTTLE", "3")
-    taskthrottle = os.getenv("RETRO_TASKTHROTTLE", "30")
+    cyclethrottle = os.getenv("CYCLETHROTTLE", "3")
+    taskthrottle = os.getenv("TASKTHROTTLE", "30")
+    cyclelifespan = os.getenv("CYCLELIFESPAN", "2")
```
guoqing-noaa, line 163: `RETRO_TASKTHROTTLE` and `RETRO_CYCLETHROTTLE` are changed. Do we have any of them in the example exp files?

scripts/exrrfs_prep_ic.sh
chunhuazhou: The check for COMINsfc is in jobs/JRRFS_PREP_IC. If not defined, it will default to COMINrrfs.

workflow/rocoto_funcs/jedivar.py
```diff
@@ -107,33 +110,19 @@ def jedivar(xmlFile, expdir, do_spinup=False):
...
-    </or>'''
+            strneqs += f'\n    <strneq><left><cyclestr>@H</cyclestr></left><right>{hr}</right></strneq>'
 
-    else:
-        final_ens_dep = ens_dep
     #
     dependencies = f'''
   <dependency>
-  <and>{timedep}
+  <and>{timedep}{strneqs}
```
guoqing-noaa, line 125: We need an `{streqs}` here, right? For those cycles, `fcst` still need to depend on `prep_ic`

chunhuazhou: We don't actually need {streqs}. Since we don't have separate cycledef for the jedivar/getkf tasks, the logic here is to kick off these tasks only for the cycles with DA.

workflow/rocoto_funcs/base.py
chunhuazhou: I will update the exp files that were using RETRO_TASKTHROTTLE and RETRO_CYCLETHROTTLE to use the new names (no "RETRO_")

workflow/rocoto_funcs/jedivar.py
guoqing-noaa: Got it. But could you keep the generation of `streqs` (just not use it)? I will need it for a following PR. Thanks!

scripts/exrrfs_prep_ic.sh
guoqing-noaa: Got it. Thanks!

workflow/rocoto_funcs/jedivar.py
guoqing-noaa: @chunhuazhou Sorry, it looks like I don't need `streqs`. So the current form is good. Thanks!

guoqing-noaa (approved):
LGTM. Thanks, @chunhuazhou!

workflow/exp/rt_ursa/exp.conus3km_RAPic
MatthewPyle-NOAA, line 1: Does this indicate that this configuration is going from 65 levels to 60 levels?

MatthewPyle-NOAA (commented):
One small question, but otherwise looked good to me.

workflow/exp/rt_ursa/exp.conus3km_RAPic
chunhuazhou: This exp file generates an init.nc from RAP data, specifically to pull the surface state into the RRFSv2X run. Because RAP has a limited model top, we use L60; note that L65 is incompatible with RAP data. Thanks!

guoqing-noaa: @MatthewPyle-NOAA as Chunhua mentioned, L60 is the only choice if initializing from RAP. Thanks!

MatthewPyle-NOAA (approved):

MatthewPyle-NOAA (dismissed):

MatthewPyle-NOAA:
It is telling me the "head branch is out of date."  I feel like some weird stuff is happening with github right now (earlier I'd approved it and it didn't give me the option to merge).  Will wait until later to see what is going on with it.

MatthewPyle-NOAA (approved):
Reapproving and hopefully can merge now

PR #1377, opened 2026-03-13
[dev-sci] Modify clean task to work with retros using Rocoto

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

As we begin running retros on conus3km, we have observed that storing both the GSI and JEDI analyses during cycling can take up a lot of space. For example, a single cycling directory on stmp takes up nearly 7 TB for a EnKF retro. Therefore, we need to clean up these directories as we go. 

There is currently a `CLEAN` task and associated `exrrfs_clean.ksh` script. However, this script and task are designed for realtime applications and do not work with retros. Here, we update this task and script to work for retros. Changes include: 

* Clean task now defines the `currentime` variable based on the cycle time for retros. It previously defaulted to the actual current time with `CDATE` 
* For retros, the clean task now checks for either the deterministic or ensemble analysis tasks to be completed before running such that it can run for every cycle. This means it always runs whether performing a EnKF/Var retro or if product generation is disabled
* The clean task is now controlled by its own cycledef (`clean_cycledefs`) rather than relying on the archive task cycle definition (which is not always enabled for retros)
* The `cyclethrottle` value can now be configured through `config.sh`. Previously it was hardcoded to 12 cycles. This required the cleaning window to exceed 12 cycles to avoid deleting directories that might still be in use

For reference, here are my current clean settings for my conus3km EnKF retros as defined in `config.sh`: 

```
CYCLE_THROTTLE=6
CLEAN_CYCLEDEF="00 00-23 ${CYCLEDAY} ${CYCLEMONTH} ${STARTYEAR} *"
CLEAN_OLDPROD_HRS=999
CLEAN_OLDLOG_HRS=999
CLEAN_NWGES_HRS=999
CLEAN_OLDRUN_HRS=12
CLEAN_OLDFCST_HRS=12
CLEAN_OLDSTMPPOST_HRS=12
```

With this configuration, the stmp directories older than 12 hours relative to the current cycle time are removed. The prod and log directories are fully retained for validation, and the nwges directories are also preserved in order to retain ensemble backgrounds needed for JEDIVAR retros.

One more note: this PR also contains a bug fix for the JCB config files for GETKF. The solver needs to use the Halo obs distribution but was incorrectly set to RoundRobin before. 

## TESTS CONDUCTED: 

This was tested with a small, 3-member EnKF retro on conus12km to confirm that all works as expected.

### Machines/Platforms:

- WCOSS2
  - [x] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [x] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
None

### DIFF
--- a/parm/FV3LAM_wflow.xml
+++ b/parm/FV3LAM_wflow.xml
@@ -399,7 +399,7 @@ define resources used for each tasks
 ]>
 
 {%- if do_retro %}
-<workflow realtime="F" scheduler="&SCHED;" cyclethrottle="12">
+<workflow realtime="F" scheduler="&SCHED;" cyclethrottle="{{ cycle_throttle }}">
 {% else %}
 <workflow realtime="T" scheduler="&SCHED;" cyclethrottle="26" cyclelifespan="01:00:00:00">
 {%- endif %}
@@ -418,6 +418,7 @@ define resources used for each tasks
 {%- endif %}
   <cycledef group="recentercyc"> {{ recenter_cycledef }} </cycledef>
   <cycledef group="archive">  {{ archive_cycledef }} </cycledef> 
+  <cycledef group="clean">  {{ clean_cycledef }} </cycledef> 
 
   <log>
     <cyclestr>&LOGDIR;/FV3LAM_wflow_&TAG;.log</cyclestr>
@@ -4094,7 +4095,7 @@ MODULES_RUN_TASK_FP script.
 ************************************************************************
 ************************************************************************
 -->
-  <task name="&CLEAN_TN;" cycledefs="archive" maxtries="{{ maxtries_run_post }}">
+  <task name="&CLEAN_TN;" cycledefs="clean" maxtries="{{ maxtries_run_post }}">
 
     &RSRV_POST;
 
@@ -4107,6 +4108,7 @@ MODULES_RUN_TASK_FP script.
     <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
     <envar><name>LOGDIR</name><value>&LOGDIR;</value></envar>
     <envar><name>CDATE</name><value><cyclestr offset="-4:00:00">@Y@m@d@H</cyclestr></value></envar>
+    <envar><name>CYCLEDATE</name><value><cyclestr>@Y@m@d@H</cyclestr></value></envar>
     <envar><name>NWGES_BASEDIR</name><value><cyclestr>&NWGES_BASEDIR;</cyclestr></value></envar>
 {%- if do_ensemble %}
     <envar><name>nens</name><value><cyclestr>{{ num_ens_members }}</cyclestr></value></envar>
@@ -4115,6 +4117,7 @@ MODULES_RUN_TASK_FP script.
 {%- endif %}
 
     <dependency>
+{%- if not do_retro %}
        <or>
           <and>
             <or>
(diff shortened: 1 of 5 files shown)


HuiLiu-NOAA:
maybe,  more specific/clear, "CLEAN_OLDSTMPPOST_HRS" refers to the hours older than the current/oldest **analysis/forecast** cycle time? This will not be affected by the CYCLE_THROTTLE?   my experience is that all the cycles older than the current/oldest analysis cycle can be removed regardless of the CYCLE_THROTTLE.

guoqing-noaa:
> maybe, more specific/clear, "CLEAN_OLDSTMPPOST_HRS" refers to the hours older than the current/oldest **analysis/forecast** cycle time? This will not be affected by the CYCLE_THROTTLE? my experience is that all the cycles older than the current/oldest analysis cycle can be removed regardless of the CYCLE_THROTTLE.

FYI, I recently worked on fine tuning the CLEAN task for RRFSv2 a lot and can provide some input on this:
cleaning old cycles works very well for realtime runs, but it is complicated in the retro run situation.
If one only activates the recent few cycles and never runs two cycles far away from each other, that's also fine.

But in practice, it is very possible (quite a few rrfs.v2 developers ran into this situation) that the current cycle is at May07_00z, but due to some reason, we still have active cycles 12, 24 or even 48 cycles before, such as May05_00z may still be active. So apparently, May07_00z cannot remove the May05_00z cycle in this sitution.  So we added a helper Python function `is_cycle_done` to make sure we only clean succeeded cycles.

https://github.com/NOAA-EMC/rrfs-workflow/blob/a740d40cbe27f004985b93a4b6baee8b13761a05/workflow/sideload/clean.py#L22-L38

ShunLiu-NOAA (approved):

SamuelDegelia-NOAA:
> maybe, more specific/clear, "CLEAN_OLDSTMPPOST_HRS" refers to the hours older than the current/oldest **analysis/forecast** cycle time? This will not be affected by the CYCLE_THROTTLE? my experience is that all the cycles older than the current/oldest analysis cycle can be removed regardless of the CYCLE_THROTTLE.

The `CLEAN` task now runs during each cycle in the workflow. These `CLEAN_OLD*` settings will remove those respective directories at N hours older than the cycle that the task is running on. With the previous default setting for `cyclethrottle=12`, it was common to see the 00z and 12z cycles running at the same time. If the 12z cycle happened to finish earlier, then using `CLEAN_OLD*=12` would risk removing the 00z cycle directories. That is why I modified `cyclethrottle` to be a variable that can now be defined in `config.sh` so that we can ensure that it is always less than the `CLEAN_OLD*` settings. 

Regarding always cleaning the previous cycle, it is possible to do this if you want by setting `CLEAN_OLD*=1`. I do not plan on doing that for my retros since I like to keep a number of cycles available to check the log files and potentially restart if there was a problem that did cause the job to fail. But if you are comfortable doing it, that is possible now after this PR.

MatthewPyle-NOAA (commented):
The RDAS template file chnage (RoundRobin --> Halo) didn't seem to be mentioned or in the overall flow of this PR, so making sure it is intended.

SamuelDegelia-NOAA:
Thanks @MatthewPyle-NOAA, I forgot to mention that. I will add it to the description. 

But yes, that change is intended. It's just a minor bug fix. The solver needs to run with Halo obs distribution. Those sections of the yaml were added in a recent PR since JCB modified how localization/distribution settings are handled. But I had a typo when I added the changes into the workflow.

HuiLiu-NOAA (commented):
Thanks @SamuelDegelia-NOAA for this.

PR #1380, opened 2026-03-13
Add ability to specify GOES IDs for nonvariational cloud analysis

ShawnMurdzek-NOAA:
## DESCRIPTION OF CHANGES: 

The nonvariational cloud analysis relies on cloud-top observations from GOES satellites in BUFR format. In order to decode these observations using `RRFS_UTILS/larccld.fd`, the IDs for GOES-East and GOES-West must be provided. The IDs of the satellites occupying the position of GOES-East and GOES-West change with time, so the GOES IDs required by the nonvariational cloud analysis are not constant. Previously, these GOES IDs were hardcoded in `RRFS_UTILS/larccld.fd`. Updates to `RRFS_UTILS/larccld.fd` have made these IDs namelist options (https://github.com/NOAA-GSL/RRFS_UTILS/pull/10). This PR adds the GOES IDs as user-specified options that can be set in the `exp.*` files used to generate the workflow, which gives greater flexibility to the user to choose the appropriate GOES IDs for the retro period being run.

I'm open to discussion regarding the best way to handle the setting of GOES IDs in `rrfs-workflow`. The implementation in this PR gives the user the power to specify the IDs that are appropriate for their retro, but if they specify the incorrect IDs (or don't specify any IDs and the default values are not appropriate), then the nonvariational cloud analysis will be suboptimal and the user may never know. Automatically setting the GOES IDs in the workflow without any user input could be an option, but we do not know the exact dates and times that the IDs change, so it is possible that the automated approach could also use the wrong IDs. A possible intermediate approach could be to raise some kind of warning in the workflow generation process if we think that the IDs are inappropriate for the retro period being used. In this case, the user should still have the option to override the warning if they have reason to believe that they are using the correct IDs. Any other ideas as to how we can handle this?

### SPECIFIC CHANGES:

- Update `RRFS_UTILS` hash to incorporate updated code with GOES IDs included as namelist options.
- Changes to `./workflow/` to allow the user to set the GOES IDs in the `exp.*` files. If not specified, default values are used.
- Changes to `./scripts/exrrfs_nonvar_bufrobs.sh` to pass GOES IDs to the `RRFS_UTILS/larccld.fd` namelist.

## TESTS CONDUCTED: 

Ran two conus12km experiments on Ursa:

1. User specifies GOES IDs in `exp.conus12km`
2. User does not specify GOES IDs in `exp.conus12km`

In both experiments, the `nonvar_bufrobs` task runs successfully. When the GOES IDs are specified, those IDs were used in the `larccld.fd` namelist. When the GOES IDs are not specified, the default values were used (as expected).

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,1 +1,1 @@
-# License
+# License 
--- a/fix/conus3km/graphinfo/conus3km.graph.info.part.192
+++ b/fix/conus3km/graphinfo/conus3km.graph.info.part.192
@@ -1,0 +1,1 @@
+../../.agent/graphinfo/conus3km.20260310/conus3km.graph.info.part.192
--- a/fix/conus3km/graphinfo/conus3km.graph.info.part.300
+++ b/fix/conus3km/graphinfo/conus3km.graph.info.part.300
@@ -1,0 +1,1 @@
+../../.agent/graphinfo/conus3km.20260310/conus3km.graph.info.part.300
--- a/fix/conus3km/graphinfo/conus3km.graph.info.part.384
+++ b/fix/conus3km/graphinfo/conus3km.graph.info.part.384
@@ -1,0 +1,1 @@
+../../.agent/graphinfo/conus3km.20260310/conus3km.graph.info.part.384
--- a/fix/conus3km/graphinfo/conus3km.graph.info.part.640
+++ b/fix/conus3km/graphinfo/conus3km.graph.info.part.640
@@ -1,0 +1,1 @@
+../../.agent/graphinfo/conus3km.20260310/conus3km.graph.info.part.640
--- a/fix/physics/hrrrv5/ccn_activate.bin
+++ b/fix/physics/hrrrv5/ccn_activate.bin
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/ccn_activate.bin
--- a/fix/physics/hrrrv5/freeze_water_data_tempo_v3
+++ b/fix/physics/hrrrv5/freeze_water_data_tempo_v3
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/freeze_water_data_tempo_v3
--- a/fix/physics/hrrrv5/qr_acr_qg_data_tempo_v3
+++ b/fix/physics/hrrrv5/qr_acr_qg_data_tempo_v3
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/qr_acr_qg_data_tempo_v3
--- a/fix/physics/hrrrv5/qr_acr_qs_data_tempo_v3
+++ b/fix/physics/hrrrv5/qr_acr_qs_data_tempo_v3
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/qr_acr_qs_data_tempo_v3
--- a/fix/physics/mpas_global/ccn_activate.bin
+++ b/fix/physics/mpas_global/ccn_activate.bin
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/ccn_activate.bin
--- a/fix/physics/mpas_global/freeze_water_data_tempo_v3
+++ b/fix/physics/mpas_global/freeze_water_data_tempo_v3
@@ -1,0 +1,1 @@
+../../.agent/physics/hrrrv5/freeze_water_data_tempo_v3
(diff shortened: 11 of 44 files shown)


guoqing-noaa (approved):
LGTM. Thanks for the change, @ShawnMurdzek-NOAA

guoqing-noaa:
I have another thought on this. @MatthewPyle-NOAA Please hold my approval until further updates.

@ShawnMurdzek-NOAA Based on the following information,  I would suggest we decide the GOES IDs automatically in `scripts/exrrfs_nonvar_bufrobs.sh` so that users don't need to manually decide it in exp files. Thanks!
```
export NONVAR_GOES_WEST_ID=272  # For retros AFTER ~1/4/2023
#export NONVAR_GOES_WEST_ID=271  # For retros BEFORE ~1/4/2023
#export NONVAR_GOES_EAST_ID=273  # For retros AFTER ~4/7/2025
export NONVAR_GOES_EAST_ID=270  # For retros BEFORE ~4/7/2025
```
For example:
```
if (( CDATE > 2023010400 )); then
  satidgoeswest=272
else
  satidgoeswest=271
fi
```

guoqing-noaa (changes requested):
Suggest setting up GOES IDs automatically in exrrfs_novar_bufrobs.sh. The threshold dates can use the estimated ones (i.e. 1/4/2023 and 4/7/2025) for now. We can fine tune that once we get more accurate information.

ShawnMurdzek-NOAA:
@guoqing-noaa I can make that change if desired! Let me try to find more exact dates/times for when the GOES IDs switch so that we don't create issues for ourselves in the future with incorrect dates/times. In the meantime, I'm curious as to whether @hu5970 has any thoughts regarding your suggested change.

hu5970:
@ShawnMurdzek-NOAA How easy to confirm the GOES ID changes in the past several years? Auto check may not be a bad idea if we can make it easy and accurate.

ShawnMurdzek-NOAA:
@hu5970 I can confirm GOES ID changes by downloading NASA Langley BUFR files from HPSS and running the `larccld.fd` program a bunch of times. I wrote a quick script to do this, so it doesn't require much effort on my part.

While confirming the GOES IDs, I noticed that `RRFS_UTILS/larccld.fd/module_larccld_bufr_io.f90` explicitly accounts for some of the ID changes for GOES-West (e.g., lines [102-103](https://github.com/NOAA-GSL/RRFS_UTILS/blob/develop/larccld.fd/module_larccld_bufr_io.f90#L103)):

```
    if( (obs_time == east_time .and. satid==satidgoeseast ) .or.  &
        (obs_time == west_time .and. (satid==satidgoeswest .or. satid==259 .or. satid==271) ) ) then
```

and line [282](https://github.com/NOAA-GSL/RRFS_UTILS/blob/develop/larccld.fd/module_larccld_bufr_io.f90#L282):

```
      if(num_satid(i) == satidgoeswest .or. num_satid(i)==259 .or. num_satid(i)==271 ) then
```

Therefore, incorrectly specifying the ID for GOES-West as 272 when it should be 271 still results in the correct data being read. If we wanted, we could add more IDs to `module_larccld_bufr_io.f90` in the same manner so that the user does not need to specify the GOES ID at all. Note that this approach assumes that an ID used for GOES-West will never be used for GOES-East and vice versa (e.g., 271 will only ever be associated with GOES-West). If that is a bad assumption, we should stick to user-specified GOES IDs.

Would it be preferable to account for the GOES IDs in `module_larccld_bufr_io.f90`? If we did this, then we would not need information regarding when exactly the GOES IDs changed. Instead, we just need to know which IDs are associated with which satellite positions (GOES-East and GOES-West). It seems to me that this would be preferable.

guoqing-noaa (approved):
@ShawnMurdzek-NOAA Looks good. Thanks for the changes!

ShawnMurdzek-NOAA:
Thanks, @guoqing-noaa! One more comment: If possible, I'd like to enable strict checking of GOES IDs for retros by setting `stop_if_wrong_ids = 1` in `larccld.fd`. I would prefer to not have this enabled for our real-time runs, though. Is there an easy way to check whether a run is a retro or real-time within an `exrrfs_*` script? I didn't find an obvious environment variable in the `workflow/exp/exp.*` files that provides this information. No worries if this capability doesn't exist at the moment! This is more of a "wishlist" item at this point.

ShawnMurdzek-NOAA:
Updated code runs successfully on Ursa and the `nonvar_bufrobs` task successfully selects the correct satellite IDs for the retro period being used (which differ from the default values).

hu5970 (approved):

MatthewPyle-NOAA:
@ShawnMurdzek-NOAA Are you ready for this one to be merged?  Wasn't sure if you still wanted to tweak something.

ShawnMurdzek-NOAA:
@MatthewPyle-NOAA let's go ahead and merge this. Thanks!

MatthewPyle-NOAA (approved):

PR #1382, opened 2026-03-13
[rrfs-mpas-jedi]Reduce filesize of history and diag files to minimal if DO_POST=false

chunhuazhou:
## DESCRIPTION OF CHANGES: 
For GETKF cycling without ensemble forecast, we only need mpasout files for the cycling. To reduce the disk usage and simplify the GETKF cycling workflow, we can set DO_POST to false. In this case, the history and diag files are not needed. This PR reduces the entries to minimal in stream_list files for history and diagnostics, if DO_POST=false. The filesizes for conus3km can be reduced from 31G to 11K for history files, 2G to 11K fro diag files. Unless the MPAS-Model can run with history_interval being none, this is a good work-around for saving disk space, and also slightly reducing the run-time.

## TESTS CONDUCTED: 
Successfully tested in real time and retro tests.

### DIFF
--- a/scripts/exrrfs_fcst.sh
+++ b/scripts/exrrfs_fcst.sh
@@ -43,6 +43,14 @@ mkdir -p graphinfo stream_list
 ln -snf "${FIXrrfs}/${MESH_NAME}"/graphinfo/* graphinfo/
 ${cpreq} "${FIXrrfs}/stream_list/${PHYSICS_SUITE}"/* stream_list/
 
+# If not doing post-processing (DO_POST is not TRUE), meaning history and diag files are not needed.
+# In this case, We would only output minimal information to history and diag files, to save space
+#
+if [[ "${DO_POST}" != "TRUE" ]]; then
+  sed -i -E '/initial_time|xtime|Time/!d'  stream_list/stream_list.atmosphere.diagnostics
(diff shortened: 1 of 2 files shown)


hu5970 (approved):

scripts/exrrfs_fcst.sh
```diff
@@ -43,6 +43,14 @@ mkdir -p graphinfo stream_list
 ln -snf "${FIXrrfs}/${MESH_NAME}"/graphinfo/* graphinfo/
 ${cpreq} "${FIXrrfs}/stream_list/${PHYSICS_SUITE}"/* stream_list/
```
hu5970, line 45: Need to add DO_POST=${DO_POST:-"TRUE"} at the beginning of this script to keep deterministic and ensemble forecast  stream right.

guoqing-noaa:
@hu5970 I discussed with @chunhuazhou . This PR will be replaced by #1383 
It worked well in my tests.

guoqing-noaa:
To complete the discussion, `DO_POST=false` only means we don't do mpassit/upp jobs, but we may still need history.nc files for plotting/examination, etc. This PR implicitly changes this expected behavior.

chunhuazhou:
Closing this PR as @guoqing-noaa has a better solution in https://github.com/NOAA-EMC/rrfs-workflow/pull/1383

PR #1390, opened 2026-03-17
Add `CYCLEDEF_PROD_EXCLUDE` to allow excluding some cycles from cycledef_prod

guoqing-noaa:
The PR addresses the urgent RRFSv2X need to exclude `1 2 13 14` from the `cycledef_prod` so that we can drop four cycles to save computing resources and hence make sure the realtime runs complete successfully in time.


chunhuazhou (approved):
Thanks for the quick action!

Junjun-NOAA (approved):

MatthewPyle-NOAA (approved):

PR #1393, opened 2026-03-18
several updates for RRFSv2X

guoqing-noaa:
1. if no DA at cold start cycles, remove the `jedivar` or `getkf` tasks from those cycles. This is achieved by introducing a new `da_nocold` cycledef with an `exclude_hours` attribute.
2. update MPAS-Model to include a cmake hotfix and the capability to output at flexible time levels.
3. introduce `export MPASOUT_TIMELEVELS='0 1'` so that we only output mpasout files at 0h and 1h forecast hours.

PR #1396, opened 2026-03-20
[rrfs-mpas-jedi] Updates related to running nonvar cloud analysis in realtime

ShawnMurdzek-NOAA:
## DESCRIPTION OF CHANGES: 

- Remove `nonvar_bufrobs` dependency on lightning data for realtime runs. These data are not used by the nonvariational cloud analysis at the moment.
- Remove `nonvar_cldana` dependencies on the `nonvar_bufrobs` and `nonvar_reflobs` tasks for realtime runs. This makes the workflow more resilient because the `fcst` tasks depends on `nonvar_cldana`. If `nonvar_cldana` is stuck because `nonvar_bufrobs` or `nonvar_reflobs` fails, then the entire workflow becomes stuck.
- Allow `nonvar_cldana` task to use the same `da_nocold` cycledef as the JEDI tasks. The `nonvar_cldana` task depends on the JEDI task, so this avoids issues where `nonvar_cldana` depends on a JEDI task that does not exist.
- Check whether processed observation files exist prior to copying in `exrrfs_nonvar_cldana.sh`. This avoids issues where the `nonvar_cldana` task fails because input data is missing.
- Change `cpreq` to `cp` when copying output from `nonvar_cldana`. This also avoids the potential for this task crashing in the unlikely event that output files are not created.

Tagging @chunhuazhou and @hu5970 to see if there are any other changes that they think are needed here.

## TESTS CONDUCTED: 
- Can successfully create realtime and retro workflows on Ursa using the updated code. Retro workflow looks very similar to previous code versions, with the only change being that `nonvar_cldana` now uses `cycledefs="da_nocold"`. 
- Currently testing whether a retro runs successfully on Gaea C6

### DIFF
--- a/scripts/exrrfs_nonvar_cldana.sh
+++ b/scripts/exrrfs_nonvar_cldana.sh
@@ -31,8 +31,8 @@ HH=${CDATE:8:2}
 
 # Model background
 if [[ -r "${UMBRELLA_PREP_IC_DATA}/init.nc" ]]; then
-  echo "INFO: Skipping nonvar cloud analysis because this is a cold start"
-  echo "INFO: The 'cldfrac' field, which is required, is not available for cold starts"
+  echo "WARNING: Skipping nonvar cloud analysis because this is a cold start"
+  echo "WARNING: The 'cldfrac' field, which is required, is not available for cold starts"
   exit 0
 else
   initial_file=mpasout.nc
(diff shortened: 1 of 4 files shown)


hu5970 (approved):

ShawnMurdzek-NOAA:
Testing update: RRFS retro runs successfully for two conus12km cycles (one cold start, one warm start) with nonvariational cloud analysis enabled on Gaea C6.

scripts/exrrfs_nonvar_cldana.sh
```diff
@@ -45,10 +45,18 @@ nlevel=$(wc -l < "${zeta_levels}")
...
-${cpreq} "${COMOUT}/nonvar_bufrobs/${WGF}/NASALaRC_cloud4mpas.bin" .
-${cpreq} "${COMOUT}/nonvar_bufrobs/${WGF}/LightningInMPAS.dat" .
-${cpreq} "${COMOUT}/nonvar_bufrobs/${WGF}/mpas_metarcloud.bin" .
-${cpreq} "${COMOUT}/nonvar_reflobs/${WGF}/RefInGSI3D.dat" .
+all_obs=( "${COMOUT}/nonvar_bufrobs/${WGF}/NASALaRC_cloud4mpas.bin" 
+          "${COMOUT}/nonvar_bufrobs/${WGF}/LightningInMPAS.dat"
+          "${COMOUT}/nonvar_bufrobs/${WGF}/mpas_metarcloud.bin"
+          "${COMOUT}/nonvar_reflobs/${WGF}/RefInGSI3D.dat" )
+for ob in ${all_obs[@]}; do
+  if [[ -r ${ob} ]]; then
```
guoqing-noaa: Suggest using `-s` instead of `-r` to avoid zero size files

workflow/rocoto_funcs/nonvar_cldana.py
```diff
@@ -7,10 +7,17 @@
 
 def nonvar_cldana(xmlFile, expdir, do_ensemble=False, do_spinup=False):
     meta_id = 'nonvar_cldana'
+    nocoldda = os.getenv('COLDSTART_CYCS_DO_DA', 'TRUE').upper() == 'FALSE'
```
guoqing-noaa, line 10: Do we want to do nonvar cloud analysis in the spinup cycles? (it looks like a yes from what is in current setup_xml.py).

If yes, you may need to do a little bit further changes, such as `do_spinup=False` -> `spinup_mode=0`, similar to the changed made to jedivar in PR #1393

scripts/exrrfs_nonvar_cldana.sh
ShawnMurdzek-NOAA: Done. Thanks for the suggestion!

workflow/rocoto_funcs/nonvar_cldana.py
ShawnMurdzek-NOAA: I changed the logic in `nonvar_cldana.py` to match `jedivar.py`. Let me know if you'd like to see other changes.

MatthewPyle-NOAA (approved):

guoqing-noaa (approved):
LGTM. Thanks, @ShawnMurdzek-NOAA

PR #1417, opened 2026-04-01
Add radar reflectivity EnVar DA

hongli-wang:
This PR add radar reflectivity EnVar DA capability into RRFV2. The reflectivity EnVar DA runs after the hybrid DA in one jedivar task. Instead of BUMP which is designed for large scale correlation compared to grid mesh resolution, a diffusion approach is used for ensemble error covariance localization. It has been test on CONUS3km grid with 15km cutoff scales on Gaea C6. 

The existing parameter DO_RADAR_REF will control whether to turn on the reflectivity ioda  and DA tasks.  The default is false. 

A previously added parameter DO_RADAR_REF_2ND_PASS in this PR was removed because a new parameter ANALYSIS_VARIABLES added by another PR can help to specify analysis variables with/without reflectivity EnVar DA. 

Thanks Guoqing for improving JEDI EnVar yaml file generation, and adding other error handling functions in this PR. 

## DESCRIPTION OF CHANGES: 

-Add a JEDI EnVar yaml file
-Update exrrfs_jedivar.sh 
-Update related scripts so that it won't affect the hybrid DA when DO_RADAR_REF=TRURE. 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules
  - [X] Gaea C6

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [X] Others: run 3 cycles (04Z-06Z) for the May 08 2024 case.


guoqing-noaa:
@hongli-wang I will be able to test and review this PR next Monday. Sorry for the long wait.

workflow/rocoto_funcs/jedivar.py
```diff
@@ -33,6 +33,7 @@ def jedivar(xmlFile, expdir, spinup_mode=0):
         'YAML_GEN_METHOD': os.getenv('YAML_GEN_METHOD', '1'),
         'COLDSTART_CYCS_DO_DA': os.getenv('COLDSTART_CYCS_DO_DA', 'TRUE').upper(),
         'DO_RADAR_REF': os.getenv('DO_RADAR_REF', 'FALSE').upper(),
+        'DO_RADAR_REF_2nd_DA': os.getenv('DO_RADAR_REF_2nd_DA', 'FALSE').upper(),
```
guoqing-noaa: Suggest changing `DO_RADAR_REF_2nd_DA` to `DO_RADAR_REF_PASS2` or `DO_RADAR_REF_2ND_PASS`. Thanks!

ush/yaml_finalize
```diff
@@ -92,7 +92,8 @@ if ytype == "jedivar":
 
 # modify variable list if do_radar_ref
 do_radar_ref = (os.getenv("DO_RADAR_REF", "FASLE").upper() == "TRUE" and start_type != "cold")
-if not do_radar_ref and ytype == "jedivar":
+do_radar_ref_2nd_da = (os.getenv("DO_RADAR_REF_2nd_DA", "FASLE").upper() == "TRUE" and start_type != "cold")
+if (not do_radar_ref or do_radar_ref_2nd_da) and ytype == "jedivar":
```
guoqing-noaa: PR https://github.com/NOAA-EMC/rrfs-workflow/pull/1424 modified this file to meet both the all sky radiance DA and 2nd loop refl DA needs. I suggest we wait for 1424 to be merged first and then update here.

parm/jedivar.ref.diff.yaml
```diff
@@ -0,0 +1,132 @@
...
+  minimizer:
+    algorithm: DRPCG
+  iterations:
+  - geometry:
+      nml_file: ./namelist.atmosphere
+      streams_file: ./streams.atmosphere
+      deallocate non-da fields: true
+      interpolation type: unstructured
+    gradient norm reduction: 1e-3
+    ninner: 75
```
guoqing-noaa: only one outer loop needed?
Do we want to use 2 outer loops each with 50 iterations?

hongli-wang: No nonlinear obs operator is involved so that one outer loop should work well. In addition, David used this setup for his previous GSI EnVar ref DA.

workflow/rocoto_funcs/jedivar.py
hongli-wang: Either will work for me. Thanks.

ush/yaml_finalize
hongli-wang: I just rebased my code and updated the script accordingly.

workflow/rocoto_funcs/jedivar.py
hongli-wang: Updated the parameter to DO_RADAR_REF_2ND_PASS.

guoqing-noaa (approved):
LGTM. Thanks, @hongli-wang!

hu5970 (approved):

guoqing-noaa (changes requested):
Sorry, @hongli-wang 
We need to talk more about your last commit before moving forward,

hongli-wang:
> Sorry, @hongli-wang We need to talk more about your last commit before moving forward,

Sure, let's talk tomorrow.

scripts/exrrfs_jedivar.sh
```diff
@@ -130,13 +130,14 @@ export beginDate=""${CDATEm2:0:4}-${CDATEm2:4:2}-${CDATEm2:6:2}T${CDATEm2:8:2}:0
...
   1) # from ${PARMrrfs}
-    cp "${EXPDIR}/config/jedivar.yaml" jedivar.yaml
+    cp "${EXPDIR}/config/jedivar.yaml" jedivar.org.yaml
     cp "${EXPDIR}/config/convinfo" .
     cp "${EXPDIR}/config/satinfo" .
     cp "${USHrrfs}/hifiyaml4rrfs.py" .
     cp "${USHrrfs}/yamltools4rrfs.py" .
     cp "${USHrrfs}/yaml_finalize" .
-    ./yaml_finalize jedivar.yaml
+    export ANALYSIS_VARIABLES="5"
```
guoqing-noaa: Line 139: We cannot hardcode 5 analysis variables here.
All sky radiance DA will need 12 variables. 

For now, we can make a rule that we can only do reflectivity DA in the second pass for JEDVAR
and hence we use the `if da_radar_ref` switch to hardcode 5 variables for that situation only

hongli-wang: You can not use 12 analysis variables as if you are using Hybrid DA, and your static B doesn't account for those cloud variables et al.

guoqing-noaa (approved):
LGTM. Thanks a lot for the great work and great discussions, @hongli-wang!

MatthewPyle-NOAA (approved):

PR #1420, opened 2026-04-02
unified gsibec workaround working for fv3-jedi and mpas-jedi

Junjun-NOAA:
Following [PR546](https://github.com/NOAA-EMC/RDASApp/commits/develop/), this PR provides an unified version of gsibec workaround codes which work with both fv3-jedi and mpas-jedi.

Previously, fv3-jedi related workaround is maintained by RDASApp, while mpas-jedi related workaround is maintained by rrfs-workflow. From this PR, only one version of workaround is needed and it is maintained in RDASApp.


guoqing-noaa (changes requested):
@Junjun-NOAA 

No, we can use the same GSIBEC code.
But we have to keep all RRFSv2 workaround codes at the rrfs-workflow level.

SamuelDegelia-NOAA:
@guoqing-noaa Part of the motivation by using the unified gsibec workarounds is that we only need to maintain a single version of those codes while the PRs are being reviewed. So just curious why you want all of the workarounds to stay in rrfs-workflow? To me it makes sense to keep things in RDASApp since then we only have to maintain one set of codes and they can be used by both the mpas-jedi and fv3-jedi workflows. 

If there is a concern about the other workarounds in RDASApp, then I should note that those existing workarounds are either specific to fv3-jedi or MGBF (which does not affect other functionalities). Again, the goal here is to make things easier to maintain and easier to share between mpas-jedi and fv3-jedi workflows.

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for the discussions!
Here are my considerations:

In theory, workarounds should not exist at all. In practice, however, they are sometimes necessary to address short-term needs. That said, they should always be treated as temporary solutions—ideally lasting less than six months, and certainly no more than one year. Rather than investing effort in forcing all workarounds into RDASApp, our time would be better spent actively advancing the relevant PRs to eliminate these workarounds as early as possible. I am glad to see that https://github.com/GEOS-ESM/GSIbec/pull/92 has been merged yesterday. I hope we can work together to get the https://github.com/JCSDA-internal/saber/pull/1088 merged in about two weeks or so.  For @TingLei-NOAA's PR https://github.com/JCSDA-internal/saber/pull/1189, I hope we may target a one-month frame.


Workarounds are inherently temporary, and it is most effective to manage them at the top level of rrfs-workflow rather than within individual submodules. In rrfs-workflow.v2, we treat RDASApp as an official repository without its own workarounds. Since workarounds may come and go across different submodules, it is more natural and efficient to keep them organized in a centralized location rather than distributing them throughout each submodule.

Compared to the retro tests required to verify that workarounds function correctly, maintaining duplicate copies of workarounds is a minimal effort and can be safely ignored. We are willing to support this tiny amount of overhead.


As you noted, RDASApp contains other workarounds, many of which are valuable contributions to JEDI (side note: the best practice is to create at least a PR (or an issue at the minimum) to JCSDA repos for each workaround). However, each of these requires retro testing to ensure correctness. Given that MPAS-JEDI and FV3-JEDI have different priorities, we cannot guarantee timely testing of every RDASApp workaround. Automatically applying all RDASApp workarounds to rrfs.v2 is too risky. Additionally, when rrfs.v2 developers work on the JEDI code under rrfs-worklfow.v2, it is important to make sure they work with a clean version focused only on MPAS-JEDI-related changes, minimizing confusion and mental workload.

Different users may also want to test different workarounds. When managed at the workflow level, it is straightforward to enable or modify them. In contrast, implementing them within RDASApp requires additional steps—adding the workaround to RDASApp first, committing changes, and then going back to update rrfs-workflow accordingly. Given the temporary nature of workarounds, these extra steps are unnecessary and inefficient.


The currnt pratice does NO harm to RDASApp. We still use the same workaround code, we still need to test the workarounds separately in fv3jedi and rrfs-workflow.v2. There is no clear drawback or limitation of this practice. 
Thanks!

SamuelDegelia-NOAA:
I do not see how keeping workarounds in RDASApp instead of rrfs-workflow increases mental load for developers. The full set of workaround code (both fv3-jedi and mpas-jedi) is still present when cloning the repo, just organized under a different subdirectory.

RDASApp also has built-in ctests specifically for the purpose you describe. They provide a much simpler and more controlled environment for validating these workarounds before moving on to full retrospective testing. While retros are still necessary, relying on RDASApp first makes the testing process more efficient and reduces the risk of issues only being discovered later in the workflow.

At this point, though, I think we have both laid out our positions clearly, so I will leave it here.

PR #1426, opened 2026-04-07
slim down the history file size; default KEEPDATA to YES in exp files; remove rt_jet

guoqing-noaa:
1. Current history files output 143 variables while mpassit and upp only need 39 variables. 
This PR removes unnecessary variables from the `output` stream. After this change, the `history.nc` file size is reduced from `31G` to `13G` and we get identical `mpassit.nc` files.
2. Lots of users expressed the need to set `KEEPDATA` to `YES` in the example exp files to facilitate developments. Since now we have very effective `clean` task to manage the disk space consumption, this PR makes this change to meet that need.
3. remove exp/rt_jet as jet has been decommissioned
4. link stream_list files from hrrrv5 to mpas_global. Currently, all stream_list files are the same between hrrrv5 and mpas_global


chunhuazhou (approved):

Junjun-NOAA (approved):

MatthewPyle-NOAA (approved):

daviddowellNOAA:
@guoqing-noaa Has this PR been tested for our various CONUS real-time runs (including RRFSv2X and MPAS_RRFSA)?

guoqing-noaa:
@daviddowellNOAA 
The mpassit.nc files are identical before and after the PR. But I did not tested in realtime runs.
@chunhuazhou may do a test on this starting today?

daviddowellNOAA:
@guoqing-noaa The real-time MPAS_RRFSA is successfully using the changes from this PR.  Thanks for the smaller history files!

guoqing-noaa:
> @guoqing-noaa The real-time MPAS_RRFSA is successfully using the changes from this PR. Thanks for the smaller history files!

@daviddowellNOAA Thanks for the updates!
I would expect it will also reduce the running time as now we avoid the writing of about 2/3 (i.e 104) variables.
Did you notice that in MPAS_RRFSA? Thanks!

daviddowellNOAA:
@guoqing-noaa Yes indeed!  It seems we reduced MPAS run time by roughly 15%.

guoqing-noaa:
> @guoqing-noaa Yes indeed! It seems we reduced MPAS run time by roughly 15%.

Wow! Good to know we can save about 15%!!  Thanks for checking this, @daviddowellNOAA

PR #1422, opened 2026-04-02
Updating ungrib_rrfs.sh to directly handle RRFSA NA 3km domain

EricJames-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR updates the ush/ungrib_rrfs.sh script to bypass the intermediate regridding needed for RRFS-A NA 3-km files prior to recent WPS updates.  The ungrib task takes about 21 min for each RRFS-A output file, so be sure to use sufficient ungrib_lbc groups in the workflow.  This PR also updates to the latest WPS which can handle RRFS-A NA 3-km grid.

## TESTS CONDUCTED: 

The code was tested for a RRFSv2 workflow (3km CONUS) on Gaea C6.  

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

PR #1441, opened 2026-04-13
add a new option 'REGRID_RRFS_GRIB2' to allow regrid NA3km rotated-lat-lon to the 3km variation of grid 130 before ungrib

guoqing-noaa:
PR #1422 updated rrfs-workflow to be able to handle the full NA3km 9GB grib2 files directly on the rotated-lat-lon grid. This is a great step forward. 

However, currently, it takes 4+ minutes more to get each ICs/LBCs. Given the limited resources of the realtime RRFSv2X, MPAS_RRFSA runs, the previous regrid then ungrib method is still preferred. And we can continue optimizing the direct handling method.

The PR adds back the previous regrid and then ungrib method. It is now turned on/off by configuring the `REGRID_RRFS_GRIB2` option in the exp files. This way, we keep the directly reading method as well as the previous method to meet different needs.

### DIFF
--- a/ush/regrid_rrfs_grib2.sh
+++ b/ush/regrid_rrfs_grib2.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+# interplate RRFS NA3km grib2 files to the variation 130 grid at 3km
+#
+# shellcheck disable=SC2154,SC2153,SC2086
+
+budget_fields=":(WEASD|APCP|NCPCP|ACPCP|SNOD):"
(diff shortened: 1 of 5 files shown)


EricJames-NOAA (approved):
Looks good to me.  Thanks for the quick fix, @guoqing-noaa

ush/regrid_rrfs_grib2.sh
MatthewPyle-NOAA, line 1: A small detail for the documentation comment on line 2, but grid 130 is a 13 km grid (https://www.nco.ncep.noaa.gov/pmb/docs/on388/tableb.html#GRID130)

MatthewPyle-NOAA (commented):
One small documentation issue to clean up.  Also wanted to confirm that you want the default for the regridding to be false.  Thanks.

guoqing-noaa:
@MatthewPyle-NOAA Thanks a lot for noticing those mismatches.
I made changes accordingly.

This is a 3km variation of grid 130, not the original grid 130 (13km). I have added some clarifications hoping this may reduce confusions.

MatthewPyle-NOAA (approved):

PR #1421, opened 2026-04-02
[rrfs-nco] Final product changes for RRFSv1

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR contains the final product changes and additions for RRFSv1.  Originally these changes would have been included with a v1.1 incremental update, but given the several month delay to the implementation date it was decided to include them with the initial v1 implementation.  Here is a summary of the changes:

1. GRIB2 file reorganization
- Add 2DFLD GRIB2 file containing all 2D fields
- Remove duplicate records from PRSLEV and NATLEV files (leaving current 2D fields in NATLEV)
2. Add variables
- Freezing spray at 10 m above MSL
- U/V components of effective layer bulk shear
- Downdraft CAPE
- U/V components of Bunkers storm motion
- Sea level pressure (subhourly)
- Vertically integrated total dust (COLMD)
- Fire radiative power (FRADPOW)
- Total dust emissions (AEMFLX)
- Aerosol layer top (HGTMD)
- Aerosol layer bottom (HGTMD)
- Aerosol layer mass-weighted centroid (HGTMD)
3. Remove variables
- HGT:planetary boundary layer (Richardson number based diagnostic)
- Haines Index
4. Add RRFS equivalent versions of RAP smoke and HYSPLIT dust products - the RAP smoke and HYSPLIT dust products would be retired in favor of these RRFS products
- smoke (sfc, column-integrated) for grids 196 (HI), 198 (Alaska), 227 (CONUS)
- dust (sfc, column-integrated) for grid 227 (CONUS)
5. Other miscellaneous changes
- Correct labeling of brightness temperature from BRTMP to BRTEMP
- Use WMO GRIB2 ID for total condensate at hybrid levels

In addition to the changes in this PR, the following UPP fix files (in fix/upp) are updated/added (can't store fix files on GitHub):
postxconfig-NT-refs_f00.txt            
postxconfig-NT-refs_f01.txt             
postxconfig-NT-refs.txt                    
postxconfig-NT-rrfs_f00.txt              
postxconfig-NT-rrfs_f01.txt              
postxconfig-NT-rrfs_firewx_f00.txt  
postxconfig-NT-rrfs_firewx_f01.txt  
postxconfig-NT-rrfs_firewx.txt         
postxconfig-NT-rrfs_subh.txt           
postxconfig-NT-rrfs.txt                    
refs_postcntrl_f00.xml                      
refs_postcntrl_f01.xml
refs_postcntrl.xml
[...]

PR #1442, opened 2026-04-13
[rrfs-nco] Remove hourly max PRATE from subhourly output, fix issue with REFS f00 2DFLD file

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- This PR includes two small fixes that did not make it into #1421 in time before it got merged.  Sorry for not catching these sooner - there were quite a lot of last minute product changes to keep track of!
1. The subhourly GRIB2 output files contain one extra field that doesn't need to be there - hourly max PRATE.  It is included with the on-the-hour parameters because subh_fields.txt contains PRATE:surface, which grabs both PRATE parameters.  A simple fix for this is to modify the wgrib2 command in exrrfs_post.sh to remove the PRATE hourly max field (-not max).
2. The REFS 2DFLD GRIB2 file at f00 does not contain SPFH and HGT at hybrid level 1 - I accidentally added them to the wrong section of the UPP control files.  I have updated REFS f00 files here on Cactus (fix/upp directory):
```
/lfs/h2/emc/vpppg/noscrub/Benjamin.Blake/post_updates_20260326/fix/upp/postxconfig-NT-refs_f00.txt
/lfs/h2/emc/vpppg/noscrub/Benjamin.Blake/post_updates_20260326/fix/upp/refs_postcntrl_f00.xml
```
This error explains why the dew point processing step crashed at f00 with the 2DFLD file, so I reverted the recent change in exrrfs_post.sh to always use the 2DFLD file for the ensemble members.

## TESTS CONDUCTED: 

- I tested the wgrib2 functionality on the command line and it works as expected.  The subhourly output files will contain 156 records instead of 157 with this fix.
- I generated f00 REFS output and can confirm the 2DFLD file now contains SPFH and HGT at hybrid level 1.  The associated workflow changes have not been tested but they can be tested in EMC's parallel - I don't anticipate any issues.

### DIFF
--- a/scripts/exrrfs_post.sh
+++ b/scripts/exrrfs_post.sh
@@ -472,7 +472,7 @@ if [ -f PRSLEV.GrbF${post_fhr} ]; then
     if [ -e $fld2d_subh -a -e $tm15 -a -e $tm30 -a -e $tm45 ]
     then
       cat $tm45 $tm30 $tm15 $fld2d_subh > 2DFLD.GrbF${fhr}_subh
-      wgrib2 2DFLD.GrbF${fhr}_subh -set center 7 -grib $fld2d_subh_combo >> $pgmout 2>> errfile
+      wgrib2 2DFLD.GrbF${fhr}_subh -set center 7 -not max -grib $fld2d_subh_combo >> $pgmout 2>> errfile
     else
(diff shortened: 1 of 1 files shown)


MatthewPyle-NOAA:
Was able to test this in an EMC parallel, and it worked as expected.

MatthewPyle-NOAA (approved):

PR #1446, opened 2026-04-16
More granularity for the com/ clean

guoqing-noaa:
There are great needs to be able to configure different retention hourcycs for different task subdirectories under the `COM/` directory. For example, we would like to remove `lbc` and `fcst` directories sooner (e.g. a retention of 2 days), keep `upp` longer (e.g. 30 days) while a retention of 5 days for other subdirectories.

This PR allows users to set up a follow-style Python dictionary string for `COM_RETENTION_CYCS ` so as to customize different retention hourcycs for different task subdirectories.
For example:
```
export COM_RETENTION_CYCS='{"default": 120, "lbc,fcst": 48, "upp": 840}' 
```
At the same time, `export COM_RETENTION_CYCS=480` is still supported for a plain set up.

Note: Rocoto uses single quote to define and pass all environmental variables, hence we have to use double quotes for the dictionary keys while single quote for the outermost part.


guoqing-noaa:
The "XML comparison" test failed as expected since we removed `COM_LBC_RETENTION_CYCS ` and changed the `COM_RETENTION_CYCS` setting for RRFSv2X. I will go ahead to make a mark in `.ignore.compare_xml_outputs` to indicate that the changes here can be ignored safely. 

image

chunhuazhou (approved):

Junjun-NOAA (approved):

MatthewPyle-NOAA (approved):

PR #1456, opened 2026-04-22
[rrfs-mpas-jedi]Changes in preparation for SFE

chunhuazhou:
## DESCRIPTION OF CHANGES: 
In preparation for the Spring Forecast Experiments, we are finalizing the rrfs-workflow, including:

- Update RDASApp to use hash efc2bb3
- Update RDASApp build to use the workaround inside RDASApp and remove the workaround from the rrfs-workflow
- Update jedivar.yaml to use VertInterp as obs operator for uv287
- Change gsibec fix files

More changes will be added in this PR later.


hu5970 (approved):

MatthewPyle-NOAA (approved):

guoqing-noaa (approved):
First, no worries, I have approved this PR, :). I understand it is good to finalize a version for SFE2026 and thanks for all the changes!

Second, I would like to raise a concern about the hidden risks of the current fragile RDASApp workaround practice.

As shown in https://github.com/NOAA-EMC/RDASApp/issues/552 and https://github.com/NOAA-EMC/RDASApp/pull/555, copying/patching codes around silently poses risks (e.g. the new SABER functionalities are overwritten unexpectedly and the latest SP/IP updates are reverted without notice). We only became aware of these risks when I tried to review all changes through the standard Github PR review process. We were lucky that this time those overwrites might not impact us. But it remains a poential risk that some unintended changes may surface unpredictably and compromise our science results silently.

So it is still preferred to go through the standard Github PR review process for all changes inside the current workarounds. Some may worry about using "temporary" forks. However, nowadays that would NOT be a barrier or an issue. Both temporary workarounds and temporary forks are inherently short-lived, but the latter is a much more robust approach. It provides a clear, easy-to-track history and allows for broader visibility and review through the normal PR process.

hu5970:
@MatthewPyle-NOAA Could you merge this PR so we can deploy RRFSv2x based on this Hash. Thanks, Ming

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@MatthewPyle-NOAA I will merge this PR since GSL is eager to deploy RRFSv2x for HWT.

Issue #1462, opened 2026-04-23
[rrfs-mpas-jedi] Restart ensemble member forecasts from background instead of IC automatically in near-real-time cycled EnKF


AmandaBack-NOAA:
Due to GETKF quirkiness, the EnKF version RRFSv2X workflow for running in near-real-time should be able to automatically stage the background in place of the analysis as an initial condition and then re-run the forecast when member forecasts fail.

PR #1457, opened 2026-04-23
Update MPAS-Model to v8.3.1-2.26 and bugfix on clean.py

guoqing-noaa:
Update MPAS-Model to v8.3.1-2.26 and bugfix on clean.py

PR #1463, opened 2026-04-23
Automatically replace GETKF analysis and rerun forecast for members whose forecast has failed.

AmandaBack-NOAA:
This PR introduces the capability to automatically replace an ensemble member's analysis with the pre-GETKF background if one of the forecasts fails. It is intended only for the real-time system so it doesn't need a babysitter around the clock. For retros this is not recommended--the added tasks that only run if needed will prevent cycles from completing and hence later cycles from starting. In addition, tasks with a dependency on the fcst task will not automatically start if the fallback fcst runs instead (this doesn't affect real-time, as the only dependent task--save_for_next--uses a datadep rather than a taskdep). Note that workarounds for these limitations do exist in rocoto and can be added to rrfs-workflow if desired, but (my opinion) people running retros should probably be attentive to their failed tasks.

The PR also fixes 2 minor bugs that needed to be corrected to allow my testing:
 - clean.py currently tries to do math with a string; the string is now passed to the function in question wrapped in int()
 - ensmean upp has a taskdep on a task that does not exist. The taskdep has been revised to reflect the correct task name.

## DESCRIPTION OF CHANGES: 
- Added variable RESILIENT_RESTART which, when set to true, will add tasks to the enkf workflow to run only in the event that a forecast task dies.
- Tasks resilient_prep_ic_m## and resilient_fcst_m## are defined using the existing scripts for prep_ic and fcst to ensure consistency between the main tasks and the restart replacements. the script for prep_ic is lightly modified to be a metatask over members ONLY for the resilient restart version. Node and walltime for resilient_fcst are added to the sample workflows.
- sideload/clean.py corrected to stop doing math with strings (1-line change)
- rocoto_funcs/upp.py writes the correct ensmean mpassit task name when defining the ensmean upp task.

## TESTS CONDUCTED: 
Tested on Gaea with conus12km cycled ensembles
 - Confirmed the system is unaffected as long as the variable RESILIENT_RESTART is false or never defined
 - Confirmed that when RESILIENT_RESTART=TRUE the desired tasks are added and will run automatically when a forecast fails. Furthermore, the system runs normally when forecasts do not fail.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 
#1462


workflow/sideload/clean.py
guoqing-noaa, line 209: @AmandaBack-NOAA FYI, this has been fixed in PR https://github.com/NOAA-EMC/rrfs-workflow/pull/1457. Thanks!

workflow/exp/exp.ens_conus12km
```diff
@@ -54,6 +54,7 @@ export DET_RECENTERCYCS_DO_DA=false # whether or not do detministic DA at RECENT
 export NODES_IC="<nodes>1:ppn=40</nodes>"
 export NODES_LBC="<nodes>1:ppn=40</nodes>"
 export NODES_FCST="<nodes>1:ppn=40</nodes>"
+export NODES_RESILIENT_FCST="<nodes>1:ppn=40</nodes>"
```
guoqing-noaa: If we name the task with leading `fcst`, like `fcst_resilient_xxxx`, then we don't need separate `NODES_RESILIENT_FCST`. The `get_cascase_env` function will use `NODE_FCST` automatically.

scripts/exrrfs_prep_ic.sh
```diff
@@ -42,6 +42,11 @@ else
   mem_list=("000") # if determinitic
 fi
 
+if [[ "${RESILIENT_RESTART:-"FALSE"}" == "TRUE" ]]; then
```
guoqing-noaa: If `RESILIENT_RESTART` applies only to ensembles, it would be preferred to include `ENS` in the variable name.
`RESTART` has some kind of special meaning in the MPAS-based rrfs-workflow. We usually would interpret it as related to `restart.nc` or `config_do_restart` 

So it will be appreciated that we choose a different name for this variable.
For example: `FCST_ENS_RERUN`, `FCST_ENS_RESILIENT`, etc

scripts/exrrfs_prep_ic.sh
```diff
@@ -187,7 +192,11 @@ done # done for all the members
 # parallel run the serial tasks
 #
 ${cpreq} "${EXECrrfs}"/rank_run.x .
-${MPI_RUN_CMD} ./rank_run.x "${DATA}/script_prep_ic_*.sh"
+if [[ "${RESILIENT_RESTART:-"FALSE"}" == "TRUE" ]]; then
+  ${MPI_RUN_CMD} ./rank_run.x "${DATA}/script_prep_ic_${pid}.sh"
```
guoqing-noaa, line 196: So this is only to run one script?
If so, we don't need to do `${MPI_RUN_CMD} ./rank_run.x `.
We can run `${DATA}/script_prep_ic_${pid}.sh` directly

AmandaBack-NOAA: We can't run it directly unless we update its execute permissions, since the scripts this script creates only have r/w enabled. rank_run.x works around that and seems harmless to me.

workflow/exp/exp.ens_conus12km
AmandaBack-NOAA: Fixed

scripts/exrrfs_prep_ic.sh
AmandaBack-NOAA: Fixed to RESILIENT_ENSEMBLE. Thanks!

workflow/rocoto_funcs/prep_ic.py
```diff
@@ -52,6 +52,28 @@ def prep_ic(xmlFile, expdir, do_ensemble=False, spinup_mode=0):
...
     if "global" in os.getenv("MESH_NAME"):
         dcTaskEnv['cpreq'] = "ln -snf"
+    if not (do_ensemble and resilient):
+        metatask = False
+        meta_id = ""
+        meta_bgn = ""
+        meta_end = ""
+        ensindexstr = ""
+    else:
+        meta_id = 'resilient_prep_ic'
```
guoqing-noaa, line 66: lIne 62: suggest changing `meta_id` to `prep_ic_resilient` so that it can inherit any resource settings configured for `prep_ic`, such as `NODE_PREP_IC`, etc.

AmandaBack-NOAA: We don't need the nodes from NODE_PREP_IC since PREP_IC for ensemble runs several things simultaneously (1:ppn=40 while the rerun just needs 1:ppn=1, which it gets automatically in the current definition). However, I added a line to the exp.ens_conus* sample setup files to set the walltime for the rerun task to 10 minutes like the main PREP_IC requests.

guoqing-noaa: @AmandaBack-NOAA 
Thanks for mentioning that the resilient runs use different NODE settings. In this situation, we can define `NODE_PREP_IC_RESILEINT` explicitly in config_resources/config.base.

We have more resource definitions, such as 'WALLTIME', `ACCOUNT`, `QUEUE`, `PARTITION`, `RESERVATION`, `NATIVE`, etc, although they take default values most times.

But we have situations when prep_ic may configure some of the resources different from the defaults. 
Using a consistent cascading way (i.e `prep_ic_resilient`) can facilitate this task to inherit changes made to prep_ic. Thanks!

AmandaBack-NOAA: We don't want it to inherit RESERVATION and it shouldn't use a START_TIME from PREP_IC either. Probably best to evaluate these settings on a case-by-case basis instead of making a broad rule of inheriting.

guoqing-noaa: @AmandaBack-NOAA Thanks for the discussion. If possible, we would like to be consistent unless there are any unresolvable tech challenges. Most users will automatically classify `prep_ic` and its resilient counter part into the same category.

AmandaBack-NOAA: Thank you for raising the point that inheriting configurations will cause the rerun tasks to inherit reservations and other maybe unforeseen configuration elements. In light of this, I'm inclined to roll back the changes I made to the forecast resilient task that make it inherit from fcst in an uncontrolled way.

guoqing-noaa: @AmandaBack-NOAA Inheritance is not a must.
If you want to use different settings, you just define them explicitly.
For example,
`export FCST_RESILIENT=xxxxx`

guoqing-noaa: In you current settings, you don't use reservations at all, right?

Reservations are only used for realtime runs at the moment.

AmandaBack-NOAA: Kindly read over the description of the PR. In particular, "It is intended only for the real-time system so it doesn't need a babysitter around the clock. For retros this is not recommended"

guoqing-noaa: I had thought the "cascading" environmental resource variable setting rule has been documented in the wiki or the slides. But it turned out it was lost during some steps.

rrfs-workflow has intentionally adopted a "cascading" mechanism to avoid defining too many trivial environmental variables while still be able to customize at the greatest fine grains when needed. The `get_cascase_env(...)` was developed to handle this situation.

Let's use an extreme (not real) example to demo this. For this example, we have a task `ungrib_lbc_spinup_g01`:
We will first check if `NODE_UNGRIB_LBC_SPINUP_G01` is defined or not. If yes, use it.
If not, check if `NODE_UNGRIB_LBC_SPINUP` is defined or not. If yes, use it.
If not, check if `NODE_UNGRIB_LBC` is defined or not. If yes, use it.
If not, check if `NODE_UNGRIB` is defined or not. If yes, use it.
If not, check if `NODE` is defined or not. If yes, use it.

There are quite a few resource variables. 
Most time, a new variant `UNGRIB_LBC_xxxx` task will naturally inherits most of the `UNGRIB_LBC` resource settings so that we don't need to define a series of new resource variables just for this variant task. At the same time we can customize some special needs for this new variant task by defining them explicitly, such as `NODE_UNGRIB_LBC_xxx`, `RESEVATION_UNGRIB_LBC_xxx`.

In this PR, apparently, both the resilient `fcst` and `prep_ic` tasks are expected to share resources settings which can be shared with the original `fcst` and `prep_ic` tasks.  Thanks!

guoqing-noaa: > Kindly read over the description of the PR. In particular, "It is intended only for the real-time system so it doesn't need a babysitter around the clock. For retros this is not recommended"

For realtime runs on Ursa, you don't modify `exp.ens_conus12km‎` or `exp.ens_conus3km‎`, which are example files mostly for retros.  You would want to modify `exp.rrfsv2x_ens`.

AmandaBack-NOAA: The version of the PR originally submitted used the default resources for RESILIENT_PREP_IC and specified resources for RESILIENT_FCST. The specified resources for RESILIENT_FCST were removed at your request @guoqing-noaa  but as you've since pointed out there could be some issues arising from that.

guoqing-noaa: For realtime runs, we need more tests to make sure a rerun of ensemble forecasts can be done in time without a reservation.

AmandaBack-NOAA: That's an interesting opinion. Right now, the procedure as delineated by @hu5970 is that someone should notice that real-time runs have stopped, should find the member whose forecast failed, manually replace that member's GETKF analysis with the background, and then reboot the forecast task. This PR automates that procedure.

guoqing-noaa: > For realtime runs, we need more tests to make sure a rerun of ensemble forecasts can be done in time without a reservation.

I mean, fcst task may stay in a queue for a long time if no reservation and hence may not resolve the issue in time.
Manual rebooting uses the same reservation.

I may misunderstand. Do you want to use reservations or not for the fcst rerun? 
Also, if this PR is to address the realtime issue, we expect to modify the file `exp/rt_ursa/exp.rrfsv2x_ens`.

AmandaBack-NOAA: Ah. In that case it's right for the backup fcst task to inherit the reservation. Thanks!

guoqing-noaa:
@AmandaBack-NOAA Sorry that I was on leave early last week and missed any related discussions.
I think I understand the context more now. 
This is a good supplementary capability, but it is only used for RRFSv2X realtime runs in some specific situations (i.e., when we have frequent ensemble forecast crashes).

Retros don't need this, the operation will not need this either. 
So I would suggest we hold this PR for the moment.

Another possible alternate to address occasional ensemble forecast crashes may be done at the automatic monitoring level instead of modifying the workflow. We can monitor any dead ensemble forecasts and if it is a DA-related crash, we can re-stage the background and then reboot the associated task.

When I ran the realtime AR-PS (Prediction System), I had a monitoring script which detects any hanging fcst jobs and then scancel and reboot them (`/scratch4/BMC/zrtrr/gge/ARPS/PEAR/exp/rrfsdet/monitor_12hrly_workflow.sh`).  We may be able to do a similar thing here. Thanks!

MatthewPyle-NOAA:
Closing for now.  Can be reopened if needed.

PR #1464, opened 2026-04-28
Turn off deterministic WMO products for all cycles except 00/06/12/18Z

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- It was discovered in an email thread with NCO dataflow that we are creating deterministic RRFS WMO products for every cycle.  We can only generate these for the 00/06/12/18Z cycles due to SBN bandwidth limitations.  This PR adds an if condition to the product generation task which will turn off these products for all other cycles. 

## TESTS CONDUCTED: 

No tests conducted, but it is a simple one line change so I don't anticipate any issues.


MatthewPyle-NOAA (approved):
Thanks Ben!

PR #1466, opened 2026-04-28
allow different NETs for different systems, such as rtma, arps, etc

guoqing-noaa:
We have been developing 3D-RTMA and AR-PS (Atmospheric River Prediction System) based on rrfs-workflow.
For those systems, we will use different `NET` and `RUN` values based on the NCO implementation standards.

Some of rrfs-workflow parts already have the flexibility to handle this but some parts are hardcoded to `rrfs`.

This PR modifies hardcoded parts so that we can use different NETs for different systems.


guoqing-noaa:
The XML output differences are due to that the default `DO_BLENDING` value is changed to `false`.
```
-  <envar><name>DO_BLENDING</name><value>true</value></envar>
+  <envar><name>DO_BLENDING</name><value>false</value></envar>
```
These changes are expected.

guoqing-noaa:
Completed 24 cycles retro test and worked as expected.

chunhuazhou (approved):

Junjun-NOAA (approved):

spanNOAA (approved):

HaidaoLin-NOAA (approved):

MatthewPyle-NOAA (approved):

PR #1480, opened 2026-05-08
run graphics using metatasks based on tiles, allow cleaning nclprd/

guoqing-noaa:
This PR enable running graphics using metatasks based on tiles and hence the behavior is consistent between realtime and retro runs. It also adds `nclprd/` to the clean schedule for realtime runs.

### DIFF
--- a/workflow/config_resources/config.base
+++ b/workflow/config_resources/config.base
@@ -118,8 +118,8 @@ export NATIVE_UPP=${NATIVE_UPP:-""}
 
 # graphics
 export NODES_GRAPHICS=${NODES_GRAPHICS:-"<nodes>1:ppn=20</nodes>"}
-export WALLTIME_GRAPHICS=${WALLTIME_GRAPHICS:-"01:00:00"}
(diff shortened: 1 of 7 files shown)


guoqing-noaa:
The XML differences are expected:
```
-  <envar><name>COM_RETENTION_CYCS</name><value>{"default": 84, "lbc,fcst": 48, "upp": 840}</value></envar>
+  <envar><name>COM_RETENTION_CYCS</name><value>{"default": 84, "lbc,fcst": 48, "upp,nclprd": 840}</value></envar>
```

workflow/config_resources/config.base
MatthewPyle-NOAA, line 1: Does the comment list of domains need to be retained on line 122 now that it is in the GRAPHICS_TILES default definition?

Junjun-NOAA (approved):

MatthewPyle-NOAA (commented):
One small item to consider cleaning up

workflow/config_resources/config.base
guoqing-noaa: We can keep that comment since the default tiles may be changed later. Thanks!

MatthewPyle-NOAA (approved):

PR #1484, opened 2026-05-10
[rrfs-mpas-jedi]Rrfsv2x updates

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Add a save_for_next_f{hr}.done file and use this as data dependency in subsequent tasks (jedivar, prep_ic) to ensure the successful copying of mpasout files
Update rrfsv2x exp files to be consistent with the real-time runs

## TESTS CONDUCTED: 
Tested in retro and applied in rrfsv2x

### DIFF
--- a/scripts/exrrfs_fcst.sh
+++ b/scripts/exrrfs_fcst.sh
@@ -139,7 +139,7 @@ err_chk
 #
 # saving log.atmosphere.0000.out
 #
-[[ -f ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
+[[ -s ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
 #
 # double check status as sometimes atmosphere_model.x exit with 0 but there are still errors (log.atmosphere*err)
 #
--- a/scripts/exrrfs_fcst.sh
+++ b/scripts/exrrfs_fcst.sh
@@ -139,7 +139,7 @@ err_chk
 #
 # saving log.atmosphere.0000.out
 #
-[[ -s ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
+[[ -f ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
 #
 # double check status as sometimes atmosphere_model.x exit with 0 but there are still errors (log.atmosphere*err)
 #
--- a/scripts/exrrfs_save_for_next.sh
+++ b/scripts/exrrfs_save_for_next.sh
@@ -55,7 +55,7 @@ if (( err != 0 )) ; then
   echo "save_for_next failed with error code ${err} "
   err_exit
 else
-  touch "${COMOUT}/fcst/${WGF}"/fcst_f$(printf "%03d" "${fcst_hr}").done
+  touch "${COMOUT}/fcst/${WGF}/fcst_f$(printf "%03d" "${fcst_hr}").done"
 fi
 
 exit 0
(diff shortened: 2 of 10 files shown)


workflow/exp/rt_ursa/exp.rrfsv2x
```diff
@@ -42,20 +42,20 @@ export GSIBEC_Y=20
...
 export DO_CYC=true
 export CYC_INTERVAL=1
-for i in {0..23};    do arr[$i]="03"; done # 3hr fcst
-for i in {0..23..03}; do arr[$i]="12"; done # 18hr fcst
-for i in {0..23..12}; do arr[$i]="48"; done # 24hr fcst every 12hrs
-for i in 00; do arr[$i]="36"; done # 36hr fcst at 00Z
+for i in {0..23};    do arr[$i]="09"; done # 3hr fcst
+#for i in {0..23..03}; do arr[$i]="12"; done # 18hr fcst
+for i in {0..23..12}; do arr[$i]="72"; done # 24hr fcst every 12hrs
+#for i in 00; do arr[$i]="36"; done # 36hr fcst at 00Z
```
guoqing-noaa: Suggest removing the comment lines 46 and 48.
Also, either removing or updating the trailing comments in lines 45 and 47.

workflow/exp/rt_ursa/exp.rrfsv2x
```diff
@@ -97,7 +97,7 @@ export NODES_IODA_MRMS_REFL="<nodes>1:ppn=40</nodes>"
 export KEEPDATA=YES
 export MAXTRIES=3
 export STMP_RETENTION_CYCS=26
-export COM_RETENTION_CYCS='{"default": 84, "lbc,fcst": 48, "upp,nclprd": 840}'  # use " inside and ' outside
+export COM_RETENTION_CYCS='{"default": 96, "lbc,fcst": 72, "upp": 960}'  # use " inside and ' outside
```
guoqing-noaa: We need to purge the `nclprd` directories as well.
Either set it the same as upp, or set a separate retention period.

scripts/exrrfs_save_for_next.sh
```diff
@@ -54,6 +54,8 @@ export err=$?
 if (( err != 0 )) ; then
   echo "save_for_next failed with error code ${err} "
   err_exit
+else
+  touch "${COMOUT}/fcst/${WGF}/save_for_next_f${cyc_interval}.done"
```
guoqing-noaa: In retros or RTMA, we may not run our own ensembles and instead use ensembles already generated by other systems or other runs.
In these situations, the name `save_for_next` may not be descriptive enough. 
Suggest changing it to `fcst_*.done`

`cyc_interval` is fixed for any rrfs runs. Does it work for saving other forecast hours? Such as f03?
Mostly, we will not save `f100`, but it is still better to use 3 placeholders (i.e. f001) to be consistent with other parts of the workflow. Thanks!

scripts/exrrfs_fcst.sh
```diff
@@ -139,7 +139,7 @@ err_chk
 #
 # saving log.atmosphere.0000.out
 #
-[[ -s ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
+[[ -f ./log.atmosphere.0000.out ]] && cat ./log.atmosphere.0000.out
```
guoqing-noaa: why change from `-s` to `-f`?

chunhuazhou: Somehow when using "-s" we are not getting any information from log.atmosphere.0000.out; changing to "-f" seems to work better. Might be disk write delay.

guoqing-noaa: The only difference between `-s` and `-f` is that `-f` allows zero size file.
Here apparently, we don't need to do anything for zero size files.

guoqing-noaa (approved):
LGTM. Thanks for all the changes, @chunhuazhou

MatthewPyle-NOAA (approved):

PR #1486, opened 2026-05-13
hotfix - coldstart_only experiments should not have the 'da_nocold' cycledef

guoqing-noaa:
Previously cycledefs `da_nocold` were generated for experiments without any DA tasks. This PR corrects this bug.


guoqing-noaa:
The XML output differences are expected:  previously cycledefs `da_nocold` were generated for experiments without any DA tasks. This PR corrects this bug.

MatthewPyle-NOAA (approved):

PR #1512, opened 2026-05-28
Update ABI obsSpace in the super yaml file

HaidaoLin-NOAA:
## DESCRIPTION OF CHANGES: 

-Add useflag check to assimilate only 3 water vapor channels instead of all 10 channels and have the flexibility to assimilate more channels if needed
-Remove the channel number for cloudAmout as CloudAmount is not channel dependent in the ioda file. Further update in bufr2netcdf might be needed 
-Change max_exclusive: to min_exclusive for split window check
-Remove “max_exclusive: true” for BT StdDev check for channels 8-10
-Move the thinning filter after a few pre filters (e.g., cloud check and BT StdDev check)
-Add QC diagnostic flags
-Create a thinning priority variable based on cloudFree and BT StdDev from channel 13 (similar to GSI) and use this new variable as a priority variable for thinning
-Add “seed time” for temporal thinning
-Other miscellaneous changes (e.g., changed OmFNBC@DerivedMetaData to DerivedMetaData/OmFNBC and similar others; turned on a few parameters for thinning, added some comments etc.)

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [x] Gaea
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Following ABI QC in GSI, partially fixed the issue of the reduced assimilated number for ABI observations in JEDI compared with GSI
- Further updates/tuning might be needed 

## CONTRIBUTORS (optional): 

Xiaoyan Zhang @xyzemc


guoqing-noaa (approved):

parm/jedivar.yaml
```diff
@@ -18155,66 +18155,197 @@ cost function:
...
+         maxvalue: 1000.0
+         actions:
+         - name: set
+           flag: BTrangeCheck
+           ignore: rejected observations
+         #- name: reject
+         - name: reduce obs space
+
+       # ---------------------------------------------------------------------------------------------------
+       # Step 5 : Creat a varirable of thinning priority based on cloudFree and StdDev from ABI channel 13
```
keenaneure: Creat -> Create

parm/jedivar.yaml
```diff
@@ -18155,66 +18155,197 @@ cost function:
...
+         - name: CloudFractionCheck_channel_10
+           initial value: false
+           force reinitialization: false
+         - name: SplitWindowCheck_channel_10
+           initial value: false
+           force reinitialization: false
+         - name: ClearSkyStdCheck
+           initial value: false
+           force reinitialization: false
+         - name: ClearSkyStdCheck_channel_8
```
keenaneure: Curious - why are these diagnostic flags separate for each assimilated channel?

ShunLiu-NOAA:
@HaidaoLin-NOAA Do you have an update on RDASApp? We would prefer to prioritize updating RDASApp first.

parm/jedivar.yaml
HaidaoLin-NOAA: @keenaneure Good catch! I will update it. Thanks!

parm/jedivar.yaml
HaidaoLin-NOAA: @keenaneure Good question.  Thanks for asking.  I used those three diagnostic flags (ClearSkyStdCheck_channel_8, ClearSkyStdCheck_channel_9, ClearSkyStdCheck_channel_10_first) in my test/debug version to verify this filer in the final version: 

> filter: Bounds Check
         filter variables:
         - name: brightnessTemperature
           channels: *abi_g18_channels
         test variables:
         - name: ClearSkyStdDev/brightnessTemperature
           channels: 8-10
         flag all filter variables if any test variable is out of bounds: true
         maxvalue: 1.3
         #max_exclusive: true
         actions:
           - name: set
             flag: ClearSkyStdCheck
             ignore: rejected observations
           #- name: reject
           - name: reduce obs space

In this filter, the key point is "or" (if StdDev value larger than 1.3 from channel 8 or channel 9 or channel 10 for one location, all ABI channels at that location will be removed.  This line is important to insure this.  

>  flag all filter variables if any test variable is out of bounds: true

When I tested the above filter, I used three filters below (separated filters for each channel, these three filters are not in the final version) to verify the above filter.  The total number of the rejected observations from the above filter should be equal to the total number from the below three filters (adding three together).  The three flag variables were used to verify the correctness of the above filter.  These three flags were only used for verifying/debug purpose and are no longer used in the final version.  We can remove those three flags or just leave there for possible future usage.

>  - filter: Bounds Check
         filter variables:
         - name: brightnessTemperature
           channels: *abi_g16_channels
         test variables:
         - name: ClearSkyStdDev/brightnessTemperature
           flag all filter variables if any test variable is out of bounds: true
           channels: 8
         maxvalue: 1.3
         #max_exclusive: true
         actions:
           - name: set
             flag: ClearSkyStdCheck_channel_8
             ignore: rejected observations
           - name: reject
           #- name: reduce obs space
           #
       - filter: Bounds Check
         filter variables:
         - name: brightnessTemperature
           channels: *abi_g16_channels
         test variables:
         - name: ClearSkyStdDev/brightnessTemperature
           flag all filter variables if any test variable is out of bounds: true
           channels: 9
         maxvalue: 1.3
         #max_exclusive: true
         actions:
           - name: set
             flag: ClearSkyStdCheck_channel_9
             ignore: rejected observations
           - name: reject
       - filter: Bounds Check
         filter variables:
         - name: brightnessTemperature
           channels: *abi_g16_channels
         test variables:
         - name: ClearSkyStdDev/brightnessTemperature
           flag all filter variables if any test variable is out of bounds: true
           channels: 10
         maxvalue: 1.3
         #max_exclusive: true
         actions:
           - name: set
             flag: ClearSkyStdCheck_channel_10_first
             ignore: rejected observations
           - name: reject

HaidaoLin-NOAA:
> @HaidaoLin-NOAA Do you have an update on RDASApp? We would prefer to prioritize updating RDASApp first.
@ShunLiu-NOAA Not yet.  I will submit a PR to include the new ABI obsSpace yaml files (g16/g18) to the RDASApp. Thanks!

parm/jedivar.yaml
```diff
@@ -18155,66 +18155,197 @@ cost function:
...
+             options:
+               variables:
+               - name: MetaData/cloudFree
+               - name: ClearSkyStdDev/brightnessTemperature
+                 channels: 13
+               coefficients: [0.1, -10.0]
+               #coefficients: [10, -1.0]
+               constant_term: 100.0
+
+       # ---------------------------------------------------------------------------------------------------
```
xyzemc, line 18307: @HaidaoLin-NOAA Just want to understand how you define the coefficients and constant_term here?

parm/jedivar.yaml
```diff
@@ -18298,8 +18444,15 @@ cost function:
            flag all filter variables if any test variable is out of bounds: true
            minvalue: 0.0
            maxvalue: 0.5
+         actions:
+           - name: set
+             flag: GeoDomainCheck
+           - name: reject
```
xyzemc, line 18445: @HaidaoLin-NOAA Can we consider to move this regional domain check to ahead of thinning filter to increase the data passed thinning and QC in the future?

parm/jedivar.yaml
HaidaoLin-NOAA: @xyzemc The coefficients here is to decide the weighting of the cloudFree and StdDev.  Since the range of cloudFree is from 0-100, and the range of StdDev is 0-10 (or large), 0.1 is used for cloudFree and 10 is used  for StdDev for kind of comparable weighting.  This is consistent with GSI. `pred=10.0_r_kind-dataabi1(1,7)/10.0_r_kind+dataabi3(1,7)*10.0_r_kind`  I also tried different coefficients (e.g, putting more weighting to cloudFree or the other), there are difference but not large.  We could do some sensitivity study (including retro runs) to have a optimal coefficients in the future if needed.  I think the constant term is optional.  I have 100 here just wanted to make sure the thinning variable is a positive number.  But this is not important as JEDI can handle negative values (bigger, more priority, no matter positive or negative).

parm/jedivar.yaml
HaidaoLin-NOAA: @xyzemc Thanks. That's good point and may be worth further testing.  The online domain check removes outside domain observations, a lot of data are not relevant.  It may affect the boundary area a bit, but seems not large.  This is my test after before and after moving the regional domain check.
before (current setting):

> QC abi_g16 brightnessTemperature_8: 1807 passed out of 5695 observations.
QC abi_g16 brightnessTemperature_9: 1778 passed out of 5695 observations.
QC abi_g16 brightnessTemperature_10: 561 passed out of 5695 observations.

after (online domain check ahead of thinning filter):

> QC abi_g16 brightnessTemperature_8: 1820 passed out of 116927 observations.
QC abi_g16 brightnessTemperature_9: 1792 passed out of 116927 observations.
QC abi_g16 brightnessTemperature_10: 566 passed out of 116927 observations.

One other thing to think about is that online domain check is a post filter.  In the current ABI yaml file, we have pre filter, prior filter and post filter at the end.  If we want to move the online domain check ahead of thinning filter (which is a pre filter), then we have to put thinning filter under the post filter section.  If we put thinning filter under the post filter section, then it looks like we can not have the `reduce obs space` (JEDI would crash if we put `reduce obs space` under the post filter section).  After moving the online domain check ahead of thinning filter, it looks like we could increase a bit of assimilated observations (at least for the single I conducted).  But considering the lost of the `reduce obs space` function (which would result in a large jdiag file) and the small difference, maybe it would be good to keep the current setting.  But we can put this as an option for future testing/tuning.  Maybe we can use  the `reduce obs space` for post filters in JEDI in the future.  Then we can consider to put the online domain check ahead of thinning filter.  Please let me know if you have more comments/suggestions on this.  Thanks.

parm/jedivar.yaml
HaidaoLin-NOAA: I will remove two diagnostic flags (channel 8/9) to avoid possible confusing (if any) since they are no longer used. I will keep ClearSkyStdCheck_channel_10 since I use this for other filter  for channels 7, 10-16 (Step 15) in the final version and the previous flag ClearSkyStdCheck_channel_10_first in my testing version has been removed already.

parm/jedivar.yaml
xyzemc: Thanks for the explanation.

parm/jedivar.yaml
xyzemc: Good to know you already have the solid conclusion.  Let's just keep the on line domain check as current. Thanks for the detailed explanation.

xyzemc (approved):

ShunLiu-NOAA:
@HaidaoLin-NOAA If you can you have a ctest using RDASApp before merging this PR, it will help us stay on the same page for FV3-JEDI and MPAS-JEDI. Thanks.

HaidaoLin-NOAA:
> @HaidaoLin-NOAA If you can you have a ctest using RDASApp before merging this PR, it will help us stay on the same page for FV3-JEDI and MPAS-JEDI. Thanks.
Yes, I will submit a PR to the RDASApp and have a ctest there.  Thanks for the suggestion.

MatthewPyle-NOAA (approved):

PR #1523, opened 2026-06-06
bug fix for running coldstart_only experiments with history_interval and lbc_interval not equal to 1; update the pyDAmonitor hash

guoqing-noaa:
1. bug fix on smart_post_groups.py to accommodate history_inteval not equal to 1
2. bug fix on exrrfs_lbc.sh and exrrfs_ungrib.sh to correctly handle LBC_INTERVAL not equal to 1 and also make the relevant code snippet more readable (i.e. convert numbers to 10-based for easier references later).
3. RRFSv2x has been using the latest pyDAmonitor, update its hash in rrfs-workflow.

Tested with conus12km cycling DA experiments and also coldstart_only experiments with history_inteval and LBC_INTERVAL equal to 3 or 6.


guoqing-noaa:
The xml output differences are as follows for `conus12km_coldstart-only.xml`:
```
-<envar><name>CYC_INTERVAL</name><value>1</value></envar>
+<envar><name>CYC_INTERVAL</name><value>12</value></envar>
```
This is expected as previously `CYC_INTERVAL` is NOT used for coldstart_only experiments so it takes a default value of 1. But now we need to set a correct pseudo value to it in order to run the workflow correctly.

MatthewPyle-NOAA (approved):

PR #1535, opened 2026-06-11
[rrfs-mpas-jedi]v2X 0611

chunhuazhou:
## DESCRIPTION OF CHANGES: 
Several updates based on recent rrfsv2x update:
 - Update exp files to use high priority queue
 - Add the 4 missing cycles (01Z, 02Z, 13Z, 14Z) and turn on 24-hour cycling
 - Turn on surface cycling with the sst update
 - Update prep_ic and simplify its dependency, assuming surface update being done at cold start cycs
 - Update archive to also depend on graphics task
 - Fix a bug for graphics

## TESTS CONDUCTED: 
Implemented in rrfsv2x and being tested now.

PR #1540, opened 2026-06-12
minor fix - don't sys.exit after alerting users to clone pygraf

guoqing-noaa:
In PR #1535, it was found that the generated `rrfsv2x.xml` is not completed, missing the `archive` and `clean` tasks.
It was due to `sys.exit` after alerting users to clone pygraf.

This PR modifies this behavior so that CI tests can generate a full `rrfsv2x.xml`.

### DIFF
--- a/tests/.ignore.compare_xml_outputs
+++ b/tests/.ignore.compare_xml_outputs
@@ -1,2 +1,2 @@
-1535
+1540
 # put the exact PR number in the first line without any empty spaces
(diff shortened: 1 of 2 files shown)


chunhuazhou (approved):

Junjun-NOAA (approved):

spanNOAA (approved):

MatthewPyle-NOAA (approved):

guoqing-noaa:
The XML difference is because after the fix, we can correctly generate the archive and clean tasks for rrfsv2x. So it is expected.

Issue #1526, opened 2026-06-09
Unable to build utils, rank_run, and mpasblend on Hera


keenaneure:
I attempted a build that failed to create mpas_blending.x, mpassit.x, gen_ensmean_recenter.exe, process_NSSL_mosaic.exe, mpas_nonvarcldana.exe, process_larccld.exe, process_metarcld.exe, process_NSSL_mosaic_nonvar.exe after building. Based on log.build.mpasblend, Hera is detected as the machine, but Ursa is assigned as the 'target' variable in ~/sorc/MPASBlend/build.sh. Therefore, these builds fail relatively quickly. 

Hera directory:
/scratch4/BMC/wrfruc/eure/20260608_hera/rrfs-workflow/sorc/

guoqing-noaa:
@keenaneure Thanks for reporting this issue and it is fixed in PR #1555

PR #1555, opened 2026-06-17
hotfix to enable building MPASBlend and RRFS_UTILS on Hera

guoqing-noaa:
`/apps/slurm_hera` no longer exists after the latest Hera maintenance and hence MPASBlend as well as RRFS_UTILS fail to build on Hera as reported in #1526.

This is because both of the above two repositories used an old version script to detect HPC platforms, which relied on the existence of `/apps/slurm_hera` to distinguish hera and ursa.

This PR fixes this issue by using the more robust and fully tested `detect_machine.sh` from rrfs-workflow in these two repositories.

Tested on all supported RDHPCS and worked as expected.

### DIFF
--- a/workflow/sideload/archive_create_filelist.py
+++ b/workflow/sideload/archive_create_filelist.py
@@ -29,7 +29,7 @@ includes = []
 for pattern in items:
     if not pattern:
         continue
-    if f'/{WGF}' not in pattern:  # add WGF if missing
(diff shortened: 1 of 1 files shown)


ShunLiu-NOAA (approved):

keenaneure:
@guoqing-noaa I tested a build and I get 20 executables. Shouldn't there be 21 total?

guoqing-noaa:
> @guoqing-noaa I tested a build and I get 20 executables. Shouldn't there be 21 total?

20 is the correct number. Thanks!

Issue #1536, opened 2026-06-11
[rrfs-nco] Dependent Data for enkf_make_lbcs_*_mem*


JamesPolly-NCO:
[Requirements](https://nws-hpc-standards.readthedocs.io/en/latest/#standards)
A. If a job requires input data, but cannot access this data, the job should fail.
OR
B. The data used by the job is "data of opportunity" and this language should clearly be used in the warning message indicating that the data was inaccessible.

The enkf_make_lbcs_{00..11}_mem0{01..30} jobs all use the following data from $COMROOT/gefs:
`/lfs/h1/ops/prod/com/gefs/v12.3/gefs.${PDY}/18/atmos/pgrb2{a,b}p5/gep{01..30}.t18z.pgrb2{a,b}.0p50.f0{06..18}`

When **all** of these files cannot be found by the make_lbcs job, several fatal errors are produced, similar to:
```
*** FATAL ERROR: import_grib: /lfs/f1/ops/para/tmp/rrfs_enkf_make_lbcs_01_mem001_00.267221281.cbqs01/gep01.t18z.pgrb2b.0p50.f009 could not be opened ***
```

The above error and others can be found in the following output log (on cactus):
/lfs/h1/ops/test/output/20260608/rrfs_enkf_make_lbcs_00_mem001_00.o267221280

This log also shows messages similar to the following:
```
Missing /lfs/h1/ops/test/com/gefs/v12.3/gefs.20260603/18/atmos/pgrb2bp5//gep01.t18z.pgrb2b.0p50.f006 for forecast hour 6 !
Need to do time interpolation!
```

### Requests and Concerns:
1. Please confirm our understanding that missing **one** of the GEFS input files would result in the "Missing" message above, and that time interpolation would be performed, but this interpolation is not possible when **all** of the GEFS files are missing, and thus result in the failure captured in the output log above.
2. The "Missing" message above needs a "WARNING: " prefix added to it.
3. This is an ops serviceability concern. The current "FATAL" message shown above references a file in $DATA which originated from $COMROOT/gefs. It would be best if the fatal message indicated that original destination.
    - One way to do this is to modify the existing error messaging to mention the original location.
    - Another way to do this is to inject a new check in advance of where the code current fails. The check would simply see if any of these GEFS files can be found in $COMROOT/gefs and immediately FATAL (with helpful error messaging) if not.

MatthewPyle-NOAA:
[...]

PR #1556, opened 2026-06-18
[rrfs-nco] Add exception handling for missing enkf lbc input on GEFS data

lgannoaa:
Address the following:
1.  Please confirm our understanding that missing one of the GEFS input files would result in the "Missing" message above, and that time interpolation would be performed, but this interpolation is not possible when all of the GEFS files are missing, and thus result in the failure captured in the output log above.
Solution: The required 3 hours GEFS data are required. Application logic added for exception handling to FATAL fail the job if those file found missing.

2.  The "Missing" message above needs a "WARNING: " prefix added to it.
Solution: Modified the message to identify file that is not generated by operational GEFS.

3.  This is an ops serviceability concern. The current "FATAL" message shown above references a file in $DATA which originated from $COMROOT/gefs. It would be best if the fatal message indicated that original destination.
        One way to do this is to modify the existing error messaging to mention the original location.
        Another way to do this is to inject a new check in advance of where the code current fails. The check would simply see if any of these GEFS files can be found in $COMROOT/gefs and immediately FATAL (with helpful error messaging) if not.
Solution: Implement exception handling to display which GEFS data is missing from the operational GEFS location.

REFS #1536


MatthewPyle-NOAA (approved):
But also wondering if we could just get rid of the print (at line 417 now) to inform that data isn't available for that forecast hour.  Since that will always be true, does it need to be said?

ShunLiu-NOAA (approved):

JamesPolly-NCO:
Thanks for the work here. There is a discussion component of the original issue that doesn't necessarily involve code changes:
>Please confirm our understanding that missing one of the GEFS input files would result in the "Missing" message above, and that time interpolation would be performed, but this interpolation is not possible when all of the GEFS files are missing, and thus result in the failure captured in the output log above.

Is this understanding correct? Please confirm if so.

MatthewPyle-NOAA:
@lgannoaa Can you answer the question from @JamesPolly-NCO ?

MatthewPyle-NOAA:
Based on discussion at a tagup on 7/7, any missing GEFS files will lead to a graceful exit.  So sounds like concerns from @JamesPolly-NCO from the issue are resolved.

Issue #1584, opened 2026-07-08
Silent failure in ensf_prdgen jobs


LinGan-NOAA:
The NCO parallel job rrfs_ensf_prdgen_mem005_f052_06.o273870074 indicated the following silent failure:
This job reported completed. However, the following FATAL error were found:

FATAL ERROR: rd_grib2_msg_seq_file for /lfs/h1/ops/para/com/rrfs/v1.0/rrfsens.20260708/06/m005/rrfs.t06z.m005.2dfld.3km.f052.na.grib2, read outside of file, bad grib file ***
FATAL ERROR: rd_grib2_msg_seq_file for /lfs/h1/ops/para/com/rrfs/v1.0/rrfsens.20260708/06/m005/rrfs.t06z.m005.prslev.3km.f052.na.grib2, read outside of file, bad grib file ***
FATAL ERROR: rd_grib2_msg_seq_file for /lfs/h1/ops/para/com/rrfs/v1.0/rrfsens.20260708/06/m005/rrfs.t06z.m005.prslev.3km.f052.na.grib2, read outside of file, bad grib file ***
FATAL ERROR: rd_grib2_msg_seq_file for /lfs/h1/ops/para/com/rrfs/v1.0/rrfsens.20260708/06/m005/rrfs.t06z.m005.2dfld.3km.f052.na.grib2, read outside of file, bad grib file ***

Root cause is this job need exception handling logic to catch and handle wgrib2 error.
@BenjaminBlake-NOAA , would you please take a look and implement exception handling capability in associated jobs?
Thanks

BenjaminBlake-NOAA:
@lgannoaa In the post job, I see the following logic with wgrib2 commands:

  wgrib2 2DFLD.GrbF${post_fhr} -set center 7 -grib ${fld2d} **>>$pgmout 2>>errfile**

To address this issue, would adding similar logic (>> $pgmout 2>>errfile) to the prdgen task whenever wgrib2 commands are used be sufficient?

LinGan-NOAA:
> [@lgannoaa](https://github.com/lgannoaa) In the post job, I see the following logic with wgrib2 commands:
> 
> wgrib2 2DFLD.GrbF${post_fhr} -set center 7 -grib ${fld2d} **>>$pgmout 2>>errfile**
> 
> To address this issue, would adding similar logic (>> $pgmout 2>>errfile) to the prdgen task whenever wgrib2 commands are used be sufficient?

[...]

PR #1593, opened 2026-07-10
[rrfs-nco] Add exception handling to prdgen job to prevent wgrib2 errors

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 

- One of the RRFS ensemble forecast prdgen jobs experienced silent failures that were related to wgrib2, resulting in bad GRIB2 output files.  In this PR, exception handling is added to several critical wgrib2 commands within the prdgen job in order to hopefully prevent similar errors from occurring in the future.

## TESTS CONDUCTED: 

No tests have been conducted with these changes, but I don't anticipate any issues with implementing them in the workflow since the same script logic is already utilized with wgrib2 commands in exrrfs_post.sh.

## ISSUE: 

- Fixes issue #1584

### DIFF
--- a/ush/prdgen/rrfs_prdgen_subpiece.sh
+++ b/ush/prdgen/rrfs_prdgen_subpiece.sh
@@ -46,7 +46,7 @@ fi
 
 # Use different parm file for each subpiece
 if [[ "$inputfile" =~ "prslev" ]]; then
-  wgrib2 $comout/${inputfile} | grep -F -f ${parmfile} | wgrib2 -i -grib inputs.grib${domain} $comout/${inputfile}
+  wgrib2 $comout/${inputfile} | grep -F -f ${parmfile} | wgrib2 -i -grib inputs.grib${domain} $comout/${inputfile} >>$pgmout 2>>errfile
   wgrib2 inputs.grib${domain} -new_grid_vectors "UGRD:VGRD:USTM:VSTM" -submsg_uv inputs.grib${domain}.uv
 else
   wgrib2 $comout/${inputfile} -new_grid_vectors "UGRD:VGRD:USTM:VSTM" -submsg_uv inputs.grib${domain}.uv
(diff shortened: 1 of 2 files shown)


WeiWei-NCO:
Please correct me if I am wrong, but it looks like the changes are to redirect those error messages, and the job would still complete?

LinGan-NOAA:
We need export err=$?; err_chk to be in place to hand the error.

BenjaminBlake-NOAA:
You are right, I forgot to include it.  Will get it added now.

MatthewPyle-NOAA (dismissed):

WeiWei-NCO:
Do we need something like `set -o pipefail` for pipe?

MatthewPyle-NOAA:
@WeiWei-NCO I wasn't familiar with that pipefail functionality, but seems like it might be useful here.

BenjaminBlake-NOAA:
@WeiWei-NCO I was also not aware of set -o pipefail.  Would you recommend adding this functionality to the beginning of exrrfs_prdgen.sh?

WeiWei-NCO:
It depends on whether we want `$?` to reflect any failure in the pipeline or not, and it only matters when there's error handling right after it. We've seen many cases when err_chk doesn't capture the failure from pipeline or err_chk kills the job unintentionally because of unnecessary pipefail.

In exrrfs_prdgen.sh, the only other pipe is `net4=$(echo ${NET:0:4} | tr '[:upper:]' '[:lower:]')` and there's no err_chk after it. So I think we can add it to the beginning of exrrfs_prdgen.sh.

ush/prdgen/rrfs_prdgen_subpiece.sh
```diff
@@ -46,7 +46,8 @@ fi
 
 # Use different parm file for each subpiece
 if [[ "$inputfile" =~ "prslev" ]]; then
-  wgrib2 $comout/${inputfile} | grep -F -f ${parmfile} | wgrib2 -i -grib inputs.grib${domain} $comout/${inputfile}
+  wgrib2 $comout/${inputfile} | grep -F -f ${parmfile} | wgrib2 -i -grib inputs.grib${domain} $comout/${inputfile} >>$pgmout 2>>errfile
+  export err=$?; err_chk
```
WeiWei-NCO, line 52: I don't remember if `set -o pipefail` inherits down to child scripts. I have the impression that it doesn't. If not, we need it here too. Can be added right above the pipe and immediately turned off after err_chk, to make less imapcts.

BenjaminBlake-NOAA: Sure I can do that.  I'll add `set -o pipefail` just before this section and `set +o pipefail` right after err_chk.  If I need to update any of the syntax let me know since I've never used this function before!

WeiWei-NCO (approved):
Looks good. Thank you!

MatthewPyle-NOAA (approved):

PR #1610, opened 2026-07-21
[rrfs-mpas-jedi] Addition of TMS S02 observation space

keenaneure:
This PR adds the TMS observation space in jedivar.yaml and getkf.yaml. Credit goes to JJ Guerrette and Ryan Honeyager for providing the majority of this observation space. 

Summary of observation space:
- Create diagnostic QC flags
- Gaussian thinning (121-km geodesic)
- PreQC check (bad data before DA is performed)
- Online regional domain check
- Calculate symmetric cloud impact (from both model and observation space)
- Calculate clear-sky BTs 
- Calculate ice and liquid scattering indices
- Mean cloud impact/limits to do mismatched cloud impact check
- Land type filter
- Use flag check

TMS uses the same bias correction predictors as the ATMS instrument. This YAML works as an all-sky configuration


keenaneure:
Not entirely sure why the compare YAML outputs check is failing after checking echo diffs.

parm/getkf.yaml
```diff
@@ -22010,3 +22010,1389 @@ observations:
...
+                 - 0.75
+                 - 0.3
+                 - 0.5
+                 - 0.5
+                 - 0.25
+                 - 0.25
+                 - 1.0
+                 - 1.0
+                 - 1.0
+                 - 1.0
```
guoqing-noaa: Lines 22268-22280: Please use more reader-friendly format for the list, i.e.
```
x0: [1.0, 1.0, 0.75, 0.3, 0.5, 0.5, 0.25, 0.25, 1.0, 1.0 1.0 1.0]
```
Change other lists similarly.

xyzemc (approved):

parm/jedivar.yaml
```diff
@@ -22074,3 +22074,1257 @@ cost function:
...
+             obsfile: jdiag_tms_s02.nc
+         simulated variables: [brightnessTemperature]
+         observed variables: [brightnessTemperature]
+         channels: 1-12
+       obs operator:
+         name: CRTM
+         Absorbers: [H2O,O3,CO2]
+         Clouds: [Water, Ice]
+         Cloud_Fraction: 1.0
+         UseQCFlagsToSkipHofX: true
```
HaidaoLin-NOAA, line 23076: @keenaneure Could you please let me know why we need this (UseQCFlagsToSkipHofX: true) for tms data? If not using, what happens?

parm/jedivar.yaml
```diff
@@ -22074,3 +22074,1257 @@ cost function:
...
+               name: ObsFunction/Arithmetic
+               channels: 1-12
+               options:
+                 variables:
+                 - name: HofX/brightnessTemperature
+                   channels: 1-12
+                 coefs:
+                 - 1.0
+
+         # Ch 1 Clear-sky Tb check (overwrites default value for each ch after doing CRTM check)
```
HaidaoLin-NOAA, line 23294: @keenaneure  For clear-sky Tb check, starting here to line 22566,  if this is not channel dependent (it looks not?), will it be possible to combine these lines into a single filter but applied for all 12 channels instead of having one filer for each channel? Thanks!

parm/getkf.yaml
```diff
@@ -22010,3 +22010,1257 @@ observations:
...
+           horizontal_mesh: 121
+           use_reduced_horizontal_grid: true
+           distance_norm: geodesic
+           actions:
+           - name: set
+             flag: Thinning
+           - name: reduce obs space
+
+       obs post filters:
+       # Use PreQC filter to remove bad data
```
HaidaoLin-NOAA, line 23099: For this filter (Use PreQC filter to remove bad data), does that mean if PreQC/brightnessTemperature ==0, then this observation will be rejected?  Usually, if QC related flag == 0, that means the data quality would be good, but this may not be all the cases (i.e., different here for tms data).  Just a double check. What are the typical values of PreQC/brightnessTemperature for tms data?  Other comment is that if PreQC/brightnessTemperature comes from the observation ioda file, then maybe this filter could be moved to the pre filter section?

parm/jedivar.yaml
```diff
@@ -22074,3 +22074,1257 @@ cost function:
...
+           - variable:
+               name: GeoVaLs/water_area_fraction
+             minvalue: 0.99
+           actions:
+           - name: set
+             flag: LandtypeCheck
+             ignore: rejected observations
+           - name: reject
+
+         # Reject based on Q20 (Ch 1)
```
HaidaoLin-NOAA, line 23728: This comment applies to Q20 and Q4M from line 22750-23188, will it be possible to combine all 12 channels into one filter (i.e, one filter for all 12 channels for Q20 +Q2M, one filter for all 12 channels for Q4M+Q40) if the same threshold and logic are used for all channels?

parm/jedivar.yaml
keenaneure: I can comment this out. It may still be good to have HofX values for kicked obs too.

parm/jedivar.yaml
keenaneure: I am not honestly sure why it is repeated individually for each channel after the initial calculation for all 12 channels simultaneously. What I can do here is test how/if the values change without explicit, individual calculations for each channel.

parm/jedivar.yaml
keenaneure: I do not think there is an easy way to combine this for all 12 channels since Q20, Q2M, Q4M, and Q40 all can be different values unfortunately. I may be wrong though.

guoqing-noaa:
@HaidaoLin-NOAA @keenaneure You may run `yj split2 jedivar.yaml` and then compare those TMS filters one by one to see whether it is possible to merge them. Thanks!

parm/jedivar.yaml
HaidaoLin-NOAA: Sorry, I don't mean you need to comment it. I just curious why we need this here.  If without it, is there any difference?  If there is a good reason to keep it and we will keep it.

keenaneure: My best guess is it does not compute HofX for obs rejected. Documentation mentions this for RTTOV but I would imagine it applies to the CRTM as well. I think we can leave this out for now.

parm/getkf.yaml
keenaneure: This filter should pass anything that has PreQC = 0. I can move it to the pre filter section. I have not seen many nonzero values here thus far.

guoqing-noaa: Is TMS provided as ioda files already and we don't need any IODA converter, right?

Will the tms ioda files always contain `PreQC`?

keenaneure: I use the [tropics_2ioda.py](https://github.com/JCSDA-internal/ioda-converters/blob/e9c233d3ce46fae54629da75581369a47d0d6c1d/src/hdf5/tropics_2ioda.py) converter with a modification to include clearFraction in the conversion (for clear-sky screening testing). The raw file are netCDF but not IODA format. However, the granules need to be binned to hourly files, which I have been doing outside of the workflow. They will always have PreQC in them.

parm/jedivar.yaml
HaidaoLin-NOAA: I C. Thanks!

parm/getkf.yaml
HaidaoLin-NOAA: > This filter should pass anything that has PreQC = 0. I can move it to the pre filter section. I have not seen many nonzero values here thus far.

I C. Thanks!

HaidaoLin-NOAA:
> @HaidaoLin-NOAA @keenaneure You may run `yj split2 jedivar.yaml` and then compare those TMS filters one by one to see whether it is possible to merge them. Thanks!

@guoqing-noaa Thanks!

MatthewPyle-NOAA:
This one seems to have stalled - still be working on? @keenaneure

keenaneure:
> This one seems to have stalled - still be working on? @keenaneure

Yes, sorry I have been at the AMS Madison Summit but I will contunue working on this.

keenaneure:
> yj split2 jedivar.yaml

@guoqing-noaa It creates individual filter YAML files for each of the filters as the yamltool documentation states. I am confused - how does this inform how we should move forward with merging these filters?

guoqing-noaa:
@keenaneure We may close this PR and then after PR #1637 is merged, let's create a new PR to add the `TMS S02` observer. We will follow the [`RRFSv2 YAML convention`](https://github.com/NOAA-EMC/rrfs-workflow/wiki/RRFSv2-YAML-convention) at that time. No worries, I will help with the whole process. At the same time, feel free to address any remaining review comments, if any. Thanks!

keenaneure:
Sure, I can redo this following that marge. I am still unclear on how to address the remaining comments on this ob space with yj split2.

guoqing-noaa:
> Sure, I can redo this following that marge. I am still unclear on how to address the remaining comments on this ob space with yj split2.

@keenaneure Sounds good. Don't worry about `yj split2`. I suggested that becuase it looks like comparing different filters one by one may be helpful in fine tuning TMS observers. But it is NOT a must. We can talk more offline. Thanks!

guoqing-noaa:
@keenaneure Let's close this PR and recreate a new PR based on the latest `rrfs-mpas-jedi`. Thanks!

PR #1620, opened 2026-07-28
[release/v1.0] define FHR for rrfs_enkf_save_restart_ensinit jobs

ShunLiu-NOAA:
## DESCRIPTION OF CHANGES: 

- define FHR 000 for rrfs_enkf_save_restart_ensinit jobs in rrfs_nco_para.def
- add FHR to ecf script

### Machines/Platforms:

- WCOSS2
  - [X] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [X] Engineering tests
  - [X] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- Fixes the issue(s) mentioned in #9999


ecf/scripts/enkf/forecast/jrrfs_enkf_save_restart_ensinit_master.ecf
```diff
@@ -32,6 +32,7 @@ module load nco/${nco_ver}
 
 export WGF="%WGF%"
 export MEMBER_NAME="%MEMBER_NAME%"
+export FHR="%FHR%"
```
WeiWei-NCO: Is it always 000? Might be easier to hardwire here instead of updating many place in ecflow.

ShunLiu-NOAA: Yes. I considered hardwiring this directly in the job card. However, to maintain consistency with how other save_restart processes are handled, I modified the ecFlow instead. Personally, I would prefer to keep it simple and just hardwire it in the job card.

WeiWei-NCO: Got it! Yes, both work just fine.

ecf/scripts/enkf/forecast/jrrfs_enkf_save_restart_ensinit_master.ecf
```diff
@@ -32,6 +32,7 @@ module load nco/${nco_ver}
 
 export WGF="%WGF%"
 export MEMBER_NAME="%MEMBER_NAME%"
+exprot FHR+"000"
```
WeiWei-NCO: Now we have a typo...

ShunLiu-NOAA: Thanks. It is fixed.

lgannoaa (changes requested):
FHR in ecflow should not be hardwired. In this case, there are 30 members. Each member has a save spinup and ensinit jobs. The spinup has FHR coded as 001 from the ecflow definition file. The ensinit jobs should follow the same design pattern.

In addition, [line 50 in ex-script](https://github.com/NOAA-EMC/rrfs-workflow/blob/2ae4e1935cad2667e143765d46524d8145b33de1/scripts/exrrfs_save_restart.sh#L50) has a bug that introduce unassigned fhr error. Recommend to fix it.

ShunLiu-NOAA:
> FHR in ecflow should not be hardwired. In this case, there are 30 members. Each member has a save spinup and ensinit jobs. The spinup has FHR coded as 001 from the ecflow definition file. The ensinit jobs should follow the same design pattern.
> 
> In addition, [line 50 in ex-script](https://github.com/NOAA-EMC/rrfs-workflow/blob/2ae4e1935cad2667e143765d46524d8145b33de1/scripts/exrrfs_save_restart.sh#L50) has a bug that introduce unassigned fhr error. Recommend to fix it.

The PR is to fix the bug in line 50 by defining FHR through job card.

LinGan-NOAA:
> The PR is to fix the bug in line 50 by defining FHR through job card.
The line 50 needs to be fixed with a default value.
The ecflow script should use value set by the definition file instead of hard code in the master file.

MatthewPyle-NOAA:
Based on discussions, an alternate solution focused just on the ex-script may be provided from @WeiWei-NCO

WeiWei-NCO:
To my understanding, `enkf_forecast_ensinit` and `enkf_save_restart_ensinit` only exist in the 07z and 19z spinup cycles, and they don't run in normal scenario. 

While handling the system issue on 07/25, 07z `enkf_forecast_ensinit` kicked off, but `enkf_save_restart_ensinit` failed to copy over restart files, which then failed 07z `enkf_prep_cyc_spinup`. The root cause was `FHR` being undefined in those `enkf_save_restart_ensinit` jobs.

After discussing with @lgannoaa , we decided to add a default value for `FHR` in `exrrfs_save_restart.sh`:
Change  `fhr=${FHR:-}` to `fhr=${FHR:-000}`

This change in now running in NCO v1.0.22.

MatthewPyle-NOAA:
@WeiWei-NCO Will you be opening a fresh PR with this change?

WeiWei-NCO:
Yep. #1630 opened

MatthewPyle-NOAA:
Issue was addressed in #1630

PR #1626, opened 2026-07-31
[rrfs-nco] remove email section from exrrfs_analysis_gsi.sh

JamesPolly-NCO:
## DESCRIPTION OF CHANGES: 

- analysis jobs are sending redundant emails that will be sent by other jobs.

## TESTS CONDUCTED: 

- no testing; removed email action

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

### DIFF
--- a/scripts/exrrfs_analysis_gsi.sh
+++ b/scripts/exrrfs_analysis_gsi.sh
@@ -617,20 +617,8 @@ done
 echo "number of obs files:" $obs_file_count
 if [ "$obs_file_count" -eq 0 ]; then
   err_exit "FATAL ERROR: no observation files were found in this analysis."
-elif [[ "$missing_data_flag" == true ]]; then
-  cat << EOF > email_warn.txt
-WARNING: Some observation files were unable to be found in this analysis:
-
-RRFS cycle will continue without these files.
-
-No immediate impact to integrity of delivered products.
-
-Ops action: identify cause of missing files at the following path(s):
-EOF
-  cat missing_file_list >> email_warn.txt
-  mail.py -s "Missing Observation Data of Opportunity for RRFS" -v ${MAIL_TO} < email_warn.txt
 fi
-#
+
 #-----------------------------------------------------------------------
 #
 # Create links to fix files in the FIXgsi directory.


LinGan-NOAA:
There is a chat room conversation on July 31. That concern about the missing observation data in analysis job:

@James Polly - NOAA Affiliate and @Wei Wei - NOAA Affiliate, we noticed that jrrfs_enkf_process_radar and jrrfs_enkf_process_radar_spinup sometimes start a bit late. One of the reasons they aren't triggering at 25 minutes past the hour is because they are waiting on the 'complete' status of the det cycle—which isn't actually necessary.
If we remove the det cycle constraint and change the trigger to :TIME >= 1922 and :TIME < 2225, we can run these two jobs 3 minutes earlier. This will help reduce the risk of missing radar observations in the EnKF analysis. Here, 19z is an example, the change should be used for all enkf cycle.

If this data is mandatory, we need robust exception handling so developers can immediately see what is missing and operations can take appropriate action. Without explicit warning messages, developer support becomes significantly harder when DATA is already being removed at the end of the job. Additionally, we should maintain a document listing all required data per analysis job to ensure consistent output quality.

JamesPolly-NCO:
@lgannoaa thanks for your thoughts Lin. Is this issue re: emails the right place to do this? Should this be in a new issue that is focused on documenting what is and what is not mandatory data?

Creating a document listing all the required data is a great idea. I've been trying to get answers on those questions via github issues for months.

LinGan-NOAA:
@JamesPolly-NCO you are right, creating a new issue is probably the best way to tackle the design flaw in the analysis job. Realistically, we can't rely on a subject matter expert every time we need to check if the job has enough input data.

lgannoaa (approved):
Looks good to me.

MatthewPyle-NOAA (approved):

PR #1293, opened 2026-02-02
[rrfs-mpas-jedi] Replace JEDI obs offline domain check with a JEDI `Polygon Check` pre-filter.

SamuelTrahanNOAA:
## DESCRIPTION OF CHANGES: 

Replaces the offline JEDI obs domain check with a domain check in a JEDI prefilter.

## DEPENDENCIES:

The new filter isn't in JEDI UFO yet; it is here:

- https://github.com/JCSDA/ufo/pull/78

Until the filter PR is merged, there are equivalent `sorc/_workarounds_` changes. They're in `sorc/build.rdas` and `sorc/_workarounds_/PolygonCheck`. They should be removed and replaced by updated JEDI code before this PR is merged.

## TESTS CONDUCTED: 
Three levels of testing:

1. Tested filter outside JEDI in a simple test program with various polygons and obs patterns to confirm it works.
2. New JEDI UFO unit test for the filter on Ursa.
3. Confirmed filter works in rrfs-workflow 3km CONUS on GAEA.

I'm waiting for more thorough independent testing. That's one of the reasons why this is a draft.

### Machines/Platforms:

Affects all machines/platforms for configurations that use MPAS JEDI.

### Test cases: 

All cases that run MPAS JEDI will use this change.

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

## ISSUE: 

- fixes: https://github.com/NOAA-EMC/rrfs-workflow/issues/1294

PR #1637, opened 2026-08-10
Use the "Polygon Check" filter and other relevant changes

guoqing-noaa:
This PR supersedes PR #1293. @spanNOAA has tested and confirmed that the `Polygon Check` worked as expected without the offline domain check python script by comparing the obs counts and JEDIVAR results. This PR uses a simplified polygon with far few vertices, generated via `shapely.geometry.Polygon.simplify`. Details on generating a simplified domain boundary polygon can be found in [this Jupyter Notebook](https://pydamonitor.github.io/docs/notebooks/polygon_boundary_mpas.html)

This PR also adds a filter `identifier` section for each filter following this PR https://github.com/JCSDA-internal/ufo/pull/4241.  Based on several rounds of group "YAML hackathon" sessions, an [RRFSv2 YAML convention](https://github.com/NOAA-EMC/rrfs-workflow/wiki/RRFSv2-YAML-convention) has been established to improve standardization and maintainability, and to prepare for enabling stricter YAML linting checks..


guoqing-noaa:
Tested 24 cycles on Gaea, worked as expected.

guoqing-noaa:
All YAML output differences are due to newly added blocks and hence expected. The new and old YAML files are functionally the same.

MatthewPyle-NOAA (approved):

Junjun-NOAA (approved):

parm/observers/cris-fsr_n21.yaml
```diff
@@ -0,0 +1,637 @@
...
+           flag: CloudDetectMinResidualIR
+           ignore: rejected observations
+         - name: reject
+
+       # --------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+       # Step 13: Surface Temperature Jacobian Check over Land, reduce influence of land pionts on window channels, surface temperaure jacobian oove land larger than 0.2, rejected
+       # ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+       - filter: Bounds Check
+         identifier:
+           name: WindowChannelExtremeResidual
```
HaidaoLin-NOAA: change WindowChannelExtremeResidual to SurfaceTempJacobianCheck

chunhuazhou (approved):

spanNOAA (approved):

HaidaoLin-NOAA (approved):

guoqing-noaa:
@MatthewPyle-NOAA We have a few upcoming PRs which will depend on this PR, such as accommodating both GETKF one-step and two steps runs in the workflow, updating to use `ModelHeightAdjustedAirTemperature` for surface DA, updating to use duplicate thinning filter, updating/adding CrIS and TMS observers, etc. It will be greatly appreciated if this PR may get a priority, thanks!

SamuelDegelia-NOAA:
Are there plans to adopt JCB for the `[rrfs-mpas-jedi]` branch? Splitting the yamls up into templates like this is very similar to how JCB works. Just want to make sure we aren’t duplicating efforts.

Also, since `[rrfs-mpas-jedi]` now makes use of the obs space yaml template format as per this PR, it would be great if we could move/merge these obs-space templates with those in RDASApp. I know this has been brought up many times before, but that would allow for better collaboration of developments between our fv3-jedi and mpas-jedi applications. For example, OMD has made a lot of changes to the obs spaces yamls (which exist as JCB templates in RDASApp) that more closely replicate hofx and error statistics from GSI. We would love to see such developments used in RRFSv2x as well. And likewise, it would be great to be able to test new functionality such as the polygon filter in our fv3-jedi implementation of RRFS. 

If GSL is opposed to JCB for some reason, it still would be possible to just use the obs-space templates in RDASApp but merge them into the full yaml using your python codes.

delippi:
I don't want to hold up any work. 

We do need to have a discussion about JCB.

guoqing-noaa:
@SamuelDegelia-NOAA @delippi  Thanks for the discussions. 

Yes, we will commit relevant observer changes to RDASApp as appropriate and integrate RDASApp updates into the RRFSv2 workflow as well.

To provide a bit more context, in our research and development work, we share many of the same observer/filter configurations across a wide range of use cases and applications (RRFS, RTMA, AR-PS, global, etc.). Because of this,  we would like to embrace a **WYSIWYG (What You See Is What You Get)** experience. This is particularly helpful for standalone JEDI testing, where researchers often need to work directly with and modify the original, human-friendly YAML configuration files with all comments and formatting preserved.

We have a slide deck that illustrates this concept in more detail: https://docs.google.com/presentation/d/1sc46nCJsfTrFK_EZDPtyC12WSJx9WDEl9AJBlXb81go

I believe the RDASApp JCB and RRFSv2 workflows already share at least 95% of the same operator/filter configurations. That is a good thing—it means we help each other. It also gives us a strong foundation to evolve to an agreed-upon solution moving forward. Thanks!

ShunLiu-NOAA:
@guoqing-noaa @SamuelDegelia-NOAA and @delippi , Thank you for your hard work and the great discussion on this. Ultimately, I think we should consolidate the YAML configuration into a single package. If we decide to keep both options for generating YAML (JCB and the method introduced in this PR), we need to test both methods thoroughly to ensure they work. This is similar to how we test any new enhancements for both FV3JEDI and MPASJEDI in RDASApp.

SamuelDegelia-NOAA:
@guoqing-noaa Thanks for the slide deck. I agree with the complaints about JCB/PyYAML not preserving formatting which makes comparison between the templates and final yamls challenging. Since we plan to stick with using JCB, I wonder how hard it would be to integrate such a change into JCB?

guoqing-noaa:
@SamuelDegelia-NOAA PyYAML focuses on reading a YAML into memory and has no intention to preserve comments and formatting.

My take is that in operation, we only need one YAML configuration (or even one super YAML file). So we are okay to do both JCB and WYSIWYG in the workflow at the same time. Since the observer and filter configurations are very close to each other between JCB and WYSIWYG, we don't need to worry about science may change at all. We currently already have a tool to convert RDASApp JCB to WYSIWYG rrfsv2 yaml. I am working on another tool to convert WYSIWYG to JCB as well. So we should be much more straightforward to switch between them in the future.

hu5970:
Thanks for the great discussion on the YAML file. We currently have two paths for generating YAML files, and we can use both for now. We will certainly look for a way to make the YAML files shareable between FV3 and MPAS-JEDI.

hu5970 (approved):
This PR looks good. It cuts three big Yaml files into many small and easy to read Yaml files and save observer part of them under observers directory. There are discussion on how to share the Yaml files between FV3-JEDI and MPAS-JEDI or between JCB and the method used here. We will find sometime soon to share the ideas of how to make those Yaml observer files sharable.

delippi:
I understand there are concerns about a user-readable final yaml. Honestly, my biggest concern is there are two sets of observation configurations that are being maintained (here in this PR and [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/observations/atmosphere)), and only one set has been fully validated against GSI. For example, the t187 (left side of image below) configuration in this PR is missing the halving of the t observation error 

from GSI's `gsd_terrain_match_surfTobs.f90`
```fortran
if(kx>179.and.kx<190) toe=toe*r0_5
```

Screenshot 2026-08-10 155214

This is one of likely many things that have been missed when maintaining two sets of configurations. Likewise, I am certain there are updates that you all have made that are missing from the JCB configurations such as this Polygon Check.

It is in my view absolutely critical that we all work with the same observation templates. We cannot forget to take the crucial first step and correctly transition all of the capabilities from GSI to JEDI and achieve similar performance. Since there is no GSI interface for MPAS, we must rely on the validation that has been done using FV3-JEDI.

Again, I'm not holding the PR back. Go ahead and merge. I don't think we should be working in parallel on this.

guoqing-noaa:
> I understand there are concerns about a user-readable final yaml. Honestly, my biggest concern is there are two sets of observation configurations that are being maintained (here in this PR and [here](https://github.com/NOAA-EMC/RDASApp/tree/develop/parm/jcb-rdas/observations/atmosphere)), and only one set has been fully validated against GSI. For example, the t187 (left side of image below) configuration in this PR is missing the halving of the t observation error
> 
> from GSI's `gsd_terrain_match_surfTobs.f90`
> 
> ```fortran
> if(kx>179.and.kx<190) toe=toe*r0_5
> ```
> 
> Screenshot 2026-08-10 155214
> This is one of likely many things that have been missed when maintaining two sets of configurations. Likewise, I am certain there are updates that you all have made that are missing from the JCB configurations such as this Polygon Check.
> 
> It is in my view absolutely critical that we all work with the same observation templates. We cannot forget to take the crucial first step and correctly transition all of the capabilities from GSI to JEDI and achieve similar performance. Since there is no GSI interface for MPAS, we must rely on the validation that has been done using FV3-JEDI.
> 
> Again, I'm not holding the PR back. Go ahead and merge. I don't think we should be working in parallel on this.

@delippi Thanks for the specific case and thanks for the GSI benchmark work!

Yes, we will test the part you referred to using one week of retros. To be frank, we haven’t extensively tested all of our GSI settings in the past, so this is a good opportunity to revisit them and confirm whether the additional tuning is actually beneficial.  It’s also helpful to have two options so we can cross-check the results and make sure we don’t miss any critical pieces.

MatthewPyle-NOAA:
Thanks for all of the discussion here - will go ahead and merge.

PR #1638, opened 2026-08-11
atms_n21 tlapse file sensor name correction

keenaneure:
This PR is a small fix to the atms_n21.tlapse.txt file needed for the lapseRate predictor in JEDI's variational bias correction. The sensor name in the first column was incorrectly written as "atms_n20" instead of "atms_n21"


MatthewPyle-NOAA (approved):

guoqing-noaa (approved):
Thanks for fixing this, @keenaneure

Issue #1641, opened 2026-08-12
Ecflow event and trigger issue in 07/19 EnKF blend ICs and EnKF ensinit forecast job


LinGan-NOAA:
The current NCO ecflow has a bug that will impact the 07z/19z enkf ICS and forecast step.
In the 07/19 EnKF blend ICs, each member can independently set the skip_ensinit ecFlow event flag if some member in the 18Z EnKF forecast is running slow. This occurs because the ex-script evaluates input files on a per-member basis and lacks built-in inter-member communication. Additionally, the 07/19 EnKF ensinit forecast job is incorrectly being marked as complete due to a faulty "or" condition.

These bugs will impact the scientific results and cause the system to fail under misleading conditions (affecting only a few enkf prep_cyc_spinup members). Resolving in a realtime system will require not only rerunning the workflow starting from the 07/19 enkf ics family, but also cleaning up the misleading flag files in the COM directory.

PR #1642, opened 2026-08-12
[release/v1.0] Fix bug in enkf 07z and 19z blend_ics event and ensinit forecast complete condition

lgannoaa:
## DESCRIPTION OF CHANGES: 
As described in issue #1641 
This PR is to fix the ecflow definition file that impact 07/19z enkf blend_ics event and ensinit forecast jobs.
## TESTS CONDUCTED: 
- WCOSS2
  - [X] Cactus/Dogwood
### Test cases: 
- [X] Engineering tests
    - [X] Retro
    - [X] Parallel


LinGan-NOAA:
@hu5970 @ShunLiu-NOAA 
Please take a look at this PR to review the change.

WeiWei-NCO:
Two changes in this PR:

1. 07z/19z enkf_blend_ics job triggers:
    - These jobs now wait for 06z/18z enkf to complete. 
    - Then the "run_blending" switch in blend job is always set to true, because 06z/18z is always available. 
    - We lose the capability to "run_ensinit". 
    - We also lose the capability to use 04z/05z and 16z/17z restart files in blend ics jobs, when 06z/18z is missing. 
    - As a result, 18z and 19z can't be recovered at the same time. 

2. 07z/19z enkf ensinit job triggers.
    - Current: All members skip ensinit, if any member has "skip_ensinit" flag
    - New: All members skip ensinit, if **all** members have "skip_ensinit" flag
    - This makes sense because we don't want members behave differently, and if one member does ensinit, the rest should follow. 

Given the change in (1), change (2) seems to be redundant. Since 07z/19z will always wait for 06z/18z enkf forecast, is there any scenario where enkf_ensinit would actually run? We might even be able to remove all enkf ensinit jobs from production ecflow.

ecf/defs/rrfs_nco_para.def
```diff
@@ -19548,181 +19548,181 @@ suite para
                 task jrrfs_enkf_make_ics_mem001
                   edit MEMBER_NAME '001'
                 task jrrfs_enkf_blend_ics_mem001
-                  trigger ./jrrfs_enkf_make_ics_mem001==complete
+                  trigger ./jrrfs_enkf_make_ics_mem001==complete and ../../../06z/enkf/forecast/jrrfs_enkf_save_restart_f1_done == complete
```
ShunLiu-NOAA: We shouldn't use ../../../06z/enkf/forecast/jrrfs_enkf_save_restart_f1_done == complete as trigger condition here. This change will overwrite fallback functions in exrrfs_blend_ics.sh and also blending job will not be able to cold-start.

ecf/defs/rrfs_nco_para.def
```diff
@@ -20169,11 +20169,11 @@ suite para
...
...it and ../ics/jrrfs_enkf_blend_ics_mem008:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem009:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem010:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem011:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem012:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem013:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem014:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem015:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem016:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem017:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem018:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem019:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem020:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem021:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem022:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem023:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem024:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem025:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem026:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem027:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem028:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem029:skip_ensinit and ../ics/jrrfs_enkf_blend_ics_mem030:skip_ensinit )
```
ShunLiu-NOAA, line 20172: The current logic dictates that if any member runs the enkf_forecast_ensinit job, all other members must run it as well. Because of this, the change proposed here is unnecessary.

hu5970:
The 07/19Z ensemble initialization and job triggering logic is indeed very complicated. I need to think about it and find a time to discuss these workflow changes.

ecf/defs/rrfs_nco_para.def
LinGan-NOAA: The current logic says if any member of the blend_ics job set "skip_ensinit" event - Not run the enkf_forecast_ensinit job. All enkf_forecast_ensinit job will automatically set to complete (Not run).

The new solution is to safe guard it by require all member set the "skip_ensinit" event to ensure workflow consistency on enkf_forecast_ensinit job proceeding.

ecf/defs/rrfs_nco_para.def
LinGan-NOAA: Our guidance is the major cycle 00,06,12,18 can not be skipped for RRFS system. Therefore, the 06z/18z will not be skipped. If the team decided to change this fundamental requirement. It will open up more option to resolve this  issue.

ecf/defs/rrfs_nco_para.def
WeiWei-NCO: I would suggest using a single event for "run_ensinit", possibly adding a new dummy job for it. If any blend_ics job needs to run ensinit, it can set "run_ensinit". If multiple blend jobs need ensinit, they all set the same "run_ensinit" event from the same dummy job. Then, forecast_ensinit for all members can be triggered by it. 

Not necessarily for v1.0 though. 

Edit: I need to think/check if a job can complete based on a clear event.

Edit2: This seems work.  `complete not ../ics/jrrfs_enkf_run_ensinit:run_ensinit`

LinGan-NOAA:
@WeiWei-NCO and @hu5970 @JamesPolly-NCO 
I had a meeting with Ming and explain the issue/solution in this PR. It is my understand that we have reached the agreement that this issue is confirmed and need to be addressed.
The forecast job complete condition change from "or" to "and" is confirmed to be necessary.
Rather than waiting for the previous cycle's forecast job to finish before triggering, two ex-scripts were fixed to ensure all members run stably and consistently. Additionally, fallback functionality for major cycles has been preserved.

Summary:
These jobs will **not need to wait** for 06z/18z enkf to complete.
We will **not lose** the capability to "run_ensinit".
We will **not lose** the capability to use 04z/05z and 16z/17z restart files in blend ics jobs, when 06z/18z is missing.
18z and 19z **can be recovered** at the same time.

@hu5970 
A test has been done. Job log can be found in Dogwood:/lfs/h2/emc/ptmp/Lin.Gan/ecfops/com/output/prod/today/2026081307. This PR is ready for its final review. Please take a look.

JamesPolly-NCO:
Notes from meeting with @lgannoaa and @hu5970 

This PR is addressing an unlikely corner case, where the workflow is recovering from a previous failure, and is intended to eliminate the possibility of different ensemble members behaving differently during this corner case scenario.

@lgannoaa please describe the test you refer to here:
> A test has been done. Job log can be found in Dogwood:/lfs/h2/emc/ptmp/Lin.Gan/ecfops/com/output/prod/today/2026081307. This PR is ready for its final review. Please take a look.

@lgannoaa are there output files associated with your second test (mentioned verbally) which test this PR under normal (i.e. not catching up from a previous failure) conditions?

Next steps:
- @hu5970 will review the test output logs provided in the above post by Lin.
- We will continue to discuss this next week to ensure that everyone is happy.

LinGan-NOAA:
The test case for 2026081307 is:
rrfs_enkf_blend_ics_mem001_07.o135296999 - show "File missing: /lfs/f2/t2o/ptmp/emc/lin.gan/ecflow_rrfs/test/com/rrfs/v1.0/enkfrrfs.20260813/06/m001/forecast/RESTART/20260813.070000.coupler.res"
While other 29 members include rrfs_enkf_blend_ics_mem030_07.o135297009 found the RESTART file with:
NUM_ENS_MEMBERS_FOUND=YES

All ensinit member running consistent:
grep "Forecast job is completed" rrfs_enkf_forecast_ensinit_mem???_07.o*|wc -l
30

Normal cycled run condition testing started on 2026081412. Will continue to 2026081419.

hu5970 (approved):
The code changes look good. Both cold start test and regular test are passed.

MatthewPyle-NOAA (approved):

ShunLiu-NOAA (approved):

WeiWei-NCO (approved):
Makes sense to me, especially since it can still fall back to earlier cycles.

PR #1656, opened 2026-08-18
Removal of GOES AMV expected normalized error checks

keenaneure:
This PR removes the expected normalized error check for GOES AMVs for all 6 observation spaces - satwnd_uv245_270, satwnd_uv245_272, satwnd_uv246_270, satwnd_uv246_272, satwnd_uv247_270, and satwnd_uv247_272.

Our IODA files do not actually have MetaData/exp_err_norm as the filter requires. The closest field is MetaData/expectedError (which is not normalized between 0 and 1 like the filter would expect), but it is not abundantly clear that including the field makes an appreciable difference in results.

### DIFF
--- a/parm/observers/satwnd_uv245_270.yaml
+++ b/parm/observers/satwnd_uv245_270.yaml
@@ -385,6 +385,28 @@
            action:
              name: reject
 
+#         # AMV expected error norm check
+#         - filter: Perform Action
+#           apply at iterations: 0,1
+#           identifier:
+#             name: AMVErrorNormCheck
+#             logging: false
+#             diagnostic flag new: false
+#           filter variables:
+#           - name: windEastward
+#           - name: windNorthward
+#           where:
+#           - variable: ObsFunction/Velocity
+#             minvalue: 0.1
+#           - variable: MetaData/exp_err_norm
+#             minvalue: 0.9
+#           - variable: ObsType/windEastward
+#             is_in: 245
+#           - variable: MetaData/satelliteIdentifier
+#             is_in: 270
+#           action:
+#             name: reject
+
          # Halve the ob error (based on read_satwnd.f90)
          - filter: Perform Action
            apply at iterations: 0,1
--- a/parm/observers/satwnd_uv245_270.yaml
+++ b/parm/observers/satwnd_uv245_270.yaml
@@ -385,28 +385,6 @@
            action:
              name: reject
 
-         # AMV expected error norm check
-         - filter: Perform Action
-           apply at iterations: 0,1
-           identifier:
-             name: AMVErrorNormCheck
-             logging: false
-             diagnostic flag new: false
-           filter variables:
-           - name: windEastward
-           - name: windNorthward
-           where:
-           - variable: ObsFunction/Velocity
-             minvalue: 0.1
-           - variable: MetaData/exp_err_norm
-             minvalue: 0.9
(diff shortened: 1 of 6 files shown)


delippi:
@keenaneure I've been adding that MetaData via a separate tool. I don't remember the impact of that filter. I could test and let you know.

keenaneure:
> @keenaneure I've been adding that MetaData via a separate tool. I don't remember the impact of that filter. I could test and let you know.

Sure - I would be curious to hear what you find.

xyzemc:
@keenaneure Have you experienced the JEDI crash or warning message when MetaData/exp_err_norm is not exist but you included it in the JEDI yaml file?

delippi:
@keenaneure removing those filters were enough to fail the following ctest: rrfs_fv3jedi_2024052700_3dvar_remote. It seems to remove 486 satwnd_winds_247_272 (all off them) from assimilation in the ctest. It also seems we set those ob errors very large such they have little impact anyway. There are a few other slight changes to observation counts for other satwnd types.

You can build RDASApp and run the rrfs_fv3jedi_2024052700_3dvar_remote ctest. You can directly modify the `rrfs_fv3jedi_2024052700_3dvar_remote.yaml` to remove those filters to test the impact.

I'm also curious about @xyzemc question.

keenaneure:
> @keenaneure Have you experienced the JEDI crash or warning message when MetaData/exp_err_norm is not exist but you included it in the JEDI yaml file?

Yes, this crashes in its current form, so I comment it out on every run. The converter would need to be edited in order to include this field.

delippi:
In case you wanted to know how we add it for [dev-sci] branch, but I know you guys are not using offline tools. It could likely be added to the converter itself.

https://github.com/NOAA-EMC/rrfs-workflow/blob/ac32e1593c675aec86aeda20184b992573f641e0/scripts/exrrfs_ioda_bufr.sh#L375

https://github.com/NOAA-EMC/RDASApp/blob/43616bd98ca5f3a747cabcbd27b1d8ce41edde18/rrfs-test/IODA/offline_ioda_patch.py#L182-L193

guoqing-noaa:
@delippi GSI does a similar computation at runtime, right? If yes and JEDI misses this capability, we would like to add it to JEDI. Thanks!

guoqing-noaa:
@keenaneure Maybe we want to comment out the `exp_err_norm` filters instead of removals? We will finally turn it on sometime in the future. Thanks!

delippi:
@guoqing-noaa Yes, GSI does this at run time. I agree commenting them out for now is a good temporary solution.

It would be great to add to JEDI eventually. Alternatively, it is also likely possible to do this in the YAML with derived quantities too.

guoqing-noaa:
Thanks, @delippi!

Yes, we may use two "variable assignment" filters to accomplish what you did in offline_ioda_patch.py. 
However, it may still be beneficial to combine them into one filter using a new obs function. We will see how other JCSDA partners say about this when we create a PR for this. Thanks!

keenaneure:
I am fine commenting it out for now. I can change that if it makes sense. Maybe part of this is dependent on whether including the filter makes a noticeable difference in results?

MatthewPyle-NOAA (approved):

guoqing-noaa (approved):
@keenaneure Thanks for the changes. Let's take a note and revisit this when exp_err_norm is available.

PR #1658, opened 2026-08-19
[rrfs-mpas-jedi]Add changes to run one-step GETKF, along with some other changes

chunhuazhou:
## DESCRIPTION OF CHANGES: 

- Re-do the PR to run one-step GETKF — change the workflow/configuration so GETKF runs as a single step rather than the previous two-step approach.
- Add back 3 surface variables — restore three surface-related variables that were previously removed from stream_list.atmosphere.da_state.
- Change recenter behavior — if the recenter task cannot find a control file, it should issue a warning and exit instead of failing. This is useful for real-time runs where a deterministic analysis may not exist.
- Update exp.conus3km and exp.ens_conus3m — adjust those experiment configurations to more closely match the settings used by RRFSV2X.

## TESTS CONDUCTED: 
Testing it in retro mode on Gaea


hu5970 (approved):
Looks good to me. This is a important PR to stable the RRFS system for a while.Thanks.

ush/yamltools4rrfs.py
```diff
@@ -164,31 +163,6 @@ def getkf_observer_tweak(data, getkf_type):
...
-            nspace = hy.strip_indentations(data[pos])[0]
-            for i in range(pos - 1, -1, -1):
-                nspace2 = hy.strip_indentations(data[i])[0]
-                if data[i].strip().startswith('#') and nspace2 <= nspace:
-                    pos = i
-                else:
-                    break  # exit the loop if non-comment
-            del data[pos:next_one]
-
-
```
guoqing-noaa, line 191: @SamuelDegelia-NOAA I recall that we need to remove the "reduce obs space" actions and "temporal thinning" filters for getkf_post tasks, right?

@chunhuazhou Do we run getkf_post in rrfsv2x at the moment?

SamuelDegelia-NOAA: @guoqing-noaa Just clarified via email that I do not think these filter/actions need to be removed anymore.

guoqing-noaa: @SamuelDegelia-NOAA Great! Thanks for the email and the clarification! 
So it is correct that we should no longer do this removal. Thanks!

chunhuazhou:
Successfully tested it for a few cycles on GAEA. 
@SamuelDegelia-NOAA Thank you for the clarification!
@guoqing-noaa No we don't do getkf_post in the real-time RRFSv2X. Thanks!

MatthewPyle-NOAA (approved):

guoqing-noaa:
@chunhuazhou Thanks for lots of great works in this PR!
As discussed in last week's joint GSL-OMD DA meeting, it will be nice to use `getkf` as the task name instead of `getkf_observer_solver` since the one-step GETKF run will be our default and to-go everyday configuration and it will be beneficial to use a concise task name. At this moment, I don't see any unresolvable tech hurdle to do this. So I will try to test this based on your latest commit. Hope I may get it ready late this afternoon. Thanks!

hu5970:
@guoqing-noaa Your new PR is going to on top of this PR, right? Should this PR get merged and your new PR to change the task name for getkf following this one?

guoqing-noaa:
> @guoqing-noaa Your new PR is going to on top of this PR, right? Should this PR get merged and your new PR to change the task name for getkf following this one?

I had planned to submit changes to Chunhua's PR branch. But if it also fine that we merge this first and then I create a new PR.

PR #1662, opened 2026-08-20
Fix checkDA for GETKF workflow

LiaofanLin-NOAA:
This PR updates checkDA to support GETKF ensemble workflows.

The previous implementation assumed that the hybrid DA variables (HYB_WGT_ENS, HYB_WGT_STATIC, etc.) were defined, which caused checkDA to fail for GETKF workflows. This update handles the GETKF case separately and uses getkf.yaml for the observation configuration.

Tested with an RRFS MPAS-JEDI ensemble workflow, and checkDA now completes successfully and reports the assimilated observations.

### DIFF
--- a/workflow/tools/qrocoto/checkDA
+++ b/workflow/tools/qrocoto/checkDA
@@ -61,23 +61,26 @@ with open(f'{expdir}/{NET}.xml') as infile:
             SAT_USELIST = get_value(line)
 
 # prepare a summary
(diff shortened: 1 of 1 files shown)


guoqing-noaa (approved):
LGTM. Thanks, @LiaofanLin-NOAA

MatthewPyle-NOAA (approved):

PR #1668, opened 2026-08-21
[release v1.0] add dbn_alert for non-WMO-headed AWIPS file

JamesPolly-NCO:
## DESCRIPTION OF CHANGES: 

- Adds a dbn_alert for the non-WMO-headed, 3-km grid, AWIPS product.

## TESTS CONDUCTED: 

- not tested, needs discussion/review.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [ ] Engineering tests
  - [ ] Non-DA engineering test
  - [ ] DA engineering test
    - [ ] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:

### DIFF
--- a/scripts/exrrfs_prdgen.sh
+++ b/scripts/exrrfs_prdgen.sh
@@ -224,6 +224,11 @@ if [ -s ${COMOUT}/${fld2d} ]; then
           ${COMOUT}/${fld2d}
       $DBNROOT/bin/dbn_alert MODEL RRFS_ENS_2DFLD_NA_IDX $job \
           ${COMOUT}/${fld2d}.idx
+    elif [[ "${PREDEF_GRID_NAME}" = "RRFS_NA_3km" ]]; then
+      $DBNROOT/bin/dbn_alert MODEL RRFS_DET_2DFLD_NA3 $job \
+          ${COMOUT}/${fld2d}
+      $DBNROOT/bin/dbn_alert MODEL RRFS_DET_2DFLD_NA3_IDX $job \
+          ${COMOUT}/${fld2d}.idx
     fi
   fi
 fi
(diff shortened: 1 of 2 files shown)


WeiWei-NCO:
I would alert them in exrrfs_prdgen.sh, like all other non-headed products, so they are isolated from the awips subroutine. 

https://github.com/NOAA-EMC/rrfs-workflow/blob/a0e9edaad29841a2af709e2b1b73bafa9030d257/scripts/exrrfs_prdgen.sh#L219-L229

JamesPolly-NCO:
@WeiWei-NCO thanks for the feedback, I've moved it. Please take a look and let me know what you think.

MatthewPyle-NOAA:
Is the thought to alert the limited set of 3 km NA products.   Or is it actually the full 2DFLD file?   I was thinking it was the more limited set, which would have to be done within the exrrfs_mkawp.sh script (after the more limited product file is assembled there).

WeiWei-NCO:
Good point, @MatthewPyle-NOAA. 

Does the non-headed file with limited set of products have the same name and only exist in /TMP? If we are going to only alert the subset, we need to copy the file to /COM with a different name, then alert the new file.

MatthewPyle-NOAA:
Yes, it would need to be copied to COM with a distinct name.  There already is a NA 3 km subset product in COM (such as rrfs.tCCz.subset.3km.fFFF.na.grib2), so not sure what an optimal name would be.

MatthewPyle-NOAA:
Have a possible way of doing this in a draft PR: https://github.com/NOAA-EMC/rrfs-workflow/pull/1671.  Not sure if the inclusion of awips in the file name would make Margaret's head explode, but seems accurate in this case.

JamesPolly-NCO:
Thanks Matt, I'm trying to keep the convo all in one place. I believe the preferred alert name is: RRFS_DET_2DFLD_NA3

If you're not opposed to that, it would satisfy the HPC DF request.

PR #1674, opened 2026-08-24
Generic uv10m

spanNOAA:
## Summary
Updates 10-meter surface wind observer configurations (`adpsfc_uv*.yaml`) to incorporate geometric height interpolation, precise observation height assignments, and updated error tuning parameters.

## Detailed Changes
- **Obs Operators**: Configured vertical interpolation in the `obs operator` based on geometric height.
- **Obs Filters**:
  - Inserted a `Variable Assignment` filter immediately following the `reduce obs space` step to set surface wind observations at model height + 10m.
  - Updated observation error for `uv 284` from `1.079` to `1.5874`, aligning with the RRFS GSI error table.
  - Enabled error inflation (`pressure check` filter) for `uv 287`.
- **Consistency Audit**: Standardized filter sequences, options, and formatting across all affected YAML files (`adpsfc_uv281`, `adpsfc_uv284`, and `adpsfc_uv287`).


guoqing-noaa:
@spanNOAA I suggest you start a new branch based on the latest rrfs-mpas-jedi instead of reuse the old branch
The commit history does not make sense since such as 28f0281, 7f6cb6c have been merged already.

parm/observers/adpsfc_uv281.yaml
```diff
@@ -140,6 +137,23 @@
...
+            - name: DerivedMetaData/height
+              type: float
+              function:
+                name: ObsFunction/LinearCombination
+                options:
+                  variables:
+                  - name: GeoVaLs/height_above_mean_sea_level_at_surface
+                  coefs: [1.0]
+                  intercept: 10.0
+
```
delippi, line 156: Just a heads up. This might mess with the ObsErrorFactorPressureCheck which also adds 10m to the to station elevation to get obs height.

```
  /// 2) AddObsHeightToStationElevation: true
  ///    obs_height = dstn + AssumedSfcWndObsHeight;
```

```yaml
               name: ObsFunction/ObsErrorFactorPressureCheck
               options:
                 variable: windEastward
                 inflation factor: 4.0
                 SetSfcWndObsHeight: true
                 AddObsHeightToStationElevation: true
                 AssumedSfcWndObsHeight: 10
```

spanNOAA: @delippi Thanks for raising this! I don't think there is a conflict here, but your comment actually highlights a great architectural point.

**Why there's no conflict currently:**

* **Local Scope:** Any height adjustments made inside `ObsErrorFactorPressureCheck` are strictly local to its internal calculations. They are not written back to the IODA container or shared with downstream operators and filters.
* **Physical Logic & Representativeness:** Setting `AddObsHeightToStationElevation: true` and `AssumedSfcWndObsHeight: 10` ensures `obs_height - stationElevation = 10m`. This forces `fact = 0.0`, evaluating 10m surface winds against **10m above the model surface** (`dpres = 10m`). The elevation difference between the station and model terrain is then captured via `drpx` (`0.005 * |dstn - zsges|`), which inflates the error to account for local terrain representativeness.

**Long-term Architectural Proposal:**

Moving forward, I think a better approach would be to **decouple height assignment from `ObsErrorFactorPressureCheck` entirely** and handle it explicitly via a `Variable Assignment` filter prior to running QC/error inflation.

**Benefits:**

* **Separation of Concerns:** `ObsErrorFactorPressureCheck` can focus purely on QC and error inflation logic, rather than relying on extra filter parameters to mutate missing `MetaData/height` values internally.
* **Single Source of Truth:** Deriving `MetaData/height` beforehand guarantees consistent height metadata across all downstream observation operators, diagnostic outputs, and other filters.
* **Cleaner Code & YAML:** We can eventually deprecate the redundant configuration options (`SetSfcWndObsHeight`, `AddObsHeightToStationElevation`, `AssumedSfcWndObsHeight`), making both the C++ implementation and YAML configurations much cleaner and easier to maintain.

Let me know what you think! @guoqing-noaa, would love to hear your thoughts on this as well.

guoqing-noaa: Good discussions!

I agree that `DerivedMetaData/height` here and `ObsFunction/ObsErrorFactorPressureCheck` (use `MetaData/height`, see below) don't affect each other

```
# src/ufo/filters/obsfunctions/ObsErrorFactorPressureCheck.cc
....
  // Include list of required data from MetaData
  invars_ += Variable("MetaData/height");
```

PR #1684, opened 2026-08-27
Add checks to nonvar cld ana obs processing tasks

ShawnMurdzek-NOAA:
Implement changes to help avoid dead tasks during realtime when input observation files to the nonvar cloud analysis are missing.

## DESCRIPTION OF CHANGES: 
- Implement a `STOP_IF_NO_OBS` flag in the `nonvar_bufrobs` and `nonvar_reflobs` tasks. If `STOP_IF_NO_OBS = 1`, the task terminates with an error if an input observation file is missing. If `STOP_IF_NO_OBS = 0`, task continues to execute if an observation file is missing, but a warning is issued.
- Set `STOP_IF_NO_OBS = 0` for realtime experiments and `STOP_IF_NO_OBS = 1` for retros.

## TESTS CONDUCTED: 
Following tests were run, and the output was as expected:
- Default `exp.conus3km` setup: Both the `nonvar_bufrobs` and `nonvar_reflobs` tasks run properly without error.
- Default `exp.conus3km` setup, but with incorrect `OBSPATH` in `nonvar_bufrobs`: The `nonvar_bufrobs` task crashes owing to missing observation files.
- Default `exp.conus3km` setup, but with incorrect `OBSPATH` in `nonvar_bufrobs` and `STOP_IF_NO_OBS = 0`: The `nonvar_bufrobs` task runs without error even though no processed observation files are created.

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules
  - [x] Gaea C6

### DIFF
--- a/scripts/exrrfs_nonvar_bufrobs.sh
+++ b/scripts/exrrfs_nonvar_bufrobs.sh
@@ -84,7 +84,7 @@ else
 if [[ ${STRICT_CHECK} == 1 ]]; then
   echo "ERROR: Missing satellite cloud-top observations"
   exit 1
-else	
+else
   echo "WARNING: Missing satellite cloud-top observations"
   echo "Cloud-top observations will NOT be processed for the nonvar cloud analysis"
 fi
@@ -135,7 +135,7 @@ else
 if [[ ${STRICT_CHECK} == 1 ]]; then
   echo "ERROR: Missing lightning observations"
   exit 1
-else	
+else
   echo "WARNING: Missing lightning observations"
   echo "Lightning observations will NOT be processed for the nonvar cloud analysis"
 fi
@@ -184,7 +184,7 @@ else
 if [[ ${STRICT_CHECK} == 1 ]]; then
   echo "ERROR: Missing METAR observations"
   exit 1
-else	
+else
   echo "WARNING: Missing METAR observations"
   echo "METAR observations will NOT be processed for the nonvar cloud analysis"
 fi
--- a/scripts/exrrfs_nonvar_bufrobs.sh
+++ b/scripts/exrrfs_nonvar_bufrobs.sh
@@ -81,9 +81,13 @@ ${cpreq} NASALaRC_cloud4mpas.bin "${COMOUT}/nonvar_bufrobs/${WGF}/NASALaRC_cloud
 
 else
 
-echo "WARNING: Missing satellite cloud-top observations"
-echo "Cloud-top observations will NOT be processed for the nonvar cloud analysis"
-echo "File: ${lgycld_bufr}"
+if [[ ${STRICT_CHECK} -eq 1 ]]; then
+  echo "ERROR: Missing satellite cloud-top observations"
+  exit 1
+else	
+  echo "WARNING: Missing satellite cloud-top observations"
+  echo "Cloud-top observations will NOT be processed for the nonvar cloud analysis"
+fi
 
 fi
 
@@ -128,9 +132,13 @@ ${cpreq} LightningInMPAS.dat "${COMOUT}/nonvar_bufrobs/${WGF}/LightningInMPAS.da
 
 else
 
-echo "WARNING: Missing lightning observations"
-echo "Lightning observations will NOT be processed for the nonvar cloud analysis"
-echo "File: ${lghtng_bufr}"
+if [[ ${STRICT_CHECK} -eq 1 ]]; then
+  echo "ERROR: Missing lightning observations"
+  exit 1
+else	
+  echo "WARNING: Missing lightning observations"
+  echo "Lightning observations will NOT be processed for the nonvar cloud analysis"
+fi
 
 fi
 
@@ -173,9 +181,13 @@ ${cpreq} mpas_metarcloud.bin "${COMOUT}/nonvar_bufrobs/${WGF}/mpas_metarcloud.bi
 
 else
 
-echo "WARNING: Missing METAR observations"
-echo "METAR observations will NOT be processed for the nonvar cloud analysis"
-echo "File: ${metar_bufr}"
+if [[ ${STRICT_CHECK} -eq 1 ]]; then
+  echo "ERROR: Missing METAR observations"
+  exit 1
+else	
+  echo "WARNING: Missing METAR observations"
+  echo "METAR observations will NOT be processed for the nonvar cloud analysis"
+fi
 
 fi
 
--- a/scripts/exrrfs_nonvar_bufrobs.sh
+++ b/scripts/exrrfs_nonvar_bufrobs.sh
@@ -56,17 +56,17 @@ if [[ -s ${lgycld_bufr} ]]; then
   fi
 
   cat << EOF > namelist.nasalarc
-   &setup
-    analysis_time = ${CDATE},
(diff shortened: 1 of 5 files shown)


ShawnMurdzek-NOAA:
@guoqing-noaa I see that I'm getting an error in the automated tests because the XML file has been updated. The XML file change was done intentionally, so is there a way for me to update the test? Thanks!

hu5970 (approved):

hu5970:
@ShawnMurdzek-NOAA Should we also remove the observation file dependence in the real-time xml task and only depend on the start time.

scripts/exrrfs_nonvar_bufrobs.sh
```diff
@@ -29,7 +29,11 @@ ${cpreq} "${meshgriddir}"/"${MESH_NAME}".grid.nc mesh.nc
 #-----------------------------------------------------------------------
 #
 
-${cpreq} "${OBSPATH}/${CDATE}.rap.t${cyc}z.lgycld.tm00.bufr_d" lgycld.bufr_d
+lgycld_bufr="${OBSPATH}/${CDATE}.rap.t${cyc}z.lgycld.tm00.bufr_d"
+
+if [[ -s ${lgycld_bufr} ]]; then
+
+${cpreq} ${lgycld_bufr} lgycld.bufr_d
```
guoqing-noaa: It is preferred to indent from line 36 to Line 80 and also lines 84-90.
I understand that this will cause lots of git differences.

To save a more trackable change history, my previous practice is that we add the changes first without changing the alignment, just like you did right now.
And then in a subsequent commit, we change the alignment.

ShawnMurdzek-NOAA:
> @ShawnMurdzek-NOAA Should we also remove the observation file dependence in the real-time xml task and only depend on the start time.

Sure, I think that should be fine as long as our time dependence is appropriate. I just added that changed and verified that the resulting realtime and retro workflows have the proper dependencies for `nonvar_bufrobs`

workflow/rocoto_funcs/nonvar_bufrobs.py
```diff
@@ -11,14 +11,20 @@ def nonvar_bufrobs(xmlFile, expdir):
...
     # Task-specific EnVars beyond the task_common_vars
     dcTaskEnv = {
         'REFERENCE_TIME': '@Y-@m-@dT@H:00:00Z',
         'OBSPATH': f'{OBSPATH}',
         'NONVAR_LARC_NPTS': os.getenv('NONVAR_LARC_NPTS', 'NONVAR_LARC_NPTS_not_defined'),
         'NONVAR_METAR_IMPACT': os.getenv('NONVAR_METAR_IMPACT', 'NONVAR_METAR_IMPACT_not_defined'),
         'NONVAR_PROJ_NAME': os.getenv('NONVAR_PROJ_NAME', 'NONVAR_PROJ_NAME_not_defined'),
-        'NONVAR_USERDX': os.getenv('NONVAR_USERDX', 'NONVAR_USERDX_not_defined')
+        'NONVAR_USERDX': os.getenv('NONVAR_USERDX', 'NONVAR_USERDX_not_defined'),
+        'STRICT_CHECK': f'{strict_check}'
```
guoqing-noaa: `STRICT_CHECK` is descriptive in the given situation. But for those who are not familiar with the context, they may wonder what check we want to be strict. 

Will `STOP_IF_NO_OBS` be a little bit more explicit?

guoqing-noaa:
> @guoqing-noaa I see that I'm getting an error in the automated tests because the XML file has been updated. The XML file change was done intentionally, so is there a way for me to update the test? Thanks!

@ShawnMurdzek-NOAA Yes, we have a mechanism to ignore the XML differences if they are expected.
Go to `rrfs-workflow/tests`,  vi `.ignore.compare_xml_yaml_outputs` and replace the first line with your PR number.

workflow/rocoto_funcs/nonvar_bufrobs.py
ShawnMurdzek-NOAA: Good suggestion! I made the change as recommended

scripts/exrrfs_nonvar_bufrobs.sh
ShawnMurdzek-NOAA: @guoqing-noaa Thanks for the suggestion! I as able to indent all lines except the `EOF` lines. Indenting those lines causes an error (as expected). Is the current formatting acceptable? Or is there another way I should handle the namelist file creation?

guoqing-noaa: @ShawnMurdzek-NOAA Thanks for making those changes!
Good point about the namelist file creation, which is an exception and we don't need to indent that part.

scripts/exrrfs_nonvar_bufrobs.sh
```diff
@@ -85,30 +101,46 @@ ${cpreq} NASALaRC_cloud4mpas.bin "${COMOUT}/nonvar_bufrobs/${WGF}/NASALaRC_cloud
...
+   &setup
+    analysis_time = ${CDATE},
+    minute=00,
+    trange_start=-10,
+    trange_end=10,
+    obs_type = "bufr",
+    proj_name = '${NONVAR_PROJ_NAME}',
+    search_rad = ${NONVAR_SEARCH_RAD},
+    debug=0
+   /
```
guoqing-noaa, line 120: Lines 111-120: As we discussed, we don't need to indent the namelist part utilizing a Heredoc. Feel free to revert the changes in these lines.  (we only need to indent the first line, i.e. line 110 `cat <<EOF....`)

guoqing-noaa: This applies to `namelist.nasalarc`, `namelist.lightning`, `namelist.metarcld`.
But if the current indented format works without any issues, it is okay to go with the current form.

ShawnMurdzek-NOAA: @guoqing-noaa It looks like only the `EOF` line needs to be unindented, but I'd prefer to have the entire namelist unindented. I just added a commit that unindents the entire namelist and verified that the updated code runs without error. Let me know if there is anything else I should change at this point!

guoqing-noaa (approved):
LGTM. Thanks for all the changes, @ShawnMurdzek-NOAA!

For future reference, you may want to modify the first post to reflect the latest situation

scripts/exrrfs_nonvar_reflobs.sh
MatthewPyle-NOAA, line 1: I'd like to retain the "FATAL ERROR" in the updated print logic here (at line 105) as it is ahead of an err_exit.

MatthewPyle-NOAA (commented):
One small request to clean up, then can get this merged.

scripts/exrrfs_nonvar_reflobs.sh
ShawnMurdzek-NOAA: No problem! I added the `FATAL ERROR` back in

MatthewPyle-NOAA (approved):

Issue #1649, opened 2026-08-17
[rrfs-mpas-jedi] Towards hourly cycling of smoke and dust for aerosol DA retro testing


huiying-noaa:
**Background**
We're looking to conduct some hourly cycled tuning and testing as the initial aerosol DA capability development concluded. After some preliminary study, May 12, 2024, within the May 2024 testing period with staged data, is selected as our test case with the presence of Canada fire smoke from British Columbia and Alberta and good VIIRS AOD data coverage (details in [Results_20260727_preliminary test period.pptx](https://docs.google.com/presentation/d/195mWDYHo1yKq6R5gZx7O_PVzs5PtyMbi/edit?usp=drive_link&ouid=103593389788427541905&rtpof=true&sd=true)). During the cycled hourly test run with met DA (no aerosol DA) from a clean clone, we found various issues with the current workflow. The next section logs all issues that has been resolved with help from various parties, including the ones tagged below, @EricJames-NOAA, and Steve Fletcher, and problems we currently have for **your further input**. For fixes with future PRs needed, please help review if everything logged is correct and I can submit PR after coordinating with @guoqing-noaa etc. Many thanks! 

Workflow version and MPAS model: 
- Clean clone on Aug 3 from the most updated workflow version [d7b77db61384443c187e9b9d20333aaac64a6eb6](https://github.com/NOAA-EMC/rrfs-workflow/tree/d7b77db61384443c187e9b9d20333aaac64a6eb6)
- build.all chemrrfs 
- Gaea location: /gpfs/f6/bil-fire2-oar/scratch/Huiying.Luo/workflow_260803_prs/rrfs-workflow/, with some minor code changes gradually as logged in the actions due to issues we found.

All tests done on Gaea with all default file structures: 
- Test 1-6 locations: /gpfs/f6/bil-fire2-oar/scratch/Huiying.Luo/workflow_260803_prs/Test1, etc. 
- Test 1-6 configurations: /gpfs/f6/bil-fire2-oar/scratch/Huiying.Luo/workflow_260803_prs/rrfs-workflow/workflow/exp.conus12km_Test1 and config.chemistry_Test1, etc.

**Issues and Actions**
**1. Failed IC at 00z in Test 1**
- Error with looped info in log and finally timed out: 
```
[...]

PR #1689, opened 2026-08-30
A few miscellaneous updates

guoqing-noaa:
1. bugfix for coldstart-only experiments where CYC_INTERVAL is not defined
2. minor updates to template exp files so that the NET part comes from the `${NET}` variable as expected.
3. update .github/workflows/compare_yaml_outputs.yaml to match the recent changes.

### DIFF
--- a/workflow/rocoto_funcs/prep_ic.py
+++ b/workflow/rocoto_funcs/prep_ic.py
@@ -15,7 +15,7 @@ def prep_ic(xmlFile, expdir, do_ensemble=False, spinup_mode=0):
     else:
         cycledefs = 'prod'
     coldhrs = os.getenv('COLDSTART_CYCS', '03 15')
-    cyc_interval = os.getenv('CYC_INTERVAL')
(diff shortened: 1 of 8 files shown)


MatthewPyle-NOAA (approved):

workflow/rocoto_funcs/prep_ic.py
huiying-noaa, line 1: Thanks, @guoqing-noaa! Does this mean we'll be able to set increment_hours =  CYC_INTERVAL  in chem_ic_update.sh after this PR for the issue in Section 1 of https://github.com/NOAA-EMC/rrfs-workflow/issues/1649? Thanks!

chunhuazhou (approved):

Junjun-NOAA (approved):

workflow/rocoto_funcs/prep_ic.py
guoqing-noaa: @huiying-noaa It is a step forward. I think it works for DA situation. But we still need some updates to make the cold-start-only experiment run correctly as well.

Issue #1702, opened 2026-09-04
[rrfs-mpas-jedi] GSIBEC crashed for conus12km and na12km


guoqing-noaa:
It was recently found that we cannot run conus12km 3DVAR retros with the recent rrfs-workflow commits. The error message is attached below. It is related to the Jb NaN issue from GSIBEC we saw before. 

We tracked this down to `v2.1.4-4.8-734df15` (PR #1648) where we updated RDASApp and removed previous GSIBEC workarounds.

Before -4.8, `v2.1.4-4.7-884be1c` ran without any issues:
`/gpfs/f6/arfs-gsl/world-shared/gge/rrfs2/regressionTests/tmp/v2.1.4-4.7-884be1c/OPSROOT/conus12km/stmp/20240527/rrfs_jedivar_07_v2.1.4/det/jedivar_07/`

Further testing revealed that NA12km 3DVAR crashed as well.
So currently, only conus3km can run 3DVAR (or hybrid 3DEnVar).

I switched from GSIBEC to BUMPBEC and then JEDI 3DVAR succeeded. 
This confirmed that the issue came for GSIBEC. 

At the time we tested the NaN fix, we only tested conus3km. So that fix may not be a complete fix.

@TingLei-NOAA Are you able to help check this? Thanks!

<br>
<details>
<summary>*** Click here to see the error messages ***</summary>

```
   gridmod fv3_full_hydro  F
   gsimod*gsimain_initialize: Complete
   control_vectors: length=      180120
   control_vectors: currently allocated=           0
   control_vectors: maximum allocated=           0
   control_vectors: number of allocates=           0
   control_vectors: number of deallocates=           0
  control_vectors: Estimated max memory used=      0.0 Mb
   gsi_metguess_mod*create_: alloc() for met-guess done
   User expected to provide guess-fields (viz. gsiguess_set)
   vars still needing to be filled filled-ps
   filled-u                      filled-v
   tv                            filled-tsen
   filled-q                      filled-prsl
  [000]guess_grids*lwi_mask_(): >>> WARNING <<< : not enough to fill LWI, all Ocean
   m_berror_stats_reg::berror_read_bal_reg(PREBAL_REG):  get balance variables"
   berror_stats".  mype,nsigstat,nlatstat =           0         127         770
   m_berror_stats_reg::berror_read_wgt_reg(PREWGT_REG):  read error amplitudes "
   berror_stats".  mype,nsigstat,nlatstat =           0         127         770
[...]

PR #1703, opened 2026-09-06
updates for spinup cycles and other changes

guoqing-noaa:
1. Currently, `jedivar` and `jedivar_spinup` both copy analysis/diag results to the `jedivar` subdirectory under the `com` directory.
They should copy to different directories. This PR fixes this issue.

2. `exrrfs_getkf.sh` now saves too many `log.out.00*` files into the com directory.  
This PR changes `log.*` to `log.out` to fix this issue.

3. temporary fix - set 'regional nn fill distance' to 5000km for conus12km,na12km, addressing issue #1702 
It will be changed once we find a permanent GSIBEC solution

4. update pyDAmonitor so that it work correctly for spinup cycles.


spanNOAA (approved):

guoqing-noaa:
Tested 24 hours of retro and worked as expected.

scripts/exrrfs_ioda_bufr.sh
MatthewPyle-NOAA, line 1: I didn't notice this change listed in the description - just double checking that removing the ioda_empty.nc copies is intended.

guoqing-noaa: @MatthewPyle-NOAA Thanks for the question. Yes, that's intended. That change is related to PR #1704, but I was working on both PRs at the same time and included it here. I think it is fine as it is just a clean up.

JEDI itself can create/handle empty ioda files internally and we don't need an outside manually created ioda_empty file. Previously, we still kept that file because a special field `lat_lon_pressure` was added offline for the "time thinning" filter of some observers.  But we have transitioned to use the "duplicate thinning" filter and removed the associated offline processing and `lat_lon_pressure` is no longer needed. So it is safe to drop this block.  Thanks!

MatthewPyle-NOAA (approved):

RuifangLi7:
Hi Guoqing,

In jedivar script, two changes still need:

1. Save initial_file as an option since operation does not save this file and this file has big size (line 198). Need to add SAVE_JEDI_ANL=false in config file.
      if [[ "${SAVE_JEDI_ANL^^}" == "TRUE" ]]; then
        cp "${DATA}/${initial_file}" "${COMOUT}/${jedivar_str}/${WGF}/${initial_file%.nc}.${timestr}.nc"
      fi

2.  Save satbias file to spinup directory. (end of script)
     cp "${DATA}"/data/satbias_out/*satbias*.nc "${COMOUT}/${jedivar_str}/${WGF}"

Thanks

guoqing-noaa:
@RuifangLi7 Thanks for that information. Please open a new PR to make those changes. Thanks!

PR #1707, opened 2026-09-08
[rrfs-mpas-jedi] Add nonvar cloud analysis information for conus3.5km

ShawnMurdzek-NOAA:
## DESCRIPTION OF CHANGES: 
- Add required environment variables for the nonvar cloud analysis to run with the conus3.5km mesh
- Change the mesh check in `setup_rocoto.py` to output a warning rather than an error

## TESTS CONDUCTED: 
Successfully generated a conus3km and conus3.5km workflow on Gaea C6. When using an invalid mesh name, the warning is raised, but there is no error from that code block (as expected).

### Machines/Platforms:
- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules
  - [x] Gaea C6


hu5970 (approved):

workflow/setup_rocoto.py
```diff
@@ -48,10 +48,8 @@
 mesh = os.getenv("MESH_NAME", "conus3km")
 # Check compatibility of setup with nonvar cloud analysis
 if os.getenv("DO_NONVAR_CLOUD_ANA", "FALSE").upper() == "TRUE":
-    if mesh not in ["conus3km", "south3.5km", "conus12km"]:
-        print(f'{mesh} is not compatible with the nonvar cloud analysis')
-        print('Please set DO_NONVAR_CLOUD_ANA=false and try again')
-        exit()
+    if mesh not in ["conus3km", "conus3.5km", "south3.5km", "conus12km"]:
+        print(f'WARN: {mesh} is not compatible with the nonvar cloud analysis')
```
guoqing-noaa: @ShawnMurdzek-NOAA Users my wonder why a domain is not compatible with nonvar cloud analysis. What about saying "has not been fully tested with nonvar cloud analysis"?

ShawnMurdzek-NOAA: Good idea! I updated the warning message to make it more descriptive

guoqing-noaa (approved):
LGTM. Thanks for the changes, @ShawnMurdzek-NOAA

MatthewPyle-NOAA (approved):
Will approve, but wondering if conus3.5km should be conus3p5km to match production norms? (I tend to see the "." as being a divider between two parts of a filename such as in production model output)

guoqing-noaa:
> Will approve, but wondering if conus3.5km should be conus3p5km to match production norms? (I tend to see the "." as being a divider between two parts of a filename such as in production model output)

Good point. @MatthewPyle-NOAA!
Since operation likes "p" instead of ".", let's follow it. Thanks!

ShawnMurdzek-NOAA:
> Will approve, but wondering if conus3.5km should be conus3p5km to match production norms? (I tend to see the "." as being a divider between two parts of a filename such as in production model output)

Changed as recommended! If we want to follow this naming convention, we should also change some of the other mesh names in the future (e.g., `fx1.25km`, `south3.5km`, `ar3.5km`), but that's beyond the scope of this PR.

hu5970:
Let's use 3p5km for 3.5 as it is may go to operation in the future. Others we can change them one by one later.

workflow/setup_rocoto.py
MatthewPyle-NOAA, line 1: south3.5km is still listed as a possibility here

ShawnMurdzek-NOAA: @MatthewPyle-NOAA yes, my interpretation of @hu5970's comment is that we'd change mesh names other than `conus3p5km` later given that we'd need to make the change in at least a few places to ensure consistency across the workflow. I can change `south3.5km` to `south3p5km` in `workflow/setup_rocoto.py` now without breaking anything, but then it wouldn't be consistent with the rest of the workflow. Let me know if you still think the change is necessary for this PR.

guoqing-noaa: @MatthewPyle-NOAA
As @ShawnMurdzek-NOAA mentioned, changing to south3p5km will need coordinated efforts on fix files.
Similar changes are needed for ar3.5km and fwx1.25km.  I will create a separate PR to fix this.  Thanks!

MatthewPyle-NOAA (approved):

PR #1719, opened 2026-09-15
[rrfs-mpas-jedi] turn off obsgrouping for aircar, aircft and sfcshp

chunhuazhou:
## DESCRIPTION OF CHANGES: 
This pull request mimics the implementation from RDASApp PR #600 by removing observation grouping (obsgrouping) for horizontally-moving observation types (aircar, aircft, sfcshp). This change serves as a bug fix explicitly targeting the one-step GETKF setup.

## TESTS CONDUCTED: 
One-week retrospective tests were conducted with obsgrouping turned off in one-step GETKF (aircar, aircft, sfcshp). When compared against the control experiment (where grouping was active), the overall forecast impact was found to be neutral.
A standalone test was performed using jedivar, evaluating the implementation both with and without observation grouping for the specified moving observation platforms. Comparing the jedivar analysis outputs confirms that the changes are negligible, as shown below:
```
nccmp-data jedivar_13.NoObGrp/mpasout.nc jedivar_13/mpasout.nc
Variable               Group Count          Sum          Min         Max       Range        Mean      StdDev          L1         L1R          L2         L2R
qv                     /         1  8.88178e-16  8.88178e-16 8.88178e-16           0 8.88178e-16           0 8.88178e-16 1.00757e-07 8.88178e-16 1.00757e-07
u                      /        47  1.36696e-06 -2.98023e-08 4.76837e-07 5.06639e-07 2.90844e-08 1.03433e-07 1.49505e-06 8.29437e-08 7.29305e-07 1.07241e-07
uReconstructZonal      /        18 -1.17497e-07 -1.19209e-07 1.19209e-07 2.38419e-07 -6.5276e-09 4.52197e-08 3.59879e-07 9.66995e-08 1.88491e-07 7.29529e-08
uReconstructMeridional /        28    5.502e-07 -5.96046e-08 4.76837e-07 5.36442e-07   1.965e-08 9.31744e-08 6.70298e-07 7.99769e-08 4.95188e-07 1.12433e-07
```


MatthewPyle-NOAA (commented):
Should the obsgrouping lines in the yaml files be deleted rather than commented out?  I'm thinking so if we definitely don't want to apply this approach for this observation types.

SamuelDegelia-NOAA:
@MatthewPyle-NOAA We should be able to re-enable `obsgrouping` once https://github.com/JCSDA-internal/ioda/pull/1886 is merged. `obsgrouping` is only needed for certain filters that aren't applied to these obs types yet. That's why removing the grouping right now is a good enough solution. But those filters might be used in the future, so I'd recommend turning the grouping back on once we can. Because of that I vote to leave the comments instead of deleting the lines. 

Maybe we add a breadcrumb comment above each block, something like `# disabled until ioda#1886 is merged`.

guoqing-noaa:
Thank @SamuelDegelia-NOAA for detailed explanation. We are good to keep those commented-out lines for now. No need to add extra comments.

@chunhuazhou Could you update the PR number in the ignore file? Thanks!

guoqing-noaa (approved):
LGTM. Thanks for the changes, @chunhuazhou

chunhuazhou:
> @MatthewPyle-NOAA We should be able to re-enable `obsgrouping` once [JCSDA-internal/ioda#1886](https://github.com/JCSDA-internal/ioda/pull/1886) is merged. `obsgrouping` is only needed for certain filters that aren't applied to these obs types yet. That's why removing the grouping right now is a good enough solution. But those filters might be used in the future, so I'd recommend turning the grouping back on once we can. Because of that I vote to leave the comments instead of deleting the lines.
> 
> Maybe we add a breadcrumb comment above each block, something like `# disabled until ioda#1886 is merged`.

Thank you Sam for your explanation!

MatthewPyle-NOAA (approved):

PR #1721, opened 2026-09-15
reverse observation error for T181

Junjun-NOAA:
This is a small fix of observation error for T181. The smaller error has caused model crash in both retro and realtime runs. This PR reverse it to a original larger value.


guoqing-noaa:
@Junjun-NOAA only t181 is affected? What about q181 and uv281?

Junjun-NOAA:
> @Junjun-NOAA only t181 is affected? What about q181 and uv281?

q and uv are fine, only t181.

guoqing-noaa (approved):

MatthewPyle-NOAA (approved):

PR #1722, opened 2026-09-15
add smart fcst groups to allow using different computing resources for different forecast lengths

guoqing-noaa:
We may run 
84 hour forecasts at 00/12z,
72 hour forecasts at 06, 18;   
12 hour forecasts every 3h;  
1h forecasts at the remaining cycles

Here, 84 hours need more nodes, longer walltime. We don't want the 12h or 1h forecasts request similar amount of computing resources and hence unnessarily wait in the queue for a long time.

So we would like to request different computing resources for different forecast lengths.

This PR enables this capability but smartly arranging fcst tasks into different groups:
```
`fcst`:  the shortest forecast length.  This is also the default group if we have only 1 forecast length
`fcst_l`:  longer `fcst`
`fcst_xl`:  extra longer `fcst`
`fcst_xxl`:  extra-extra longer `fcst`
....
```
The groups only affect the rotoco workflow task names. The `com/` and `stmp/` directory structure keeps unchanged.

Here is an example to use different WALLTIME for different forecast lengths:
```
export WALLTIME_FCST="0:25:00"
export WALLTIME_FCST_L="0:38:00"
export WALLTIME_FCST_XL="1:00:00"
```


guoqing-noaa:
tested 24 cycles and worked as expected.

MatthewPyle-NOAA:
The fcst_h and fcst_hh kind of syntax makes me think it is a one-digit hour and two-digit hour, respectively.  Maybe use a different letter to remove this confusion?  And wasn't sure if something like fcst_m, fcst_l, and fcst_xl might work - make it like clothing sizes?  Just some random thoughts...

guoqing-noaa:
> The fcst_h and fcst_hh kind of syntax makes me think it is a one-digit hour and two-digit hour, respectively. Maybe use a different letter to remove this confusion? And wasn't sure if something like fcst_m, fcst_l, and fcst_xl might work - make it like clothing sizes? Just some random thoughts...

Good comment about "h" and "hh", :). I like the idea of "l", "xl", ....

One comment about "m".  If one experiment only has 2 fcst lengths, would it be more intuitive to use `fcst, fcst_l` instead of `fcst, fcst_m`?

Will `fcst, fcst_l, fcst_xl, fcst_xxl, fcst_xxxl, ...` work for you?

guoqing-noaa:
@MatthewPyle-NOAA I will move forward with `fcst, fcst_l, fcst_xl, fcst_xxl, fcst_xxxl, ...` unless I hear otherwise from you.  Note, this only affects the rocoto workflow task names.  ecflow will use `fcst` for all cycles.

MatthewPyle-NOAA:
> > The fcst_h and fcst_hh kind of syntax makes me think it is a one-digit hour and two-digit hour, respectively. Maybe use a different letter to remove this confusion? And wasn't sure if something like fcst_m, fcst_l, and fcst_xl might work - make it like clothing sizes? Just some random thoughts...
> 
> Good comment about "h" and "hh", :). I like the idea of "l", "xl", ....
> 
> One comment about "m". If one experiment only has 2 fcst lengths, would it be more intuitive to use `fcst, fcst_l` instead of `fcst, fcst_m`?
> 
> Will `fcst, fcst_l, fcst_xl, fcst_xxl, fcst_xxxl, ...` work for you?

That sounds good to me

MatthewPyle-NOAA (approved):

Issue #1728, opened 2026-09-17
[rrfs-nco] 20260917 rrfs_det_process_smoke failure


WeiWei-NCO:
On 20260917, 18z det_process_smoke failed.

Log: `/lfs/h1/ops/prod/output/20260917/rrfs_det_process_smoke_18.o289903614`
```
+ 89 + cpreq -p /lfs/f1/ops/prod/tmp/rrfs_det_process_smoke_18.289903614.cbqs01/RAVE-HrlyEmiss-3km_v2r0_blend_s202609171500000_e202609171559590_c202609171702110.nc.tmp /lfs/h1/ops/prod/com/rrfs/v1.0/RAVE_INTP/rave_intp.20260917
cp: cannot stat '/lfs/f1/ops/prod/tmp/rrfs_det_process_smoke_18.289903614.cbqs01/RAVE-HrlyEmiss-3km_v2r0_blend_s202609171500000_e202609171559590_c202609171702110.nc.tmp': No such file or directory
```

The missing file was linked to /dcom. However, the actual RAVE file was late and still a `nc.tmp` version when the link was created. 
```
lrwxrwxrwx 1 ops.prod prod 123 Sep 17 18:35 /lfs/f1/ops/prod/tmp/rrfs_det_process_smoke_18.289903614.cbqs01/RAVE-HrlyEmiss-3km_v2r0_blend_s202609171500000_e202609171559590_c202609171702110.nc.tmp -> /lfs/h1/ops/prod/dcom/20260917/rave/RAVE-HrlyEmiss-3km_v2r0_blend_s202609171500000_e202609171559590_c202609171702110.nc.tmp
```

In `exrrfs_process_smoke.sh`, all files are linked with wildcards. 
` ${LN} -snf ${FIRE_RAVE_DIR}/${PDY}/rave/RAVE-HrlyEmiss-3km_* ${fire_rave_dir_work}/.`

Please update the logic so only complete files are getting linked/processed. 

@opsprod

MatthewPyle-NOAA:
Is this restarting the 30 day clock?

WeiWei-NCO:
> Is this restarting the 30 day clock?

No :)

WeiWei-NCO:
Fixed with #1729

PR #1729, opened 2026-09-17
Filter DCOM file type to be linked

LinGan-NOAA:
## DESCRIPTION OF CHANGES: 
Fix dead link issue discovered in process_smoke job. Instead of linking all existing files in the source directory in DCOM. It only link the final netcdf file to avoid accidently use of the incomplete tmp file.

## ISSUE: 
- Fixes the issue(s) mentioned in #1728


WeiWei-NCO (approved):
Looks good. Thank you, Lin.

JamesPolly-NCO (approved):
Thanks very much Lin!

MatthewPyle-NOAA (approved):
Thanks for getting this done so quickly.

PR #1731, opened 2026-09-18
[rrfs-dev] Add Ursa settings file, suite generator and helper utilities [1/3]

SamuelDegelia-NOAA:
## DESCRIPTION OF CHANGES: 

Part 1 of 3 in a series that adds the ability to run RRFS retros with ecflow on Ursa. The three PRs depend on each other and are meant to land in order (1 → 2 → 3); they are split only to keep each one reviewable.

This PR adds the pieces needed to configure and inspect an RRFS run on Ursa. Nothing here changes how the suite behaves on WCOSS2.

Changes include: 

- **`ecf/defs/ursa_config.sh`** — one place for the settings we control: work directories, Slurm account and QOS, ecflow host, fix tree, staged data, retro period. It is sourced by the three scripts below, and environment variables still override it for one-off runs. A single `URSA_WORK_BASE` moves `submit/` (jobs and logs), `output/`, `ptmp/` (COM) and `stmp/` (work directories) together. This is the closest thing the ecflow workflow has to the Rocoto `ush/config.sh`; job resources are deliberately not here, because they come from `fix/workflow/<WGF>/workflow.conf` at run time.
- **`ecf/defs/make_ursa_def.sh`** — writes an Ursa copy of `nco_para`. The task tree is untouched; only suite-wide settings differ: `sbatch --ignore-pbs` as the job command (plain `sbatch` also parses the `#PBS` lines and rejects the queue), Ursa paths, account and QOS, `ECF_LOGHOST` (which `head.h` reads and no server variable defines), and the staged-data search paths. With `RETRO=YES` it also relaxes the 399 clock-time triggers in the suite so cycles are not gated on the wall clock.
- **`ecf/setup_ecf_links.sh`** — on Ursa, links `fix/` into a copy of the WCOSS2 fix tree (`FIX_RRFS_DIR`) and applies the Ursa-only `workflow.conf` overrides: GRIB2 deterministic boundaries, and 48 forecast ranks per node. Both edits are guarded by hostname, so the WCOSS2 path is unchanged.
- **`ush/ursa/compath.py`** — prod_util's copy exits unless `/lfs/h1` exists, which would stop every job on Ursa.
- **`ush/ursa/rrfsstat`** — prints the suite state as text, in the spirit of `rocotostat`, so progress can be read without the GUI: a per-cycle rollup by default, `-t` for task detail with the matching Slurm state, filters by cycle and state, and `-o` to write a file.
- **`.gitignore`** — ignores the two generated suite definitions which are outputs of the generators and should be regenerated rather than committed.

## TESTS CONDUCTED: 

After these three PRs, a full **deterministic** cold-start cycle now runs on Ursa (2024-05-06 03z): boundaries, cold start, 60-node hybrid GSI analysis, diagnostics, cloud analysis, 42-node forecast, restarts saved. Not yet exercised, and therefore not claimed to work: the **EnKF** (30 members), **ensemble forecasts**, **fire weather** (its EnKF analysis asks for 90 nodes, above Ursa's 75-node per-job limit), **smoke and dust processing** (`generate_fire_emissions.py` runs `python` without `srun`, so ESMF's MPI init aborts). 

### Machines/Platforms:

- WCOSS2
  - [ ] Cactus/Dogwood
  - [ ] Acorn
- RDHPCS
  - [ ] Hera
  - [x] Ursa
  - [ ] Jet
  - [ ] Orion
  - [ ] Hercules

### Test cases: 

- [x] Engineering tests
  - [ ] Non-DA engineering test
  - [x] DA engineering test
    - [x] Retro
    - [ ] Ensemble
    - [ ] Parallel
- [ ] RRFS fire weather
- [ ] RRFS_A:
- [ ] RRFS_B:
- [ ] RTMA:
- [ ] Others:


SamuelDegelia-NOAA:
Oops, did not mean to open this into `NOAA-EMC:release/v1.0`.

ShunLiu-NOAA:
Thank you Sam. As the initial effort to port the ecFlow workflow to URSA, this is a good starting point.

ShunLiu-NOAA (approved):

MatthewPyle-NOAA (approved):

